PR #28, opened 2024-01-28
fixes: add some additional logic checks and unit tests.

robpolak:
**What changed? Why?**
1. Add unit tests
2. Fixed some logic + wrote unit tests

**Notes to reviewers**

**How has it been tested?**

PR #29, opened 2024-01-28
feat: factor out parseFrameMessage

cnasc:
**What changed? Why?**
While working with frames, it's beneficial to be able to get the whole validated message. Since we parse and validate the message as part of `getFrameAccountAddress`, I pulled the relevant code out into a new function.

**Notes to reviewers**

**How has it been tested?**
yolo'd this, would love to hear any strategies for testing
edit: did manually test in a frame by copying code over and it appears to work

### DIFF
--- a/src/core/getFrameAccountAddress.ts
+++ b/src/core/getFrameAccountAddress.ts
@@ -1,18 +1,9 @@
-import { HubRpcClient, Message, getSSLHubRpcClient } from '@farcaster/hub-nodejs';
-
-/**
- * Farcaster Hub for signature verification, consider using a private hub if needed:
- * https://docs.farcaster.xyz/hubble/hubble
- */
-const HUB_URL = 'nemes.farcaster.xyz:2283';
+import { parseFrameMessage } from './parseFrameMessage';
 
 type FidResponse = {
   verifications: string[];
 };
 
-function getHubClient(): HubRpcClient {
-  return getSSLHubRpcClient(HUB_URL);
-}
 /**
  * Get the Account Address from the Farcaster ID using the Frame.  This uses a Neynar api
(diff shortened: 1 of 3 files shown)


src/core/getFrameAccountAddress.ts
```diff
@@ -20,16 +15,7 @@ async function getFrameAccountAddress(
...
-  // Get the message from the request body
-  const frameMessage: Message = Message.decode(
-    Buffer.from(body?.trustedData?.messageBytes ?? '', 'hex'),
-  );
-  // Validate the message
-  const result = await client.validateMessage(frameMessage);
-  if (result.isOk() && result.value.valid && result.value.message) {
-    validatedMessage = result.value.message;
-  }
+  const validatedMessage = await parseFrameMessage(body);
```
cnasc, line 18: maybe we want to make this a param? That way if you're parsing the message to get other data anyway you can just pass the result in to this function rather than loading same data twice. Could do this with an overload

Zizzamia:
Ciao @cnasc, after merging https://github.com/coinbase/onchainkit/pull/28 form @robpolak we got some conflic. Love your idea of `parseFrameMessage`, please resolve the conflict and go ahead with it.

Zizzamia (approved):
DOOOPE!!!

As follow up would you mind add this in the README with a nice example on how to use it.

After that I will cut a new release.

cnasc:
Ehm @robpolak's tests are broken in this PR

Zizzamia:
😅 @cnasc can you help us get on a solid testing case, so that we can add it in CI, and than deploy, please.

Ideally would be dope to release a new OnchainKit version in the next couple of hours.

PR #37, opened 2024-01-29
fix: Refactor Farcaster typing to be explicit

robpolak:
**What changed? Why?**
1. Be more explicit on our Farcaster request / response types.
   2.  Decided against importing the Farcaster types from @farcaster/hub-nodejs as the object types were more complex and could confuse consumers.  Extract + simplify the interface.
1.   Added a farcaster message verification integration test.

**Notes to reviewers**

**How has it been tested?**

```
 yarn test
 PASS  src/utils/neynar/user/neynarUserFunctions.test.ts
 PASS  src/core/getFrameValidatedMessage.test.ts
 PASS  src/core/getFrameMetadata.test.ts
 PASS  src/core/getFrameAccountAddress.test.ts

Test Suites: 4 passed, 4 total
Tests:       8 passed, 8 total
Snapshots:   0 total
Time:        1.876 s, estimated 2 s
Ran all test suites.

```

```
❯ yarn test:integration         
 PASS  src/utils/neynar/neynar.integ.ts
 PASS  src/core/farcaster.integ.ts

Test Suites: 2 passed, 2 total
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        2.399 s
Ran all test suites.

```

### DIFF
--- a/src/core/farcasterTypes.ts
+++ b/src/core/farcasterTypes.ts
@@ -1,0 +1,38 @@
+export interface FrameRequest {
+  untrustedData: FrameData;
+  trustedData: {
+    messageBytes: string;
+  };
+}
+
+export interface FrameResponse {
+  data: FrameData;
+}
+
+export interface FrameData {
+  fid: number;
+  url: string;
+  messageHash: string;
+  timestamp: number;
+  network: number;
+  buttonIndex: number;
+  castId: {
+    fid: number;
+    hash: string;
+  };
+}
+
+export function convertToFrame(json: any) {
+  return {
+    fid: json.fid,
+    url: json.frameActionBody?.url.toString(),
+    messageHash: json.messageHash,
+    timestamp: json.timestamp,
+    network: json.network,
+    buttonIndex: json.frameActionBody?.buttonIndex,
+    castId: {
+      fid: json.frameActionBody?.castId?.fid,
+      hash: json.frameActionBody?.castId?.hash,
+    },
+  };
+}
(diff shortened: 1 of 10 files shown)


src/core/getFrameValidatedMessage.ts
```diff
@@ -18,9 +19,9 @@ export function getHubClient(): HubRpcClient {
  *
  * @param body The JSON received by server on frame callback
  */
-async function getFrameValidatedMessage(body: {
```
Zizzamia, line 21: Open question should we keep this as `getFrameValidatedMessage` or have `getFrameMessage`.
As now we are returing `isValid`.

Zizzamia: Also should is `data` more correct than `message`?

src/core/farcasterTypes.ts
```diff
@@ -0,0 +1,39 @@
+export interface FrameRequest {
+  untrustedData: FrameData;
+  trustedData: {
+    messageBytes: string;
+  };
+}
+
+export interface FrameValidationResponse {
```
cnasc, line 8: I think this could be a tagged union, like:
```ts
type FrameValidationResponse =
  | { isValid: true, data: FrameData }
  | { isValid: false, data: undefined }
```

Then if `isValid` is true the typechecker should know that data is defined and vice-versa

robpolak: that is pretty cool

src/core/getFrameValidatedMessage.ts
robpolak: agree and agree .. pr updated

Zizzamia (approved):
Amazing!!!
I am going to merge and then release a `0.2.0` version at 5pm PST.
With explicit call on this breaking changes.

PR #49, opened 2024-01-31
feat: new OnchainName component

alvaroraminelli:
**What changed? Why?**

Adding the first `OnchainName` component of our  Identity Kit 👤

The `OnchainName` component primarily focuses on showcasing ENS names for given Ethereum addresses, and defaults to displaying a sliced version of the address when an ENS name isn't available.

```tsx
import { OnchainName } from '@coinbase/onchainkit';

<OnchainName address="0x1234567890abcdef1234567890abcdef12345678" sliced={false} />;
```

**Notes to reviewers**

**How has it been tested?**

- Unit tests only.


README.md
```diff
@@ -213,6 +213,29 @@ type FrameMetadataResponse = {
 <br />
 <br />
 
+## OnchainName 📛
```
Zizzamia, line 216: This new section will be called Identity Kit, which will have a callection of components and utilities around Onchain Identity.

README.md
```diff
@@ -213,6 +213,31 @@ type FrameMetadataResponse = {
 <br />
 <br />
 
+## Identity Kit 👤
+
+### OnchainName
+
+The `OnchainName` component primarily focuses on showcasing ENS names for given Ethereum addresses, and defaults to displaying a sliced version of the address when an ENS name isn't available.
```
Zizzamia, line 220: How about
```
The `OnchainName` component is used to display ENS names associated with Ethereum addresses. When an ENS name is not available, it defaults to showing a truncated version of the address.
```

README.md
```diff
@@ -213,6 +213,31 @@ type FrameMetadataResponse = {
...
+
+```tsx
+import { OnchainName } from '@coinbase/onchainkit';
+
+<OnchainName address="0x1234567890abcdef1234567890abcdef12345678" sliced={false} />;
+```
+
+**@Param**
```
Zizzamia, line 228: Oh, I think we can use types going forward to better describe what's going on. Check the latest README.

README.md
```diff
@@ -213,6 +213,31 @@ type FrameMetadataResponse = {
...
+## Identity Kit 👤
+
+### OnchainName
+
+The `OnchainName` component primarily focuses on showcasing ENS names for given Ethereum addresses, and defaults to displaying a sliced version of the address when an ENS name isn't available.
+
+```tsx
+import { OnchainName } from '@coinbase/onchainkit';
+
+<OnchainName address="0x1234567890abcdef1234567890abcdef12345678" sliced={false} />;
```
Zizzamia, line 225: I think we can do a bit more here to showcase a real example, like have a small component using `<OnchainName>`

And have as comment how the address will result.

src/components/OnchainName.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { useMemo } from 'react';
+import { getSlicedAddress } from '../core/address';
+import { useEnsName } from '../hooks/useEnsName';
+import type { Address } from 'viem';
+
+type OnchainNameProps = {
+  address?: Address;
```
Zizzamia, line 7: I think the `address` should be mandatory

src/components/OnchainName.tsx
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @param {Address} address - Ethereum address to be displayed.
+ * @param {string} [className] - Optional CSS class for custom styling.
+ * @param {boolean} [sliced=true] - Determines if the address should be sliced when no ENS name is available.
+ * @param {React.HTMLAttributes<HTMLSpanElement>} [props] - Additional HTML attributes for the span element.
+ */
+export function OnchainName({ address, className, sliced = true, props }: OnchainNameProps) {
+  const { ensName } = useEnsName(address);
+
+  const normilizedAddress = useMemo(() => {
```
Zizzamia, line 26: Add a comment here why we have `useMemo`

Zizzamia: Also should we avoid have CPU execution for the `normilizedAddress` if in case of ensName we will use that anyway?

src/store/client.ts
```diff
@@ -0,0 +1,7 @@
+import { createPublicClient, http } from 'viem';
+import { mainnet } from 'viem/chains';
+
+export const publicClient: ReturnType<typeof createPublicClient> = createPublicClient({
```
Zizzamia, line 4: I think `client` should go in `src/network` or more a location where we handle Data Layer stuff. cc @robpolak

src/store/inMemoryStorageService.ts
```diff
@@ -0,0 +1,32 @@
+import { StorageInterface } from '../types';
+
+// InMemoryStorageService: A service to store data in JavaScript memory.
+// It's a simple key-value storage using a Map. The data is stored as long as the page is not reloaded.
+export class InMemoryStorageService implements StorageInterface {
```
Zizzamia, line 5: Let's convert this to functional at some point.

Zizzamia:
Before we merge, let's have also a quick review from @cnasc, @robpolak, @wespickett and @Sneh1999 on this epic first Identity component.

src/components/OnchainName.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    (useEnsName as jest.Mock).mockReturnValue({ ensName: 'testname.eth' });
+    render(<OnchainName address={testAddress} />);
+
+    expect(screen.getByText('testname.eth')).toBeInTheDocument();
+  });
+
+  it('displays sliced address when ENS name is not available and sliced is true as default', () => {
+    (useEnsName as jest.Mock).mockReturnValue({ ensName: null });
+    (getSlicedAddress as jest.Mock).mockImplementation(
+      (addr) => addr.slice(0, 6) + '...' + addr.slice(-4),
```
cnasc, line 30: Does this need to be mocked? I notice that the implementation here is slightly different from the one in the real function. Is that significant?

src/components/OnchainName.tsx
cnasc: I think the idea here is to always display 
```suggestion
  const normalizedAddress = useMemo(() => {
```

cnasc: Also, I think the idea here is that the input address will display until ens name resolves. I don't think that's unreasonable, but maybe we'd prefer an explicit loading state?

src/store/inMemoryStorageService.ts
cnasc: Do we see this as a frontend-only use case? If so, it might be best to expose this as a context

src/components/OnchainName.tsx
alvaroraminelli: Great call. Yes, we need loading state here. Wondering how the UX would look like when you have a list of users in the UX, will things shift and change like crazy. User might opt-in to display loading state. 

Need some UX thinking here.

src/store/inMemoryStorageService.ts
alvaroraminelli: I see this evolving towards sharing in context as well. We might want build a light version of react-query for actions.

README.md
alvaroraminelli: soon we might have autogen docs that will be based on type correct?

README.md
alvaroraminelli: I am thinking add png file with little screenshot of the result? or we can have a link to stackbliz that show case the components.

src/components/OnchainName.tsx
alvaroraminelli: agreed with @cnasc. vote to accept to make less cumbersome for consumers.

src/store/client.ts
alvaroraminelli: LGTM

package.json
```diff
@@ -17,11 +17,20 @@
     "release:publish": "yarn install && yarn build && changeset publish",
     "release:version": "changeset version && yarn install --immutable"
   },
+  "dependencies": {
+    "viem": "^2.5.0"
```
alvaroraminelli, line 21: build fails if we remove the dependency

README.md
Zizzamia: Maybe... There is a difference between `type` information, and how to teach how to use something.

Vitepress will automate docs creation in some capacity, but the ART of teaching how to use a function is still a manual work.

README.md
Zizzamia: Let's circle back on this one. We still miss the default Tailwind css here.

src/store/inMemoryStorageService.ts
Zizzamia: > I see this evolving towards sharing in context as well. 

React Context are a slippery slop, where folks think everything should be a context. Library should have like 90% of their state being just variable shared between functions. And for very very small exential stuff you can have Context that optimizes the Hooks story.

So before we build any React Context, it's important to ask the question, on where is the benefit for the end application. If it's just a nice to have for us as developer, than we don't need that. But again, for Hooks Rendering optimization, that we can talk about it.

package.json
Zizzamia: let's add it back.

wespickett:
lgtm. We ended up switching to extend the next/jest config in build-onchain-apps, but if tests run successfully as is, then it should be fine with the current config https://github.com/coinbase/build-onchain-apps/blob/main/template/jest.config.js#L1

src/components/OnchainName.test.tsx
alvaroraminelli: tried the approach below but it is not working. fallback to mocking the entire function. 

```
import * as addressUtils from '../core/address';

beforeEach(() => {
    jest.clearAllMocks();
    jest.spyOn(addressUtils, 'getSlicedAddress');
  });

expect(addressUtils.getSlicedAddress).toHaveBeenCalledTimes(0);
```

src/components/OnchainName.tsx
```diff
@@ -0,0 +1,43 @@
...
+  // wrapped in useMemo to prevent unnecessary recalculations.
+  const normalizedAddress = useMemo(() => {
+    if (!ensName && !isLoading && sliced) {
+      return getSlicedAddress(address);
+    }
+    return address;
+  }, [address, isLoading]);
+
+  if (isLoading) {
+    return null;
```
Zizzamia, line 35: Not for later, this will cause a Layout Shift. Something to consider in follow up PR.

Zizzamia (approved):
BRILLIANT

Issue #44, opened 2024-01-30
Feature Request: getFrameMetadata should return a whole NextJS Metadata object with OG tags filled


cnasc:
### Describe the solution you'd like

Frames don't work if full OG tags aren't there, by returning a full Metadata object for the user to export we can make our code a bit more error-proof.

Saw this bite someone earlier today, and they were confused why the function didn't result in an object that just worked.

Zizzamia:
Solved with `v0.6.1`

PR #50, opened 2024-01-31
Add `getFrameHtml`

taycaldwell:
**What changed? Why?**
Added a new server-side helper method for creating a response containing a valid frame.

- Shares the same `FrameMetadata` types as the client for constructing page metadata
- Surface level API is simple, easy to understand, and uses idiomatic Javascript/TypeScript
- Removes the need for developers to manually handcraft HTML strings on the server-side

**Usage / Screenshots**
Screenshot 2024-01-31 at 1 38 40 PM

**Notes to reviewers**
N/A

**How has it been tested?**
Manually. Will write and commit test files tomorrow.


taycaldwell:
On second thought - this is probably too tightly coupled to Next.js with the use of `NextResponse`.

Might be better to decouple by creating a class that is just the HTML content that can be fed into a NextResponse.

package.json
```diff
@@ -8,7 +8,9 @@
     "@farcaster/fishery": "^2.2.3",
     "@farcaster/hub-nodejs": "^0.10.21",
     "ethers": "^6.10.0",
+    "next": "13.5.6",
```
Zizzamia, line 11: Is it `next` necessary for this on? I will guess `react-dom` is the only really dependency need it.

src/core/createFrameResponse.ts
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Creates a response with a new valid frame to be rendered.
+ *
+ * @param buttons: The buttons to use for the frame.
+ * @param image: The image to use for the frame.
+ * @param post_url: The URL to post the frame to.
+ * @param refresh_period: The refresh period for the image used.
+ * @returns An HTML response containing metadata for the frame.
+ */
+function createFrameResponse({
```
Zizzamia, line 13: There is something about this name that I am not fully sold.

Part of me thinks `getFrameReponse` is better, as you are describing the end goal vs the action.

Also I am a bit overthinknig if something like `getFrameNextResponse` is more correct. As the type returned is `NextResponse`.

@taycaldwell  @cnasc @robpolak thoughts?

src/core/createFrameResponse.ts
```diff
@@ -0,0 +1,58 @@
+import { NextResponse } from 'next/server';
+import { Button, FrameMetadata } from './sharedTypes';
+
+/**
```
Zizzamia, line 4: Btw we are missing unit-tests for this utility, can you please add them and cover all different cases.

src/core/createFrameResponse.ts
```diff
@@ -0,0 +1,58 @@
...
+function createFrameResponse({
+  buttons,
+  image,
+  post_url,
+  refresh_period,
+}: FrameMetadata): NextResponse {
+  const imageHtml = image ? `<meta property="fc:frame:image" content="${image}" />` : '';
+  // Ensure only up to 4 buttons are processed
+  let buttonsHtml = '';
+  if (buttons) {
```
Zizzamia, line 22: Isn't the type already checking if the buttons exist? I think we can avoid this condition.

src/core/createFrameResponse.ts
```diff
@@ -0,0 +1,58 @@
...
+    const validButtons = buttons
+      .slice(0, 4)
+      .filter((button): button is Button => button !== undefined);
```
Zizzamia, line 25: I think also those checks are not need it, and `FrameMetadata` type should cover those automatically.

src/core/createFrameResponse.ts
```diff
@@ -0,0 +1,58 @@
...
+    buttonsHtml = validButtons
+      .map((button, index) => {
+        let buttonHtml = `<meta property="fc:frame:button:${index + 1}" content="${button.label}" />`;
+        if (button.action) {
+          buttonHtml += `\n<meta property="fc:frame:button:${index + 1}:action" content="${button.action}" />`;
+        }
+        return buttonHtml;
+      })
+      .join('\n');
```
Zizzamia, line 34: If the code is in the end just shipped back to Next, I think we don't need `\n`, right? It's consmetic that is not even visible to the user. Oh no?

src/core/createFrameResponse.ts
```diff
@@ -0,0 +1,58 @@
...
+  }
+  const postUrlHtml = post_url ? `<meta property="fc:frame:post_url" content="${post_url}" />` : '';
```
Zizzamia, line 36: We don't need to check `post_url` the type should already check that, I believe.

src/core/createFrameResponse.ts
```diff
@@ -0,0 +1,58 @@
...
+        <meta property="fc:frame" content="vNext" />
+        ${imageHtml}
+        ${buttonsHtml}
+        ${postUrlHtml}
+        ${refreshPeriodHtml}
+      </head>
+    </html>`;
+
+  const response = new NextResponse(html);
+  response.headers.set('Cache-Control', 'private, no-cache, no-store, max-age=0');
```
Zizzamia, line 53: Why are we setting these headers? is it Farcaster already doing their own caching and this headers are not really used?

src/core/getFrameMetadata.ts
```diff
@@ -29,14 +19,18 @@ export const getFrameMetadata = function ({
...
   };
-  buttons.forEach((button, index) => {
-    metadata[`fc:frame:button:${index + 1}`] = button.label;
-    if (button.action) {
-      metadata[`fc:frame:button:${index + 1}:action`] = button.action;
-    }
-  });
   metadata['fc:frame:image'] = image;
-  metadata['fc:frame:post_url'] = post_url;
+  if (buttons) {
```
Zizzamia, line 23: Why changing this code? Types already checked for buttons to exist.

src/core/getFrameMetadata.ts
```diff
@@ -29,14 +19,18 @@ export const getFrameMetadata = function ({
...
+    buttons.forEach((button, index) => {
+      metadata[`fc:frame:button:${index + 1}`] = button.label;
+      if (button.action) {
+        metadata[`fc:frame:button:${index + 1}:action`] = button.action;
+      }
+    });
+  }
+  if (post_url) {
```
Zizzamia, line 31: Why checking for `post_url`? The type should already check for this.

src/core/sharedTypes.ts
```diff
@@ -0,0 +1,11 @@
+export type Button = {
```
Zizzamia, line 1: `sharedTypes` is a odd name file. I will probably just move these types into the other type file and rename Button into `FrameButtonMetadata` so it's more explicit.

Zizzamia:
Btw the CI is failing, please check what's missing as well.

Zizzamia:
Also open question, should Next be a dependency? Or we should more just create the HTML that the response needs to use.

Than if someone uses any NodeJS library they can still use this method.

@taycaldwell @cnasc @robpolak thoughts?

src/core/createFrameResponse.ts
cnasc: The type does check that, though I think the type doesn't fully conform to the spec. The only required properties are `fc:frame`, `fc:frame:image`, and `og:image`. (and our function doesn't itself handle the og image, see #44)

Zizzamia: The OG story is tricky, as Next has their own way to handle OG. @cnasc let's keep that conversation in the #44 issue. But is not as easy to solve. Would love your thoughts on that.

src/core/createFrameResponse.ts
taycaldwell: I updated `FrameMetadata` to reflect the required/optional fields, so `buttons` is now optional, as it is in the Farcaster API.

src/core/createFrameResponse.ts
taycaldwell: Removed these checks.

src/core/createFrameResponse.ts
taycaldwell: Yeah, not needed. Removed.

src/core/createFrameResponse.ts
taycaldwell: Removed the dependency on Next.js by having it return the HTML string instead of wrapping it in a NextResponse and returning that.

Because of this and the suggestions above, I updated the name to `getFrameHtml`.

src/core/createFrameResponse.ts
taycaldwell: Removed dependency to Next.js, and the unnecessary headers.

Users can now use whatever Response type and headers they wish to set themselves.

src/core/getFrameMetadata.ts
taycaldwell: I updated `FrameMetadata` fields. `buttons` and `post_url` are now considered optional (as they are in the Farcaster Frames API).

src/core/getFrameMetadata.ts
taycaldwell: I updated `FrameMetadata` fields. `buttons` and `post_url` are now considered optional (as they are in the Farcaster Frames API).

src/core/sharedTypes.ts
taycaldwell: Moved the types into the existing `farcasterTypes.ts` file and renamed `Button` to `FrameButtonMetadata` as suggested.

taycaldwell:
I like the idea of keeping Onchainkit framework agnostic and removing the dependency of Next.js while still providing the utility.

src/core/createFrameResponse.ts
taycaldwell: Added test file

src/core/getFrameMetadata.ts
Zizzamia: Wait, Buttons are optional, that seems very odd!

taycaldwell: Agreed. Frames aren't much different than posting an image preview of a site without the buttons... but that's what their docs say:

Screenshot 2024-01-31 at 1 42 07 PM

src/index.ts
```diff
@@ -2,6 +2,7 @@
 const version = '0.3.0';
 
 export { version };
+export { getFrameHtml } from './core/getFrameHtml';
```
Zizzamia, line 5: I like the name `getFrameHtml`

src/core/getFrameMetadata.ts
Zizzamia: sad, but I can see their point

Zizzamia: Thank you for triple checking their standard

Zizzamia (approved):
Amazing work @taycaldwell 
I am going to merge this, and take care for some extra polishing, docs and cut the release.

PR #52, opened 2024-01-31
fix: make docs more clear re: app vs pages router

### LINKED ISSUE #51: Bug: @grpc/grpc-js library is only for node and causing issues in NextJS with getFrameMetadata()

When importing `getFrameMetadata` from the latest `@coinbase/onchainkit` version (0.3.0) and previous version (0.2.1) fails due to @grpc/grpc-js imports.

There should be no errors when importing `getFrameMetadata` to my page.tsx

- OnchainKit: [0.3.0]

cnasc:
This PR resolves #51 

**What changed? Why?**
Added documentation indicating requirement to only call server-side in Pages Router. Ideally users should be able to use `getFrameMetadata` anywhere since it does not rely on any dependencies, but it seems like the other functions spoil it for now.

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # [OnchainKit](https://github.com/coinbase/onchainkit/)
 
-> OnchainKit is a collection of tools to build world-class onchain apps with CSS, React, and Typescript.
+> OnchainKit is a collection of tools to build world-class onchain apps with NextJS.
 
 ## Getting Started
 
@@ -20,6 +20,10 @@ npm install @coinbase/onchainkit
 # Use PNPM
 pnpm add @coinbase/onchainkit
 ```
+
+### NextJS Pages Router
+
+If your app uses [NextJS Pages Router](https://nextjs.org/docs/pages), you'll need to ensure that Framekit is only called server-side (e.g., in `getServerSideProps` or `getStaticProps`). In [NextJS App Router](https://nextjs.org/docs/app) pages are React Server Components by default, so this is only a concern if you are calling from a client component.
 
 <br />
 <br />


README.md
```diff
@@ -2,7 +2,7 @@
 
 # [OnchainKit](https://github.com/coinbase/onchainkit/)
 
-> OnchainKit is a collection of tools to build world-class onchain apps with CSS, React, and Typescript.
+> OnchainKit is a collection of tools to build world-class onchain apps with NextJS.
```
cnasc, line 5: Right now we have a hard dependency on NextJS, so it makes sense to mention that up front

Zizzamia: I don't think NextJS should be a hard dependency. I do believe we should actually change our code to make it easier to use in any Node / TS enviroment.

README.md
```diff
@@ -21,6 +21,10 @@ npm install @coinbase/onchainkit
 pnpm add @coinbase/onchainkit
 ```
 
+### NextJS Pages Router
+
+If your app uses [NextJS Pages Router](https://nextjs.org/docs/pages), you'll need to ensure that Framekit is only called server-side (e.g., in `getServerSideProps` or `getStaticProps`). In [NextJS App Router](https://nextjs.org/docs/app) pages are React Server Components by default, so this is only a concern if you are calling from a client component.
```
Zizzamia, line 26: This is too much knowledge to share... I think it's important to be more explicit in our way to share import.

Docs shouldn't be the first solution, but instead the question I want to answear is, how to improve the library to be more flexible.

README.md
cnasc: Agreed that the goal should be a universal library.

Zizzamia:
Closing this as we solved the issue in a different way.

PR #70, opened 2024-02-02
fix: tsc build

wespickett:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


Zizzamia:
@wespickett when you have a second can you run `yarn check` please.

PR #71, opened 2024-02-02
feat: added FrameMetadata component

Zizzamia:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/components/FrameMetadata.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { FrameMetadata } from './FrameMetadata';
+
+describe('FrameMetadata', () => {
+  it('renders', () => {
+    expect(FrameMetadata).toBeDefined();
+  });
+
+  it('renders with image', () => {
+    const meta = render(<FrameMetadata image="https://example.com/image.png" />);
+    expect(meta.container.querySelector('meta[property="fc:frame:image"]')).not.toBeNull();
```
cnasc, line 15: nit: the not null check is redundant when we also check that the value is defined in the next expect

Zizzamia: Nice

PR #90, opened 2024-02-06
feat: Add wrapper component to FrameMetadata

syntag:
**What changed? Why?**
Adds optional `wrapperComponent` prop to `FrameMetadata` component that defaults to `React.Fragment` when not passed (original behavior).  Also exported `FrameMetadataResponse` which can be useful when using `getFrameMetadata` in a TS project.

Example use-case for project using `react-helmet-async` (which expects only valid head tags to be rendered directly inside `Helmet`):
```js
import { FrameMetadata } from '@coinbase/onchainkit';
import { Helmet } from 'react-helmet-async';

const PageComponent = () => (
  <FrameMetadata
    image="https://zizzamia.xyz/park-1.png"
    buttons={[{ label: 'Click Me', action: 'post_redirect' }]}
    wrapperComponent={Helmet}
  />
);
```

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/components/FrameMetadata.test.tsx
+++ b/src/components/FrameMetadata.test.tsx
@@ -78,4 +78,17 @@ describe('FrameMetadata', () => {
     ).toBe('10');
     expect(meta.container.querySelectorAll('meta').length).toBe(3);
   });
+
+  it('renders with wrapperComponent', () => {
+    const meta = render(
+      <FrameMetadata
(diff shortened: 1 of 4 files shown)


Zizzamia (approved):
Ok that's very interesting, I am going to merge it, polish it more this adternoon and than cut a release.

PR #96, opened 2024-02-07
[WIP] Add XMTP Support

neekolas:
**What changed? Why?**

- Added support for resolving POST requests coming from either XMTP clients OR Farcaster clients
- Added support for verifying signatures from XMTP clients and determining the `verifiedWalletAddress` field. This is done without any network requests, since XMTP POST payloads can be verified as a self-contained payload.

**Notes to reviewers**

This will need to wait on [a PR](https://github.com/xmtp/xmtp-node-js-tools/pull/143) in `@xmtp/frames-validator` to make it support being imported from CommonJS modules. That should land tomorrow.

**How has it been tested?**

### DIFF
--- a/src/utils/xmtp/validation.ts
+++ b/src/utils/xmtp/validation.ts
@@ -1,4 +1,4 @@
-import { FramePostPayload } from '@xmtp/frames-validator/dist/src/types';
+import { FramePostPayload } from '@xmtp/frames-validator';
 import { FrameValidationResponse } from '../../core/types';
 import { validateFramesPost } from '@xmtp/frames-validator';
 
--- a/src/utils/xmtp/validation.ts
+++ b/src/utils/xmtp/validation.ts
@@ -1,0 +1,28 @@
+import { FramePostPayload } from '@xmtp/frames-validator/dist/src/types';
+import { FrameValidationResponse } from '../../core/types';
+import { validateFramesPost } from '@xmtp/frames-validator';
+
+export function isXmtpFrameResponse(json: any): json is FramePostPayload {
+  return (
+    !!json?.untrustedData?.opaqueConversationIdentifier && !!json.trustedData?.messageBytes.length
+  );
+}
+
+export async function validateXmtpFrameResponse(
+  data: FramePostPayload,
+): Promise<FrameValidationResponse> {
+  try {
+    const { verifiedWalletAddress, actionBody } = await validateFramesPost(data);
+    return {
+      isValid: true,
+      message: {
+        ...actionBody,
+        verifiedWalletAddress,
+      },
+      clientType: 'xmtp',
+    };
+  } catch (e) {
+    console.error(e);
+    return { isValid: false, message: undefined };
+  }
+}
--- a/yarn.lock
+++ b/yarn.lock
@@ -705,6 +705,7 @@ __metadata:
     "@types/jest": "npm:^29.5.11"
     "@types/react": "npm:^18"
     "@types/react-dom": "npm:^18"
+    "@xmtp/frames-validator": "npm:^0.2.0"
     jest: "npm:^29.7.0"
     jest-environment-jsdom: "npm:^29.7.0"
     jest-extended: "npm:^4.0.2"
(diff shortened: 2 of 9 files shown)


yarn.lock
```diff
@@ -724,6 +725,415 @@ __metadata:
   languageName: unknown
   linkType: soft
 
+"@ethersproject/abi@npm:5.7.0, @ethersproject/abi@npm:^5.7.0":
```
Zizzamia, line 728: These are quite a lot of dependencies is taking in.

src/utils/xmtp/validation.ts
```diff
@@ -0,0 +1,28 @@
+import { FramePostPayload } from '@xmtp/frames-validator/dist/src/types';
+import { FrameValidationResponse } from '../../core/types';
+import { validateFramesPost } from '@xmtp/frames-validator';
```
Zizzamia, line 3: I wonder if this can be implemented independently. 

I am a bit worry of how big is `@xmtp/frames-validator` dependency for OnchainKit.

neekolas: Frames validator itself is very small, but it relies on `xmtp-js`. The good news is that we have someone already working on removing our dependency on ethers, which is the biggest dependency of that library.

Just for my own understanding, is there a plan to run `onchainkit` in browsers? Or is this going to remain purely for servers? If we want browser support, there are a few other changes I'd need to make to remove dependencies on `node:crypto`.

Zizzamia: For context, any feature in OnchainKit has to meet three bars:
- being able to run on Web client
- being able to run on latest Backend Node.js
- being able to run on React Native

Zizzamia: As OnchainKit has to work out of the box for all Coinbase products using these three case scenarios.

neekolas: OK. That's very good to know. We can work with those requirements. RN is the most challenging, since it doesn't have a built in `crypto` module or `wasm` support, but we have a RN SDK so none of this is new to us.

I'll get back to you on a new timeline with those requirements in mind.

neekolas: I'm curious if you'd considered breaking Frame Kit into a separate library. All the current functionality only makes sense on a server, and a lot of it is specifically designed around Next.js. Would be easier to maintain and develop without the additional requirements of web/RN support.

Zizzamia: > I'm curious if you'd considered breaking Frame Kit into a separate library.

If we do that, it will make harder for us support it and make sure is well adopted.

Zizzamia: > All the current functionality only makes sense on a server, and a lot of it is specifically designed around Next.js

Yeah that's fair. I think we did a lot of simplification with our Frames approach by using a third-party provider for validating messages.

Zizzamia: Part of me is like, we should create a https://github.com/Zizzamia/a-frame-in-100-lines version that showcase how to use your library in backend for node.js and just call it the day.

After all OnchainKit is a client library, and it avoids the backend part as much as possible.

Zizzamia: Or in some way, OnchainKit knows how to recognize if is XMTP, an that's it, if that's the case we advice to use your library to complete other parts.

Zizzamia: You see what I mean, I think it's fair to find a way to make the two working well together, instead of trying to force OnchainKit to be something is not.

neekolas: These are good points. I'm going to take a step back and try again with a more abstract approach, where the validation logic and cryptographic libraries don't have to be embedded in `onchainkit`. That avoids the dependency issues and library bloat, plus it lets you potentially support more than just XMTP Frames.

Zizzamia:
Closing this, in favour of https://github.com/coinbase/onchainkit/pull/116

Issue #97, opened 2024-02-07
Bug: Old dependencies and build fails


FedererKK:
### Describe the bug and the steps to reproduce it

A fresh install of @coinbase/onchainkit with this sample frame fails do to missing libraries when building

Perhaps they should be added as dependencies?

page.tsx

  ```
import { FrameMetadata } from "@coinbase/onchainkit";

export default function Home() {
  return (
    <FrameMetadata
      buttons={[
        {
          label: "Tell me the story",
        },
        {
          action: "link",
          label: "Link to Google",
          target: "https://www.google.com",
        },
        {
          action: "post_redirect",
          label: "Redirect to cute pictures",
        },
      ]}
      image="https://zizzamia.xyz/park-1.png"
      input={{
        text: "Tell me a boat story",
      }}
      post_url="https://zizzamia.xyz/api/frame"
    />
  );
}

```

  ```
 npm ERR! Missing: viem@2.7.6 from lock file
 npm ERR! Missing: @adraffy/ens-normalize@1.10.0 from lock file
 npm ERR! Missing: @noble/curves@1.2.0 from lock file
 npm ERR! Missing: @noble/hashes@1.3.2 from lock file
 npm ERR! Missing: @scure/bip32@1.3.2 from lock file
 npm ERR! Missing: @scure/bip39@1.2.1 from lock file
 npm ERR! Missing: abitype@1.0.0 from lock file
 npm ERR! Missing: isows@1.0.3 from lock file
 npm ERR! Missing: ws@8.13.0 from lock file
 npm ERR! Missing: @scure/base@1.1.5 from lock file
```

Zizzamia:
https://github.com/coinbase/onchainkit/releases/tag/v0.5.2 fix this.

Thank you @FedererKK

PR #98, opened 2024-02-07
fix: Update viem dependency version to 2.7.0

FedererKK:
**What changed? Why?**
Bumping viem version to 2.7.0 fixes ISSUE #97 

**Notes to reviewers**
Only changed viem

**How has it been tested?**
Fresh install and build is now succesfull
image
image


Zizzamia:
@FedererKK thank you for the fix.
Btw would you mind run `yarn format` it seems a Format check is failing.

PR #100, opened 2024-02-07
docs: initial docs

n3ps:
**What changed? Why?**
Initial docs using ~~VitePress~~ ~~Vocs~~ VitePress

**Notes to reviewers**
To run: `yarn docs:dev`

**How has it been tested?**
Localhost


n3ps:
Vocs issues:
* doesn't like type=commonjs
* switching to ESM breaks jest (if we're using Vite might as well use Vitest?)
* also breaks `format`
* and crashes in Windows

Zizzamia (approved):
BRILLIANT

Issue #102, opened 2024-02-08
Feature Request: make type parameters use consistent lowerCamelCase


spennyp:
### Describe the solution you'd like

A nit, but currently type parameters mix snake_case and lowerCamelCase.

For example:

snake_case (post_url, refresh_period):
```ts
export type FrameMetadataType = {
    buttons?: [FrameButtonMetadata, ...FrameButtonMetadata[]];
    image: string;
    input?: FrameInputMetadata;
    post_url?: string;
    refresh_period?: number;
};
```

lowerCamelCase (isValid): 
```ts
export type FrameValidationResponse = {
    isValid: true;
    message: FrameValidationData;
} | {
    isValid: false;
    message: undefined;
};
```

Typescript generally uses lowerCamelCase here: 
* https://www.typescriptlang.org/docs/handbook/typescript-tooling-in-5-minutes.html#interfaces
* https://google.github.io/styleguide/tsguide.html#naming-rules-by-identifier-type

Zizzamia:
Yeah, that's a big miss from my side! you right!

I wonder when we should do this breaking changes, if today or in a few weeks when things cool down.

@spennyp thoughts?

spennyp:
Probably wait to bundle it with other breaking changes in a major release makes sense

Zizzamia:
Yeah, right a lot of people are shipping some cool Frames, so I think this can be a nice breaking changes to do on Monday next week.

Zizzamia:
@spennyp check `v0.5.3` https://github.com/coinbase/onchainkit/releases/tag/v0.5.3 
Has a couple of things you requested.

PR #106, opened 2024-02-08
feat: made all `FrameMetadataType` parameters use consistent lowerCamelCase

Zizzamia:
**What changed? Why?**
Related to https://github.com/coinbase/onchainkit/issues/102

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/.changeset/clever-maps-marry.md
+++ b/.changeset/clever-maps-marry.md
@@ -1,5 +1,5 @@
 ---
-"@coinbase/onchainkit": patch
+'@coinbase/onchainkit': patch
 ---
 
-- **feat**: ok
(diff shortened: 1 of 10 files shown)


.changeset/clever-maps-marry.md
```diff
@@ -0,0 +1,6 @@
+---
+'@coinbase/onchainkit': patch
+---
+
+- **feat**: all `FrameMetadataType` parameters have been updated to use consistent lowerCamelCase. It's important to note that we are not deprecating the old method, at least for a few weeks, to allow time for migration to the new approach. By @zizzamia #106
+- **feat**: while there is no real issue in using either `property` or `name` as the standard for a metadata element, it is fair to respect the Open Graph specification, which originally uses `property`. By @zizzamia #106
```
cnasc, line 6: +1 on this, per [the spec ](https://docs.farcaster.xyz/reference/frames/spec#properties) it should be property, and there are probably clients building exactly to the spec and not considering other meta tag conventions

Zizzamia: 😅

PR #110, opened 2024-02-09
Add support for image aspect ratio metadata

taycaldwell:
**What changed? Why?**
Add support for optional metadata tag `fc:frame:image:aspect_ratio`

See: https://docs.farcaster.xyz/reference/frames/spec

**Screenshot**

Aspect Ratio: "1:1"

Screenshot 2024-02-08 at 7 56 20 PM

Screenshot 2024-02-08 at 7 55 05 PM

**Notes to reviewers**
Backwards compatible. Image metadata can be defined as a string (defaults to aspectRatio of `1.91:1`) or an object with a `src` URL string and `aspectRatio` of `1.91:1` or `1:1`.

**How has it been tested?**
Manually

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ export default function HomePage() {
           label: 'Redirect to cute pictures',
         },
       ]}
-      image="https://zizzamia.xyz/park-1.png"
+      image='https://zizzamia.xyz/park-1.png'
       input={{
         text: 'Tell me a boat story',
       }}
@@ -116,7 +116,7 @@ export default function HomePage() {
 **@Props**
 
 ```ts
-type Button =
+type ButtonMetadata =
   | {
       action: 'link' | 'mint';
       label: string;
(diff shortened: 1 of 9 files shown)


docs/frame-kit.md
```diff
@@ -168,16 +168,21 @@ export default function Page() {
...
 
 ```ts
-type Button = {
+type ButtonMetadata = {
   label: string;
   action?: 'post' | 'post_redirect';
 };
 
+type ImageMetadata = {
+  src: string;
```
Zizzamia, line 177: I am probably going to change this to `target` in the next PR, just to keep it clean with the other APIs.

README.md
```diff
@@ -218,11 +223,16 @@ type InputMetadata = {
   text: string;
 };
 
+type ImageMetadata = {
+  src: string;
```
Zizzamia, line 227: Thoughts on this one @spennyp ?
I am thinking to set this to `target` instead, so there is some balance with the new `link` and `mint` actions.

Or we think `src` is better for an image?

docs/frame-kit.md
Zizzamia: Or maybe not 🤔

Zizzamia (approved):
LEGEND

PR #114, opened 2024-02-09
feat: Get the EVM address for a given Farcaster User

Sneh1999:
**What changed? Why?**

Ability to extract the EVM address for a given Farcaster user

It gives the client ability to either get the verified address or the custody address

**Notes to reviewers**

**How has it been tested?**
`yarn test`
`yarn integration:test`


src/core/getFarcasterUserAddresses.test.ts
```diff
@@ -0,0 +1,90 @@
+import { neynarGetCustodyAddressesForFid } from '../utils/neynar/frame/neynarGetCustodyAddressForFid';
```
Zizzamia, line 1: I will say, let's start creating the `src/farcaster` folder, so we can play with the namespaces idea we talk about

Sneh1999: makes sense! updated this

src/farcaster/getFarcasterUserAddresses.ts
```diff
@@ -0,0 +1,35 @@
...
+  try {
+    const { fid, farcasterAddressType } = getFarcasterUserAddressRequest;
+    if (farcasterAddressType === FarcasterAddressType.VerifiedAddresses) {
+      const addresses = await neynarGetVerifiedAddressesForFid(fid, options?.neynarApiKey);
+      return addresses;
+    }
+    const address = await neynarGetCustodyAddressesForFid(fid, options?.neynarApiKey);
+    return address;
+  } catch (e) {
+    throw e;
```
Zizzamia, line 31: The `throw e;` doesn't match with the title.

If this fails should return null?

src/farcaster/getFarcasterUserAddresses.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Get the user address for a given fid
+ * @param fid  The frame id
+ * @param FarcasterAddressType represents whether the client wants a verified address or a custody address
+ * @returns the validation addresses or the custory address. If there is an error, it throws an error
+ */
+
+async function getFarcasterUserAddresses(
+  getFarcasterUserAddressRequest: GetFarcasterUserAddressesRequest,
+  options?: GetUserAddressOptions,
+): Promise<string | string[]> {
```
Zizzamia, line 21: Why have a string in one case a string array in an other? This feels like it will make the coding experience a bit more challenging as someone has to check what type is it before using it.

src/farcaster/getFarcasterUserAddresses.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+  | undefined;
+
+/**
+ * Get the user address for a given fid
+ * @param fid  The frame id
+ * @param FarcasterAddressType represents whether the client wants a verified address or a custody address
+ * @returns the validation addresses or the custory address. If there is an error, it throws an error
+ */
+
```
Zizzamia, line 17: nit. remove the extra space

package.json
```diff
@@ -30,7 +30,7 @@
     "@changesets/cli": "^2.26.2",
     "@testing-library/jest-dom": "^6.4.0",
     "@testing-library/react": "^14.2.0",
-    "@types/jest": "^29.5.11",
+    "@types/jest": "^29.5.12",
```
Zizzamia, line 33: Dope,
also can you please run `yarn changeset` as patch and describe the change for this PR.

src/farcaster/types.ts
```diff
@@ -0,0 +1,19 @@
...
+ *  Note: exported as public Type
+ */
+
+export enum FarcasterAddressType {
+  VerifiedAddresses,
+  CustodyAddress,
+}
+
+/**
+ * GetFarcasterUserAddressesRequest
```
Zizzamia, line 12: From where the developer will get this? Like from a Frame is a message, in this case do we have an idea how this request is shaped?

src/farcaster/types.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * Note: exported as public Type
+ */
+export interface GetFarcasterUserAddressesRequest {
```
Zizzamia, line 16: Where can we see an example of this kind of request?

src/farcaster/getFarcasterUserAddresses.ts
```diff
@@ -0,0 +1,35 @@
...
+  options?: GetUserAddressOptions,
+): Promise<string | string[]> {
+  try {
+    const { fid, farcasterAddressType } = getFarcasterUserAddressRequest;
+    if (farcasterAddressType === FarcasterAddressType.VerifiedAddresses) {
+      const addresses = await neynarGetVerifiedAddressesForFid(fid, options?.neynarApiKey);
+      return addresses;
+    }
+    const address = await neynarGetCustodyAddressesForFid(fid, options?.neynarApiKey);
+    return address;
```
Zizzamia, line 29: I am lost why is this called twice?

src/utils/neynar/user/neynarGetVerifiedAddressesForFid.ts
```diff
@@ -0,0 +1,35 @@
+import { version } from '../../../version';
+import { FetchError } from '../exceptions/FetchError';
+import { NEYNAR_DEFAULT_API_KEY } from '../frame/neynarFrameFunctions';
+
+export async function neynarGetVerifiedAddressesForFid(
+  fid: number,
+  apiKey: string = NEYNAR_DEFAULT_API_KEY,
+): Promise<string[]> {
+  const options = {
```
Zizzamia, line 9: If we are starting making many Neynar call, it feels like we might want to abstract the creation of this optios.

Zizzamia:
@Sneh1999 can you share where in the Farcaster spec those are specified. Also how they are used.

In particular I am curios to understand from where the Request is coming from.

src/farcaster/types.ts
Sneh1999: This can be used to display the user's address.
In case of searchcaster they wrote the code [here](https://github.com/gskril/searchcaster/blob/03fb485d8dbe52a2248d3353dfc2f15164caa949/src/pages/profiles.js#L135) and [here](https://github.com/gskril/searchcaster/blob/03fb485d8dbe52a2248d3353dfc2f15164caa949/src/pages/api/profiles.js#L141) to show the user's profile

src/farcaster/getFarcasterUserAddresses.ts
Sneh1999: Makes sense! I will update this to return string

src/utils/neynar/user/neynarGetVerifiedAddressesForFid.ts
Sneh1999: okay makes sense! I will abstract this part

src/farcaster/getFarcasterUserAddresses.ts
Sneh1999: sounds good!

Sneh1999:
> @Sneh1999 can you share where in the Farcaster spec those are specified. Also how they are used.
> 
> In particular I am curios to understand from where the Request is coming from.

Hmm, Are you asking what `Verified address` and `custody address` are?

I found [this](https://github.com/farcasterxyz/protocol/blob/b535172081fc7cb37503a3ef4bf62ca980721e6f/docs/SPECIFICATION.md?plain=1#L365) about verifications in the protocol docs
and this about [custody address](https://github.com/farcasterxyz/protocol/blob/b535172081fc7cb37503a3ef4bf62ca980721e6f/docs/SPECIFICATION.md?plain=1#L22)


Custody address is used to check if a given [message is valid or not](https://github.com/farcasterxyz/protocol/blob/b535172081fc7cb37503a3ef4bf62ca980721e6f/docs/SPECIFICATION.md?plain=1#L500). A FID is connected to a custody address. 

Verified address on the other hand can be multiple like `snehkoul.eth` or `zizzamia.eth`. You can add as many verified addresses to your account as you want but you need to verify that you indeed hold the private keys for these addresses.

src/farcaster/types.ts
Zizzamia: Ok so is this the body we are getting?
```
body: {
        id: p.id,
        address: p?.owner || null,
        username: p.username,
        displayName: p.display_name,
        bio: p.bio,
        followers: p.followers,
        following: p.following,
        avatarUrl: p.avatar_url,
        isVerifiedAvatar: p.avatar_verified,
        registeredAt: new Date(p.registered_at).getTime(),
      },
      connectedAddress:
        p.verifications && p.verifications[0]
          ? p.verifications[0].address
          : connected_address || null,
      connectedAddresses:
        p.verifications && p.verifications[0]
          ? p.verifications.map((v) => v.address)
          : [],
```

and from where this request is coming?

I guess I am still not sure the end-to-end flow. Like:
Step 1. the body comes from this scenario
Step 2. we use the body in this way
Step 3. we have the address

.changeset/thin-windows-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@coinbase/onchainkit': patch
+---
+
+Adding `getFarcasterUserAddress` utility to extract user's custody or verified address. This PR also updates the version of `@types/jest` package
```
Zizzamia, line 5: Nice, also as something is already added, it's correct to use the past tense.

```
- **feat**: added ~~~
```

Zizzamia: > This PR also updates the version of `@types/jest` package

this is a separete line, as this is documetning atomic changes, not PRs.

```
- **chore**: upddated ~~~~
```

src/farcaster/getFarcasterUserAddress.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Get the user address for a given fid
+ * @param fid  The frame id
+ * @param FarcasterAddressType represents whether the client wants a verified address or a custody address
+ * @returns the validation addresse or the custory address. If there is an error, it returns null
+ */
+async function getFarcasterUserAddress(
+  getFarcasterUserAddressRequest: GetFarcasterUserAddressRequest,
+  options?: GetUserAddressOptions,
+): Promise<string | null> {
```
Zizzamia, line 20: I think we can return all addresses, and have the return type being  `Address[] | null`.
Also if this is an address we should probably use the Address type from Viem.

src/farcaster/types.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * Frame address type to determine if the client is requesingt  verified address or a custody address
+ *  Note: exported as public Type
+ */
+
```
Zizzamia, line 5: nit. extra line

src/farcaster/types.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+/**
+ * GetFarcasterUserAddressRequest
+ *
+ * Note: exported as public Type
+ */
+export interface GetFarcasterUserAddressRequest {
+  fid: number;
+  farcasterAddressType: FarcasterAddressType;
```
Zizzamia, line 18: I am still confused where this type is coming from. Like do we ask someone to shape the data like this, or actually is coming like this from an API

Zizzamia: Like is `GetFarcasterUserAddressRequest` and API from somewhere?

src/index.ts
```diff
@@ -1,6 +1,7 @@
 // 🌲☀️🌲
 export { version } from './version';
 export { getFrameHtmlResponse } from './core/getFrameHtmlResponse';
+export { getFarcasterUserAddress } from './farcaster/getFarcasterUserAddress';
```
Zizzamia, line 4: If we want to start doing namespaces, than the `index.ts` to use is `/farcster/index.ts`. 

And then we will have folks doing the long import
```ts
import {  getFarcasterUserAddress } from '@coinbase/onchainkit/farcaster';
```

Zizzamia: Same for the types.

src/utils/neynar/neynar.ts
```diff
@@ -0,0 +1,44 @@
+import { version } from '../../version';
+import { FetchError } from './exceptions/FetchError';
+import { NEYNAR_DEFAULT_API_KEY } from './frame/neynarFrameFunctions';
+
+export async function fetchDataFromNeynar(url: string, apiKey: string = NEYNAR_DEFAULT_API_KEY) {
```
Zizzamia, line 5: Split each function in their file.
It will help organize them better and over time it's easier to mantain.

The reason we want files with the same name of the function, is that when you are looking around the folder it's intuitive to see who does what and where.

Those litle adjustemnts make an OSS library easier to mantain and understand from the outside.

Zizzamia: On the naming you can keep this as `getData` and `postData`. So it's easy to tell which is which, for a second I read `fetch` and I thought was a funciton that does both get and post.

src/farcaster/types.ts
Sneh1999: its not an API, its something we will ask the user to provide us

Sneh1999: `we ask someone to shape the data like this` -  this

src/farcaster/types.ts
Sneh1999: We created this type to ask the user whether they want a given user's `custody` or `verified address`
For verified address, we are using the api [here](https://docs.neynar.com/reference/verifications-v1)
For custody address, we are using the api [here](https://docs.neynar.com/reference/custody-address-v1)

What I was pointing to in `searchcaster` is the use case of `getFasrcasterUserAddress`. 

So if I am a `searchcaster` dev and I want to use `onchainkit` to get a user's address to build a profile object. I can do this (sudo code)

```
const userProfile =  {
 custodyAddress: getFarcasterUserAddress({fid:12, FarcasterAddressType.CustodyAddress}, {neynarAPIKey: '123'})
 verifiedAddresses: getFarcasterUserAddress({fid:12, FarcasterAddressType.VerifiedAddresses})
}
```

src/farcaster/types.ts
Sneh1999: Let me know if this makes sense:
https://github.com/coinbase/onchainkit/pull/114#discussion_r1485263295

src/farcaster/getFarcasterUserAddress.ts
Sneh1999: Do you mean we should return both custody and verified addresses in the same call?

src/farcaster/getFarcasterUserAddress.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Get the user address for a given fid
+ * @param fid  The farcaster id
+ * @param GetFarcasterUserAddressOptions  The options to specify the type of addresses to get and the neynar api key
+ * @returns the custory address and/or verified addresses. If there is an error, it returns null
+ */
+async function getFarcasterUserAddress(
+  fid: number,
+  options?: GetFarcasterUserAddressOptions,
+): Promise<GetFarcasterUserAddressResponse | null> {
+  try {
```
Zizzamia, line 23: @Sneh1999 side not for the next PR, large try/catch are a bad practice in Javascript, as each Try/Catch will hold on memory almost everything inside a `try` till the end of the execution. It's advice to have smaller try/catch.

src/farcaster/getFarcasterUserAddress.ts
```diff
@@ -0,0 +1,48 @@
...
+    const hasCustodyAddresses = options?.hasCustodyAddresses ?? true;
```
Zizzamia, line 24: `hasCustodyAddresses` @Sneh1999 did you mentioned there can be only one address for custody?

Zizzamia: So why not `hasCustodyAddress` as singular?

Zizzamia (approved):
Niiiice!!!!

Great work @Sneh1999

PR #116, opened 2024-02-09
[WIP] Add Open Frames Neynar validator

neekolas:
**What changed? Why?**

This PR adds a proof-of-concept for support of Open Frames. Opening it up for discussion. It shows a path for how `onchainkit` can be extended to support multiple types of Frames client beyond just Farcaster.

The PR adds no new runtime dependencies. It just picks up some common types from [`@open-frames/types`](https://www.npmjs.com/package/@open-frames/types).

**Notes to reviewers**

If we decide this approach is reasonable, I'll add some tests. And we can discuss how we can add some better examples for extensibility.

**How has it been tested?**

Not at all

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
   "devDependencies": {
     "@changesets/changelog-github": "^0.4.8",
     "@changesets/cli": "^2.26.2",
+    "@open-frames/types": "^0.0.5",
     "@testing-library/jest-dom": "^6.4.0",
     "@testing-library/react": "^14.2.0",
     "@types/jest": "^29.5.11",
--- a/src/utils/neynar/openframes/neynarValidator.ts
+++ b/src/utils/neynar/openframes/neynarValidator.ts
@@ -1,0 +1,54 @@
+import { OpenFramesRequest, RequestValidator, ValidationResponse } from '@open-frames/types';
+import { FarcasterOpenFramesRequest, FrameRequest, FrameValidationData } from '../../../core/types';
+import { NEYNAR_DEFAULT_API_KEY, neynarFrameValidation } from '../frame/neynarFrameFunctions';
+import { Options } from './types';
+
+export class NeynarValidator
+  implements RequestValidator<FarcasterOpenFramesRequest, FrameValidationData, 'farcaster'>
+{
+  readonly protocolIdentifier = 'farcaster';
+  private options: Required<Options>;
+
+  constructor(options?: Options) {
+    this.options = {
+      neynarApiKey: options?.neynarApiKey || NEYNAR_DEFAULT_API_KEY,
+      castReactionContext: options?.castReactionContext || true,
+      followContext: options?.followContext || true,
+    };
+  }
+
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -28,6 +28,7 @@
   "devDependencies": {
     "@changesets/changelog-github": "^0.4.8",
     "@changesets/cli": "^2.26.2",
+    "@open-frames/types": "^0.0.5",
```
Zizzamia, line 31: Veeeery interesting

src/utils/neynar/openframes/neynarValidator.ts
```diff
@@ -0,0 +1,54 @@
+import { OpenFramesRequest, RequestValidator, ValidationResponse } from '@open-frames/types';
+import { FarcasterOpenFramesRequest, FrameRequest, FrameValidationData } from '../../../core/types';
+import { NEYNAR_DEFAULT_API_KEY, neynarFrameValidation } from '../frame/neynarFrameFunctions';
+import { Options } from './types';
+
```
Zizzamia, line 5: Oh side note, we prefer Functional code vs Object Oriented, as it is much lighter in terms of memory and also easier to write test for.

src/utils/neynar/openframes/neynarValidator.ts
```diff
@@ -0,0 +1,54 @@
...
+    return `${this.protocolIdentifier}@VNext`;
+  }
+
+  isSupported(payload: OpenFramesRequest): payload is FarcasterOpenFramesRequest {
+    const isCorrectClientProtocol =
+      !!payload.clientProtocol && payload.clientProtocol.startsWith('farcaster@');
+    const isTrustedDataValid = typeof payload.trustedData?.messageBytes === 'string';
+
+    return isCorrectClientProtocol && isTrustedDataValid;
+  }
```
Zizzamia, line 30: So I see, `isSupported` is where the magic happens.

And ideally we can have as many client as we want.

There are some implementation details that we need to align, but I like where your head is @neekolas

Zizzamia:
@neekolas I think this can work, after all we are just simply saying, let's see what Client can validate this message, and when there is a match, we let that client connect with a third-party API and take care of it.

src/utils/neynar/openframes/neynarValidator.ts
neekolas: Fair enough. All I really care about is having something that satisfies a common interface so that developers can build their own request validators and have confidence they will be able to integrate.

Whether the implementation of the interface is a module, an object, or a class is up to the implementer.

src/utils/neynar/openframes/neynarValidator.ts
neekolas: Exactly. [Here is an example of the XMTP RequestValidator](https://github.com/xmtp/xmtp-node-js-tools/pull/147) that can be safely passed in to `getOpenFramesMessage` as an argument and will allow for processing of button clicks coming from XMTP.

PR #122, opened 2024-02-13
feat: new function `getEASAttestations` to fetch attestation from Ethereum Attestation Service (EAS) - part 1

alvaroraminelli:
**What changed? Why?**

The new `getEASAttestations` function fetches Ethereum Attestation Service (EAS) attestations for a given address and chain. Optionally filtered by schemas associated with the attestation.

Example:

```javascript
import { getEASAttestations } from '@coinbase/onchainkit'
import { base } from "viem/chains";

const attestations = await getEASAttestations("0x1234567890abcdef1234567890abcdef12345678", base)
// returns [
//   {
//       "attester": "0x357458739F90461b99789350868CD7CF330Dd7EE",
//       "expirationTime": 0,
//       "id": "0x93016a60f13e7cfe0257116aedfce7088c2c0020787a325ea9f6b4ba11d07598",
//       "recipient": "0x44a7D120beA87455071cebB841eF91E6Ae21bC1a",
//       "revocationTime": 0,
//       "schemaId": "0x1801901fabd0e6189356b4fb52bb0ab855276d84f7ec140839fbd1f6801ca065",
//       "timeCreated": 1707269100,
//       "txid": "0x88448267566c9546ff31b9e6be229fb960f12bec8bc441259c7b064ae4159d34"
//   },
// ]
```

**Notes to reviewers**

This is Part 1 of the feature implementation, which aims to validate the function APIs and EAS chain definitions for the blockchains we intend to support. Part 2 will focus on the actual implementation, including fetching data from GraphQL.

**How has it been tested?**

- Unit Tests


src/core/types.ts
```diff
@@ -126,3 +126,45 @@ export type FrameMetadataType = {
  * Note: exported as public Type
  */
 export type FrameMetadataResponse = Record<string, string>;
+
+/**
+ * EAS Attester Address
+ * The Ethereum address of the attester who created the attestation.
+ */
+type EASAttesterAddress = `0x${string}`;
```
Sneh1999, line 134: Could this type be `Address` type from viem?

src/core/types.ts
```diff
@@ -126,3 +126,45 @@ export type FrameMetadataType = {
...
+
+/**
+ * EAS Attestation
+ * GraphQL response for EAS Attestation
+ * Note: exported as public Type
+ */
+export type EASAttestation = {
+  id: string; // The unique identifier of the attestation.
+  decodedDataJson: string; // The attestation data decoded to JSON.
+  recipient: string; // The Ethereum address of the recipient of the attestation.
```
Sneh1999, line 151: should this be an `Address` type from viem?

src/core/getEASAttestations.ts
```diff
@@ -0,0 +1,57 @@
...
+      throw new Error(
+        `Chain is not supported. Supported chains: ${Object.keys(easSupportedChains).join(', ')}`,
+      );
+    }
+
+    // TODO: Implement the function that get the EAS Attestations in part 2
+
+    return [];
+  } catch (error) {
+    throw new Error(`Error in getEASAttestation: ${(error as Error).message}`);
```
Sneh1999, line 55: @Zizzamia  suggested on my last PR related to `getFarcasterUserAddress` that we should return null, instead of throwing an error. Not sure, if we should maintain a pattern and return null here as well?

src/definitions/base.ts
```diff
@@ -0,0 +1,14 @@
+import { base } from 'viem/chains';
+import { EASChainDefinition } from '../core/types';
+
+export const easChainBase: EASChainDefinition = {
+  id: base.id,
+  easGraphqlAPI: 'https://base.easscan.org/graphql',
+  schemaUids: [
+    // VERIFIED_COUNTRY
```
Sneh1999, line 8: Could we add some more info on `VERIFIED COUNTRY` and account is here?

src/definitions/optimism.ts
```diff
@@ -0,0 +1,23 @@
...
+import { EASChainDefinition } from '../core/types';
+
+// More details in https://docs.optimism.io/chain/identity/schemas
+export const easChainOptimism: EASChainDefinition = {
+  id: optimism.id,
+  easGraphqlAPI: 'https://optimism.easscan.org/graphql',
+  schemaUids: [
+    // N_A
+    '0xac4c92fc5c7babed88f78a917cdbcdc1c496a8f4ab2d5b2ec29402736b2cf929',
+    // GITCOIN_PASSPORT_SCORES_V1:
```
Sneh1999, line 11: similarly for this, maybe a link? not sure

src/definitions/base.ts
alvaroraminelli: what kind of info?

src/core/getEASAttestations.ts
Sneh1999: yea, thats a good idea!

src/core/types.ts
```diff
@@ -126,3 +126,45 @@ export type FrameMetadataType = {
...
+/**
+ * Ethereum Attestation Service (EAS) Chain Definition
+ * The definition of a blockchain chain supported by EAS attestations.
+ * Note: exported as public Type
+ */
+export type EASChainDefinition = {
+  id: number; // blockchain source id
+  easGraphqlAPI: string; // EAS GraphQL API endpoint
+  schemaUids: EASSchemaUid[];
+  attesterAddresses: EASAttesterAddress[];
```
david-roth-cb, line 169: nit probably an edge case or something to worry about in the future, but a user may not always want all attestations for a given recipient from all attesters on that chain. 

ex: Gitcoin and Coinbase both make attestations against a generalized {verified: true} schema on Base. as a user, I may only want to know if a given recipient has attestations for that schema that were attested to by the coinbase attester and not the gitcoin attester.

src/core/getEASAttestations.test.ts
```diff
@@ -0,0 +1,27 @@
...
+  const mockAddress = '0x1234567890abcdef1234567890abcdef12345678';
+  const mockOptions: GetEASAttestationsOptions = { schemas: ['0x12345'] };
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('throws an error for unsupported chains', async () => {
+    await expect(getEASAttestations(mockAddress, opBNBTestnet, mockOptions)).rejects.toThrow(
+      `Chain is not supported. Supported chains: ${Object.keys(easSupportedChains).join(', ')}`,
```
Zizzamia, line 19: Side note, we need to standardize the way we have `toThrow` in OnchainKit.

@alvaroraminelli can you have an Issue on this. This is more a nit for later.

src/core/getEASAttestations.ts
Zizzamia: I guess there is an interesting point of:
1. this is just a BAD coding, like using the wrong chain

vs

2. you are doing all good, it's just empty so returns `[]` or `null`

so I think going forward the `throw` is where we want to say OUT LOUD, this is NOT ALLOWED

vs

hey all good, you just got an empty result.

src/core/getEASAttestations.ts
```diff
@@ -0,0 +1,57 @@
+import { isChainSupported, easSupportedChains } from '../utils/easAttestation';
+import { EASAttestation, EASSchemaUid } from './types';
+import type { Address, Chain } from 'viem';
+
+export type GetEASAttestationsOptions = {
+  schemas?: EASSchemaUid[];
+};
+
+type GetEASAttestationsResponse = EASAttestation[];
```
Zizzamia, line 9: I like this

src/core/getEASAttestations.ts
```diff
@@ -0,0 +1,57 @@
...
+export async function getEASAttestations<TChain extends Chain>(
+  // @ts-ignore - WIP - Add support for address in part 2
+  address: Address,
+  chain: TChain,
+  // @ts-ignore - WIP - Add support for options in part 2
+  options?: GetEASAttestationsOptions,
+): Promise<GetEASAttestationsResponse> {
+  try {
+    if (!isChainSupported(chain)) {
+      throw new Error(
```
Zizzamia, line 46: I like this `throw`

src/core/getEASAttestations.ts
```diff
@@ -0,0 +1,57 @@
...
+        `Chain is not supported. Supported chains: ${Object.keys(easSupportedChains).join(', ')}`,
+      );
+    }
+
+    // TODO: Implement the function that get the EAS Attestations in part 2
+
+    return [];
+  } catch (error) {
+    throw new Error(`Error in getEASAttestation: ${(error as Error).message}`);
```
Zizzamia, line 55: So I like the first `throw`, but I am not conviced on the second.

Mostly because the fact there is `throw` in production is quite alarming to me as it is something that folks might miss and completely stop an application.

Again I think bad coding needs a `throw`, but bad responses, should have a soft lending, and than if something is odd, folks can just check the API on the network tab.

Ot we can raise a `console.error`.

src/core/types.ts
```diff
@@ -126,3 +126,45 @@ export type FrameMetadataType = {
...
+ * The Ethereum address of the attester who created the attestation.
+ */
+type EASAttesterAddress = Address;
+
+/**
+ * Ethereum Attestation Service (EAS) Schema Uid
+ * The schema identifier associated with the EAS attestation.
+ * Note: exported as public Type
+ */
+export type EASSchemaUid = `0x${string}`;
```
Zizzamia, line 141: Open question should this be `Address`, and if not, maybe we should write that in the comment.

src/core/types.ts
```diff
@@ -126,3 +126,45 @@ export type FrameMetadataType = {
...
+
+/**
+ * Ethereum Attestation Service (EAS) Attestation
+ * GraphQL response for EAS Attestation
+ * Note: exported as public Type
+ */
```
Zizzamia, line 147: nit. we should have it in alphabetical order.

Zizzamia: Also, is there a place where this is shared as standard? or is more a format we decided?

src/definitions/base.ts
```diff
@@ -0,0 +1,19 @@
+import { base } from 'viem/chains';
+import { EASChainDefinition } from '../core/types';
+
+export const easChainBase: EASChainDefinition = {
```
Zizzamia, line 4: I guess Sepolia will be added in the next PR?

Zizzamia (approved):
Great first PR,
we can follow up on the notes left by the team in the next one.

PR #128, opened 2024-02-16
DRAFT: added upload to IPFS support

polluterofminds:
**What changed? Why?**
Added a method to allow frame developers to upload media to IPFS and use the URL in their frames. 

**Notes to reviewers**

**How has it been tested?**
Yes


package.json
```diff
@@ -26,6 +26,7 @@
   "devDependencies": {
     "@changesets/changelog-github": "^0.4.8",
     "@changesets/cli": "^2.26.2",
+    "@farcaster/hub-nodejs": "^0.11.0",
```
Zizzamia, line 29: Is this a peer dependency?

Also this library is quite massive https://bundlephobia.com/package/@farcaster/hub-nodejs@0.11.0

As OnchainKit is used internally at Coinbase, we have strict requirements to support both web, react native and node. And keep the dependency and kbs of each feature extremely lean.

Can we operarte the sam result but without this library? Or maybe with a different one?

src/core/uploadFrameMediaToIPFS.ts
```diff
@@ -0,0 +1,50 @@
+import fs from 'fs';
```
Zizzamia, line 1: Are we looking to only support this to `node`?

Zizzamia:
> Added a method to allow frame developers to upload media to IPFS and use the URL in their frames.

To upload a media file, I guess the file is already online? and we want to move it from a web2 location to IPFS?

Also beside Frames, where other we see this experience working out well?

package.json
polluterofminds: This was a mistake. I meant to remove it. Will do that in the next push.

src/core/uploadFrameMediaToIPFS.ts
polluterofminds: What runtimes does OnchainKit support? I can try to make this general purpose.

polluterofminds:
> > Added a method to allow frame developers to upload media to IPFS and use the URL in their frames.
> 
> To upload a media file, I guess the file is already online? and we want to move it from a web2 location to IPFS?
> 
> Also beside Frames, where other we see this experience working out well?

In this PR, the media is not already online, it is from a local path in the developer's project. Using Next.js as an example, the goal is to get this media on IPFS instead of having it locked up in Vercel/S3. It's a baby step toward open media, but that's the over-arching goal of this PR. 

The biggest example of this working out well is NFTs. Much like frames, NFTs started with media almost entirely hosted on S3 and centralized storage systems. People quickly realized that was a mistake and the entire ecosystem virtually all pivoted to IPFS.

src/core/uploadFrameMediaToIPFS.ts
```diff
@@ -0,0 +1,50 @@
...
+};
+
+/**
+ * Take a path string, read the file, and upload it to IPFS
+ *
+ * If the path is valid and upload is successful, return the IPFS gateway URL for use in the frame, otherwise, return an error.
+ *
+ * @param mediaOptions The JSON that includes the path string, the Pinata JWT, and the Gateway URL to use.
+ */
+async function uploadFrameMediaToIPFS(
```
Zizzamia, line 16: @cnasc @Yuripetusko what's your take on this utility?

I am reflecting if it's worth to have something that helps in general upload things to IPFS.

Yuripetusko: I think it's good, my only comment would be that there should be options for other providers (filebase, nftstorage, etc)

Yuripetusko: Ah actually just noticed that this is part of onchainkit. This should always be a backend code due to api key exposure and formadata handling, I am wondering if `build-onchain-apps` is a better place, and treat it more like an example to copy

Zizzamia:
Closing this for now, we will re-evaluate IPFS support later on

PR #137, opened 2024-02-18
feat: basic frame debugger

cnasc:
**What changed? Why?**
Added a new directory, `framegear`, at top level. This directory contains a standalone Next.js app which can validate the initial frame tags -- additional debugging and other features will be added in future PRs.

We'll also want to determine how we want to ship this (e.g., if you globally install `onchainkit` does this become an available command in your path?) in the near future.

**Notes to reviewers**

**How has it been tested?**

https://github.com/coinbase/onchainkit/assets/9300702/c6145140-f3f4-4944-8f58-f468c12c2bde


framegear/.gitignore
```diff
@@ -0,0 +1,36 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
```
Zizzamia, line 2: nit. for next PR, do we need this? My guess is not

framegear/components/Header/Header.tsx
```diff
@@ -0,0 +1,54 @@
+import { APP_NAME, BORDER_COLOR, MAX_WIDTH } from '@/utils/constants';
+import Link from 'next/link';
+
+export function Header() {
+  return (
+    <div className={`flex w-full flex-col items-center gap-8 border-b ${BORDER_COLOR} py-8`}>
```
Zizzamia, line 6: nit. next PR start using `clsx` to better organize tailwind css

framegear/.gitignore
cnasc: ah probably not, was autogenerated. will clean up on the next one

framegear/tailwind.config.ts
```diff
@@ -0,0 +1,20 @@
+import type { Config } from "tailwindcss";
+
+const config: Config = {
+  content: [
+    "./pages/**/*.{js,ts,jsx,tsx,mdx}",
```
Zizzamia, line 5: I believe we don't have `pages` so this could be removed

Zizzamia (approved):

framegear/.gitignore
cnasc: following up on this, worth including because after removing it there are ~1000 files git tries to track. Updating the outer `.gitignore` with this stuff would probably be confusing so going to keep it

PR #148, opened 2024-02-20
wip: debugging

cnasc:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/framegear/yarn.lock
+++ b/framegear/yarn.lock
@@ -438,6 +438,18 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@coinbase/onchainkit@portal:./..::locator=framegear%40workspace%3A.":
+  version: 0.0.0-use.local
+  resolution: "@coinbase/onchainkit@portal:./..::locator=framegear%40workspace%3A."
+  peerDependencies:
+    graphql: ^14
+    graphql-request: ^6
+    react: ^18
+    react-dom: ^18
+    viem: ^2.7.0
+  languageName: node
+  linkType: soft
+
 "@cspotcode/source-map-support@npm:^0.8.0":
   version: 0.8.1
   resolution: "@cspotcode/source-map-support@npm:0.8.1"
@@ -3179,6 +3191,7 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "framegear@workspace:."
   dependencies:
+    "@coinbase/onchainkit": "portal:./.."
     "@radix-ui/react-icons": "npm:^1.3.0"
     "@testing-library/jest-dom": "npm:^6.4.2"
     "@testing-library/react": "npm:^14.2.1"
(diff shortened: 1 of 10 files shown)


src/index.ts
```diff
@@ -1,5 +1,6 @@
 // 🌲☀️🌲
 export { version } from './version';
+export { getDebugFrameRequest, type DebugFrameRequestOptions } from './core/getDebugFrameRequest';
```
Zizzamia, line 3: All exported types, let's have it in `types.ts`

src/core/getDebugFrameRequest.ts
```diff
@@ -0,0 +1,50 @@
...
+type DebugFrameRequest = FrameRequest & { onchainkitDebug: Required<FrameValidationData> };
+
+/**
+ * Modify a standard frame request to include simulated values (e.g., indicate the viewer
+ * follows the cast author) for development/debugging purposes.
+ * @param request A standard frame request.
+ * @param options An object containing values we will pretend are real for the purposes of debugging.
+ * @returns
+ */
+function getDebugFrameRequest(
```
Zizzamia, line 23: Here me out, what if this is actually called `getMockFrameRequest`, I think can be very powerful also to write tests.

I see the word Mock feel more like utilities tests vibes.

src/core/getFrameMessage.ts
```diff
@@ -20,9 +22,19 @@ type FrameMessageOptions =
  * @param body The JSON received by server on frame callback
  */
 async function getFrameMessage(
-  body: FrameRequest,
+  body: FrameRequest | DebugFrameRequest,
   messageOptions?: FrameMessageOptions,
 ): Promise<FrameValidationResponse> {
+  // Skip validation only when allowed and when receiving a debug request
+  if (messageOptions?.allowDebug) {
```
Zizzamia, line 29: if this is a boolean, we should either use `has` or `is` as  prefix.

so maybe `isMockRequest: boolean`

framegear/yarn.lock
```diff
@@ -3179,6 +3191,7 @@ __metadata:
   version: 0.0.0-use.local
   resolution: "framegear@workspace:."
   dependencies:
+    "@coinbase/onchainkit": "portal:./.."
```
Zizzamia, line 3194: Inteeeeerestinggggggggg

What the heck which magic is this?

cnasc: https://yarnpkg.com/protocol/portal

cnasc: Once we finalize probably want to specify the exact version, but this is great for dev

cnasc: Now you're thinking with portals

src/core/getFrameMessage.ts
cnasc: In this case, not every request is necessarily a mock request. I think it's most clear as is: if set, it will allow mock requests to pass through, otherwise it will reject them.

PR #154, opened 2024-02-20
Add Syndicate Frame API functions & Syndicate Frame Chain config

sammdec:
**What changed? Why?**
This PR adds
- the ability to interact with the Syndicate Frame API from a set of async functions.
- Adds a Syndicate Frame Chain Viem config for those wishing to interact with the chain using their own tools
- Updates the docs to contain how to use the new functions

**Notes to reviewers**
Is there an agreed upon way of throwing errrors for the users when developing with onchainkit, currently im using `console.error` but happy to throw errors if thats better and more consistent.

**How has it been tested?**
There are unit tests for each of the functions added and I have used the SDK to write some scripts to interact with the Frame API using the SDK.


sammdec:
Closing out the PR as it's got very stale

PR #155, opened 2024-02-20
feat: add <Balance />

eragon512:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**
Unit test

### DIFF
--- a/src/components/Balance.tsx
+++ b/src/components/Balance.tsx
@@ -1,0 +1,53 @@
+import React from 'react';
+import { formatEther, parseEther } from 'viem';
+import type { Address } from 'viem';
+import { publicClient } from '../network/client';
+import { useOnchainActionWithCache } from '../hooks/useOnchainActionWithCache';
+
+type BalanceProps = {
+  address: Address;
+  className?: string;
+  decimalDigits?: number;
+  props?: React.HTMLAttributes<HTMLSpanElement>;
+};
+
+const balanceAction = (address: Address) => async (): Promise<string> => {
+  try {
+    return await publicClient
+      .getBalance({
+        address,
+      })
+      .toString();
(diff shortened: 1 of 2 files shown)


src/components/Balance.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * Balance is a React component that renders the account balance for the given Ethereum address.
+ * It displays the user's balance, with the no. of decimals specified by the 'decimalDigits' prop.
+ * By default, 'decimalDigits' is set to 3.
+ *
+ * @param {Address} props.address - The Ethereum address for which to display the balance.
+ * @param {string} [className] - Optional CSS class for custom styling.
+ * @param {number} [decimalDigits=3] - Determines the no. of decimal digits to be displayed.
+ * @param {React.HTMLAttributes<HTMLSpanElement>} [props] - Additional HTML attributes for the span element.
+ */
+export function Balance({ address, className, decimalDigits = 3, ...props }: BalanceProps) {
```
Sneh1999, line 36: Can we add this functionality to the avatar is possible?

eragon512: not sure if that's the best approach here - generally, it's best to keep each component small and atomic, with a single, clear responsibility

adding balance data to the avatar might add multiple responsibilities inside a single component

if users need balance data in their avatar, having a separate component allows them to stitch the avatar and balance components together in the way they feel fits their use case best

LMK your thoughts on this

Zizzamia: This week is hard for us to ship a new component.

Let's table this till the end of ETH Denver.

But, I love we at least started the convo.

Zizzamia:
Closing this for now, as `balance` is not activelly develop yet.

PR #163, opened 2024-02-21
fix: modules imports

Zizzamia:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,36 @@
         "default": "./lib/index.js"
       },
       "default": "./lib/index.js"
+    },
+    "./core": {
+      "types": "./lib/core/index.d.ts",
+      "browser": {
+        "types": "./esm/core/index.d.ts",
+        "module": "./esm/core/index.js",
+        "import": "./esm/core/index.js",
+        "default": "./lib/core/index.js"
+      },
+      "default": "./lib/core/index.js"
+    },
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -84,6 +84,36 @@
...
+        "import": "./esm/farcaster/index.js",
+        "default": "./lib/farcaster/index.js"
+      },
+      "default": "./lib/farcaster/index.js"
+    },
+    "./xmtp": {
+      "types": "./lib/xmtp/index.d.ts",
+      "browser": {
+        "types": "./esm/xmtp/index.d.ts",
+        "module": "./esm/xmtp/index.js",
```
neekolas, line 112: I think we're going to want the `module` outside of the browser block too, since this code is going to be loaded from ESM apps on the server.

neekolas (approved):

src/xmtp/isXmtpFrameRequest.ts
```diff
@@ -1,4 +1,4 @@
-import { XmtpOpenFramesRequest } from '@xmtp/frames-validator';
+import type { XmtpOpenFramesRequest } from '@xmtp/frames-validator';
```
neekolas, line 1: Thanks for fixing this

PR #185, opened 2024-02-24
Resolved Dependency Issue + QoL Improvements to frame gear

huntertdiamond:
**What changed? Why?**
## Dependencies
I added `graphql-request` and `viem` as dependencies to frame gear. I was receiving runtime errors when making requests from the frame and the installation of the above dependencies resolved that issue
## Styling

I changed  the styling in the components/ValidationResults/ValidationResults file to be more efficient with spacing and layout. 

### before the changes 
(screenshots were taken on a 1920 x 1080 monitor, laptop shows similar improvements). 
frameGearPreChangesMonitor
### After the changes 
changes1440x900

**Notes to reviewers**
none
**How has it been tested?**
No changes in core functionality made, testing was limited to changing the dimensions of my screen
I tested:
- frames with 1:1 aspect ratio
- frames with 4 buttons
-  frames with text input
- 1440x900 screen size
- 1920x1080 screen size

## tldr
The addition of new dependencies (`graphql-request` and `viem`) to resolve runtime errors and a revamp of the ValidationResults component's styling

### DIFF
--- a/framegear/components/Frame/Frame.tsx
+++ b/framegear/components/Frame/Frame.tsx
@@ -20,9 +20,10 @@ export function Frame() {
 }
 
 function ValidFrame({ tags }: { tags: Record<string, string> }) {
-  const { image, imageAspectRatio, input, buttons } = useMemo(() => {
+  const { image, imageAspectRatioClassname, input, buttons } = useMemo(() => {
     const image = tags['fc:frame:image'];
-    const imageAspectRatio = tags['fc:frame:image:aspect_ratio'] === '1:1' ? '1/1' : '1.91/1';
+    const imageAspectRatioClassname =
+      tags['fc:frame:image:aspect_ratio'] === '1:1' ? 'aspect-square' : 'aspect-[1.91/1]';
     const input = tags['fc:frame:input:text'];
     // TODO: when debugger is live we will also need to extract actions, etc.
     const buttons = [1, 2, 3, 4].map((index) => {
@@ -39,7 +40,7 @@ function ValidFrame({ tags }: { tags: Record<string, string> }) {
     });
     return {
       image,
-      imageAspectRatio,
+      imageAspectRatioClassname,
       input,
       buttons,
     };
@@ -49,7 +50,7 @@ function ValidFrame({ tags }: { tags: Record<string, string> }) {
     <div>
       {/* eslint-disable-next-line @next/next/no-img-element */}
       <img
-        className={`w-full rounded-t-xl aspect-[${imageAspectRatio}] object-cover`}
+        className={`w-full rounded-t-xl ${imageAspectRatioClassname} object-cover`}
         src={image}
         alt=""
       />
--- a/framegear/package.json
+++ b/framegear/package.json
@@ -10,7 +10,7 @@
     "test": "jest"
   },
   "dependencies": {
-    "@coinbase/onchainkit": "^0.9.4",
+    "@coinbase/onchainkit": "0.9.4",
     "@radix-ui/react-icons": "^1.3.0",
     "jotai": "^2.6.4",
     "next": "14.1.0",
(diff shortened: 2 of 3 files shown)


cnasc:
Hey @huntertdiamond thanks for the PR. 

> I was receiving runtime errors when making requests from the frame and the installation of the above dependencies resolved that issue

Can you paste the errors you saw here? I went and deleted `node_modules` from both `onchainkit` and `onchainkit/framegear`, then ran `yarn install` in `onchainkit/framegear`, ran the app and requested a frame and didn't see any errors come up. I know both of those are peer deps for `onchainkit` and I'm wondering if we might have something leaking -- `framegear` shouldn't require those to be present to run.

> I changed the styling in the components/ValidationResults/ValidationResults file to be more efficient with spacing and layout.

What platform/browser are you on? Noticing in your before screenshot it looks like existing styles weren't loading correctly.

Checking out your fork, it looks like we have different results on the "after" so there's probably a deeper issue we should fix

image

huntertdiamond:
hey @cnasc , 

I'm no longer getting the dependency error (am now using 0.9.4) so I can't paste it, but iirc it was similar to #187. 


For the styling:
I'm using arc, so the lack of a top bar might be contributing to the discrepancies between what I'm seeing from my fork and what you are seeing. My first thought is that the `frame` is taking the actual size of the fc:frame_image instead of a fixed size for 1:1 and 1:91. 

My latest commit sets a fixed size of 462px x 462px for 1:1 images and 462px x 241px for 1.91:1 images.


I've also added some simple functionality in `components/FrameInput/FrameInput.tsx` that stores the current frame URL as a query parameter in the URL

cnasc:
> I'm no longer getting the dependency error (am now using 0.9.4) so I can't paste it, but iirc it was similar to https://github.com/coinbase/onchainkit/pull/187.

I'm wondering if this is because we had carets in `package.json` so people who cloned later got incompatible onchainkit versions. It _shouldn't_ work that way pre 1.0, but 🤷‍♂️ looks like we've got it mostly sorted.

> My first thought is that the frame is taking the actual size of the fc:frame_image instead of a fixed size for 1:1 and 1:91.

This should be covered by the `aspect-` classes, and if the image doesn't fit the aspect ratio it should clip the edges until it does. Will double check to see if that's what's going on. Overall would prefer to have these fluid rather than fixed.

Also agree overall with making the spacing more efficient, but I think we can come up with a neater looking solution. Will try to cook something up today!

cnasc:
image

@huntertdiamond Looking at consolidating some items (e.g., all related tags in same row) and I'm not sure if the juice is worth the squeeze on this. We do make the results shorter vertically, but it comes at the cost of making it less easy to scan down the list (goes from 1-dimensional to 2-dimensional). I wonder if we'd be better off replacing this component with a filterable table (e.g., only show errors, only show buttons, etc.). That will allow users to focus just on the information relevant to them at the time. Any thoughts on that approach?

huntertdiamond:
here is what the accordion looks like. The default for each row is open. I personally think it solves the issue, but would be curious to hear your thoughts
Screenshot 2024-02-24 at 9 14 35 PM

Zizzamia:
Ciao @huntertdiamond, I believe @cnasc merged some stuff.
Would you mind rebase your branch so we can get a good review with your ideas mix with the lastest `main`.

Btw for what is worth, I believe we should not have the main with be more than **1280 px**, I might be old fashion, but there my semi-designer eye feel weird and a bit lost as soon things get too wide.

The golden dimension for web usually are:
- sm (640px)
- md (768px)
- lg (1024px)
- xl (1280px)

I would prefer we optimize for those.

huntertdiamond:
hey @Zizzamia , I accidentally closed this PR when rebasing, going to submit a new one momentarily with my changes and some context.

PR #188, opened 2024-02-24
add content type to framegear post frame request

DonGambas:
**What changed? Why?**
- added `'Content-Type': 'application/json'` to post frame request headers.
- sending this request without the content-type  can cause server parsing issues. happened for me with express `4.17.21`

**Notes to reviewers**

**How has it been tested?**
- tested locally using express server. don't think this should cause issues with other servers, but you might want to do further testing.


cnasc (approved):
LGTM! good find, did this on the client side and forgot to pass through on api route

cnasc:
cc @Zizzamia

PR #190, opened 2024-02-24
feat: framegear add styles for light mode, make legible 

DonGambas:
**What changed? Why?**

I know, I know we should all be using darkmode all the time, but my machine is on auto, meaning light mode during the day. Right now lightmode in framegear makes most of the content illegible, and most importantly you can't read the URL that you're entering into the fetch frame input. I made simple changes that seem to be inline with the rest of the styling to support light mode, don't want to disrespect any plans you might have for implementing a design system etc, but this makes light mode usable for now. 

_This is what light mode looks like currently_ 

lightmode_current

_This is what light mode looks like with changes_

lightmode_new

**Notes to reviewers**

**How has it been tested?**

Tested in Chrome Version 1.62.165 Chromium on MacOs 14.0 with light mode enabled, dark mode enabled and auto enabled.

### DIFF
--- a/framegear/components/Header/Header.tsx
+++ b/framegear/components/Header/Header.tsx
@@ -16,7 +16,7 @@ export function Header() {
 function Banner() {
   return (
     <div
-    className={`max-w-layout-max border-pallette-line bg-content-light dark:bg-banner flex w-full items-center justify-between rounded-lg border p-6`}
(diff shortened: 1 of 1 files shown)


cnasc:
> I know, I know we should all be using darkmode all the time

You know I actually prefer light mode but I just went with what the designer gave me 😆 

LGTM overall! The button link inside the top banner doesn't show up well, but we can add more comprehensive support in a future PR.

Issue #146, opened 2024-02-19
Meta Issue: framegear


cnasc:
### Describe the solution you'd like

At some point it would be nice to group these into a project, but for the time being listing some outstanding ideas:

- [ ] show frame payload HTML tags (and copy button)
- [ ] show frame payload as an `onchainkit` metadata object (and copy button)
- [ ] verification
  - [x] show required tags even when not present in payload (#176)
  - [ ] show all possible tags (possibly collapsed) even if not present in payload
- [x] distinct button icons for distinct button actions (e.g., `post_redirect`, `mint`, and `link` buttons) (#203)
- [ ] rendered frame should support all button actions
  - [x] `post`
  - [x] `post_redirect`
  - [x] `link`
  - [ ] `mint`
- [x] rendered frame should include text input in mock request (https://github.com/coinbase/onchainkit/pull/201)
- [ ] allow user to select certain simulated conditions (e.g., liked/recasted, followed, etc.)
- [x] favicon (awaiting design) (https://github.com/coinbase/onchainkit/pull/204)
- [ ] Identify when a frame can accept debug actions, inform user what is possible
- [ ] more comprehensive light/dark mode styling (see #190)
- [x] support `fc:frame:state` tag (see https://github.com/coinbase/onchainkit/pull/193#issuecomment-1963041332) (#232)
- [ ] frame input bug (state seems to persist after next frame renders)
- [ ] more sophisticated state handling (page through previous, time travel debugger)

taycaldwell:
PR: https://github.com/coinbase/onchainkit/pull/200

Rendered frame should support all button actions
Distinct button icons for distinct button actions
  -  `post_redirect`
  - `link`

cnasc:
@huntertdiamond I see you added a commit to add simulated conditions to the frame. Do you want to drive on that one? And/or cool if I build on it sometime in the next few days?

anader123:
gm there appears to be a bug with framegear where post action buttons don't hit the target endpoint and posts default go to the posturl

cpcramer:
Frames has been deprecated.

PR #193, opened 2024-02-25
re: #185: Styling changes

huntertdiamond:
**What changed? Why?**
- switched from grid layout to a simple flex container with decimal sizing. There was only 1 "row" below the `header` in the `home` component, so I didn't see a reason to use grid layout over a more intuitive flex container
- added accordion option to `ValidationResults`. The previous iteration was inefficient with screen real estate and this gives users the option to show / hide tags. It defaults to open for all FC related tags
- changed the header to have the `appname` container and the `banner` in one row instead of two. Having the header below wasted a lot of vertical space.
- changed the fc frame spec link to an `<a>` element instead of the next/link. 

### Changes Visualized
Screenshot 2024-02-24 at 10 15 32 PM (2)

**Notes to reviewers**
check #185 for more detail

>Re: max width of 1280

The previous PR (#185) had a maxwidth of 1280. The header did not, due to a pretty common tailwind quirk where centering the item reduces the size. This resolves that as well.

**How has it been tested?**

### DIFF
--- a/framegear/components/Frame/Frame.tsx
+++ b/framegear/components/Frame/Frame.tsx
@@ -2,7 +2,7 @@ import { postFrame } from '@/utils/postFrame';
 import { frameResultsAtom } from '@/utils/store';
 import { useAtom } from 'jotai';
 import { ChangeEvent, PropsWithChildren, useCallback, useMemo, useState } from 'react';
-import { ArrowTopRightIcon } from '@radix-ui/react-icons';
+import { ExternalLinkIcon, ResetIcon, RocketIcon } from '@radix-ui/react-icons';
 import { useRedirectModal } from '@/components/RedirectModalContext/RedirectModalContext';
 
 export function Frame() {
@@ -153,21 +153,31 @@ function FrameButton({
     }
     // TODO: implement other actions (mint, etc.)
   }, [button?.action, button?.index, button?.target, inputText, openModal, setResults]);
+
+  const buttonIcon = useMemo(() => {
+    switch (button?.action) {
+      case 'link':
+        return <ExternalLinkIcon />;
+      case 'post_redirect':
+        return <ResetIcon />;
+      case 'mint':
+        return <RocketIcon />;
+      default:
+        null;
+    }
+  }, [button?.action]);
+
   return (
     <button
-      className="border-button w-[45%] grow rounded-lg border bg-white px-4 py-2 text-black"
+      className="border-button flex w-[45%] grow items-center justify-center gap-1 rounded-lg border bg-white px-4 py-2 text-black"
       type="button"
       onClick={handleClick}
       disabled={isLoading || button?.action === 'mint'}
     >
-      <span className="block max-w-full overflow-hidden text-ellipsis whitespace-nowrap">
+      <span className="block max-w-[90%] overflow-hidden text-ellipsis whitespace-nowrap">
         {children}
-        {button?.action === 'post_redirect' || button?.action === 'link' ? (
-          <ArrowTopRightIcon className="ml-1 inline" />
-        ) : (
-          ''
-        )}
       </span>
+      {buttonIcon}
     </button>
   );
 }


Zizzamia:
I am going to have @cnasc share his thoughts as well.

But let's go one by one

> switched from grid layout to a simple flex container with decimal sizing. There was only 1 "row" below the header in the home component, so I didn't see a reason to use grid layout over a more intuitive flex container

What value is there in swithing from `grid` to `flex`? Is there a real value for the project? or is more personal preference? It seems to me is a personal preference. 

For what is worth, the theory says, that when is about organizing grid, using `display grid` is raccomended, but when organizing element inside a block, then `display flex` is raccomended.

We can disagree on this, but that's just how the CSS API was designed.

```
The basic difference between CSS grid layout and [CSS flexbox layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout) is that flexbox was designed for layout in one dimension - either a row or a column. Grid was designed for two-dimensional layout - rows, and columns at the same time.
```
Quote from https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_grid_layout/Relationship_of_grid_layout_with_other_layout_methods

**So my advice here is to keep the Grid choice**

Zizzamia:
> added accordion option to ValidationResults. The previous iteration was inefficient with screen real estate and this gives users the option to show / hide tags. It defaults to open for all FC related tags

What ineficiency means here? We can improve efficiency in other ways, like https://warpcast.com/~/developers/frames does where similar tags are grouped together, and a simple scroll is enough.

Show and hide is usually never a good UX becuase eventually it distracts folks.

What we should do instead is learn from https://warpcast.com/~/developers/frames and group values together.

Zizzamia:
> changed the header to have the appname container and the banner in one row instead of two. Having the header below wasted a lot of vertical space.

Honestly that's just ugly all in one row. I agree we can optimize space, but at what coast? if it's ugly than it is not enjoable to use.

I agree we can improve, as that was just a V0, but all in one line is not an answear.

Zizzamia:
> changed the fc frame spec link to an <a> element instead of the next/link.

Why? what's the problem we are solving by doing that?

Zizzamia:
@huntertdiamond, we appreciate your PR, and I think there are several point of reflection with the team we have to take in improving the layout of Framegear.

More than the code solution, I am curios to understand the thinking behind the code. And what are you looking into each decision you tried to make.

cnasc:
IMO a great way to handle the verification results would be something like the network tab in browser dev tools. This is an interface that most devs should be comfortable with and it allows sorting and filtering (e.g., "show me only tags starting with `fc:frame:button`" or "only show me tags with errors") which allows the user to scope the information to the task at hand.

image

Zooming out, here are example use cases I see for the validation results (from coarse to fine):

1. Is this frame valid? (all I care about is the top line)
2. If not, where are my errors? (all I care about are the tags with errors)
3. I'm iterating on one specific response, are the values I see the ones I expect? (all I care about is one specific tag or set of tags)

I think the network-tab-style display handles those cases better than either the existing design or the proposed accordion design.

cnasc:
Regarding the app name and banner, agree that we can make more compact but I think it requires a different design. The banner in particular is probably a bit overdone, we can assume that the user who specifically launched this tool understands that it is a frame debugger. The most useful piece in there is the link to the official spec. We can probably make a more traditional header where we have the app name on one end and then links across the rest of the row.

huntertdiamond:
gm @Zizzamia 

> What value is there in swithing from `grid` to `flex`? Is there a real value for the project? or is more personal preference? It seems to me is a personal preference.

The value is that it adheres to:

> The basic difference between CSS grid layout and [CSS flexbox layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout) is that flexbox was designed for layout in one dimension - either a row or a column. Grid was designed for two-dimensional layout - rows, and columns at the same time.


Quote from https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_grid_layout/Relationship_of_grid_layout_with_other_layout_methods



Looking at the three instances I changes the grid to a flexbox, all three are resolving single dimension layout issues. I made the changes because I agree that there isn't a point to use a grid when there is only one dimension within the container in which the styling is being applied :)


>What ineficiency means here? We can improve efficiency in other ways, like https://warpcast.com/~/developers/frames does where similar tags are grouped together, and a simple scroll is enough.

Inefficiency meaning that there is a ton of wasted space on tags that aren't important to the user who is using the debugger.

I don't necessarily agree that when making a `pro` tool, giving them the option to hide redundant data is distracting, but i could very well be wrong. 


>Honestly that's just ugly all in one row. I agree we can optimize space, but at what cost? if it's ugly than it is not enjoable to use.

I agree we can improve, as that was just a V0, but all in one line is not an answear.

I agree-its ugly. I don't agree that it make the debugger unenjoyable to use, but I appreciate your commitment to aesthetic. My understanding that as a v0 I was going to submit changes that I felt were needed after using it for a few dozen hours. I didn't want to overstep and delete entire portions, but agree it makes the header extremely ugly- albeit efficient and makes room for the information I'm using the app for.



> changed the fc frame spec link to an element instead of the next/link.
> 
> Why? what's the problem we are solving by doing that?




the `<Link>` element doesn't open up a new tab when pressed. It was annoying me.

huntertdiamond:
@cnasc yes I agree that is the perfect implementation. check 
[https://ping-frame.vercel.app/api/dev]

cnasc:
Just to pare this down a bit so we don't discuss too many things all at one, I opened #194 which should take care of the header concerns for now. 

IMO we should table (no pun intended) the validation results for now, and we can start working toward a nicer more full featured display in separate tasks.

I think that leaves just the layout concerns, unless I've missed something.

Zizzamia:
@huntertdiamond thank you for the back and forward. I see @cnasc already taking a step forward into using some your feedback in practice.

My 2cents also is that I act very pragmatic, so decisions are made with having all tradeoffs took into considerations. Sharing your tradeoff thinking helps me and the team understand your decisions and much faster act on those.

Decisions like Flex vs Grid can enter very quickly into the religious world of CSS, and it's my job to keep everyone in check and stay the most pragmatic as possible. And for what is worth, I love equally both, I don't have preferences, I just like a good pragmatic conversation with all tradeoffs took into considerations.

And thank you again, this is a great convo 🌊

huntertdiamond:
@Zizzamia  Agreed! Very important to manage potential distractions in an early product such as this

I've been submitting my PRs because I'm using the debugger **a lot**, and felt I might as well share the changes that I've been making to improve the experience, and I'll continue doing so (without spamming / adding personal preference into the mix) :)


In regard to important changes that don't delve into the murky land of CSS purism:

I made a very quick component that displays the body of the new `fc:frame:state` tag as formatted JSON from the passed URI encoded string. 

Screenshot 2024-02-25 at 2 29 45 PM
It isn't ready for prod, but thought I'd put it on the radar, and maybe there should be something simliar added to #146 

Enjoy the rest of your weekend!

cnasc:
@huntertdiamond appreciate the feedback and the code! When I get a chance I'll post some design comps we have indicating the original idea, and we can figure out how to work in the state tag.

PR #201, opened 2024-02-29
feat: pass input text through post action

cnasc:
**What changed? Why?**
Part of #146 
Passing input text value in the frame action.

Also cleaned up some lint/type errors while I was in there.

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/framegear/components/Frame/Frame.tsx
+++ b/framegear/components/Frame/Frame.tsx
@@ -1,7 +1,7 @@
 import { postFrame } from '@/utils/postFrame';
 import { frameResultsAtom } from '@/utils/store';
 import { useAtom } from 'jotai';
-import { PropsWithChildren, useCallback, useMemo, useState } from 'react';
+import { ChangeEvent, PropsWithChildren, useCallback, useMemo, useState } from 'react';
 import { ArrowTopRightIcon } from '@radix-ui/react-icons';
 import { useRedirectModal } from '@/components/RedirectModalContext/RedirectModalContext';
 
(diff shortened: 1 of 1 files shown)


framegear/components/Frame/Frame.tsx
```diff
@@ -119,13 +128,16 @@ function FrameButton({
...
             hash: '0xthisisnotreal',
           },
-          inputText: '',
+          inputText,
           fid: 0,
           messageHash: '0xthisisnotreal',
           network: 0,
           timestamp: 0,
         });
-        setResults((prev) => [...prev, result]);
```
cnasc, line 128: we got an error on this because `postFrame` may return an undefined value, and only defined values should go in the results array.

framegear/components/Frame/Frame.tsx
```diff
@@ -119,13 +128,16 @@ function FrameButton({
...
+        // TODO: handle when result is not defined
```
cnasc, line 137: Probably should be an error toast or something

framegear/components/Frame/Frame.tsx
```diff
@@ -140,7 +152,7 @@ function FrameButton({
       openModal(onConfirm);
     }
     // TODO: implement other actions (mint, etc.)
-  }, [button, setResults]);
+  }, [button?.action, button?.index, button?.target, inputText, openModal, setResults]);
```
cnasc, line 155: Had a warning here because `openModal` wasn't included as dep. The other formatting was done automatically by IDE.

Zizzamia (approved):

PR #205, opened 2024-02-29
feat: transform raw data into FrameMetadata (framegear)

cnasc:
**What changed? Why?**
So far we've been working with the frame data just as a `Record<string, string>` which causes some unnecessary verbosity and offers the opportunity for typos to lead to runtime errors.

I think it would make sense for framegear to take advantage of the `FrameMetadata` type that `@coinbase/onchainkit` provides. After validation, we can pass the results into this converter function and then components can operate on known structured data. This lets us compartmentalize the error prone indexing into the base record into one function.

While I was in there I:
- updated `@coinbase/onchainkit` to latest to get the latest type definitions
- updated the validation and associated test
  - to support `state`
  - to fix a misleading message
  - to fix a misleadingly named test

**Notes to reviewers**
Putting this up as draft for now, don't want to change the code too radically to use this just yet. Open to thoughts on this direction

**How has it been tested?**


cnasc:
@Zizzamia this is ready to go in, will start using it once this is merged so we have smaller diffs to review

Zizzamia (approved):

PR #208, opened 2024-03-04
feat: add `tx` as a Frame action option

Zizzamia:
**What changed? Why?**
 Added `tx` as a Frame action option, enabling support for Frame Transactions. 

Spec: https://warpcast.notion.site/Frame-Transactions-Public-Draft-v2-9d9f9f4f527249519a41bd8d16165f73

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/frame/types.ts
+++ b/src/frame/types.ts
@@ -77,7 +77,7 @@ export function convertToFrame(json: any) {
  */
 export type FrameButtonMetadata =
   | {
-      action: 'link' | 'mint';
+      action: 'link' | 'mint' | 'tx';
       label: string;
       target: string;
     }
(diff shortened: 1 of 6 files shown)


src/frame/types.ts
```diff
@@ -77,7 +77,7 @@ export function convertToFrame(json: any) {
  */
 export type FrameButtonMetadata =
   | {
-      action: 'link' | 'mint';
+      action: 'link' | 'mint' | 'tx';
```
tudormunteanu, line 80: tiny nitpick: can we have `txn` instead of `tx` (Texas?)

Zizzamia: That's a great point!

OnchainKit will follow the Frame spec https://warpcast.notion.site/Frame-Transactions-Public-Draft-v2-9d9f9f4f527249519a41bd8d16165f73 

if they change it to `txn`, we will do too.

Feel free to ask there if they will consider `txn` vs `tx.

PR #212, opened 2024-03-05
fix: update neynar frame validation type

Flickque:
**What changed? Why?**

updated NeynarFrameValidationInternalModel type, for type safe handling of state and input in a validated message


Zizzamia:
Niiice, thank you for this!

PR #229, opened 2024-03-06
word-break break-all for validator entry values

dannyDNS:
**What changed? Why?**

Added `break-all` to frame `ValidationEntry` value container to allow longer values to freely wrap without breaking the layout.

**Notes to reviewers**

Before
before

After
after

**How has it been tested?**

Visually.


Zizzamia:
@dannyDNS would you mind run `yarn check` at the root of onchainkit, so it fix the format, please.

Zizzamia (approved):

PR #232, opened 2024-03-07
feat: Frame now uses FrameMetadata instead of string record

cnasc:
**What changed? Why?**
Following on #205, using the generated results within the `Frame` component. For the time being `ValidationResults` continues to use the legacy `Record<string, string>` tag object because it needs to be able to point to the actual tag name rather than the names from `FrameMetadata`.

Additionally started handling the frame state when posting back to api (part of #146).

**Notes to reviewers**

**How has it been tested?**
End-to-end against `a-frame-in-100-lines` to ensure results were same as before.

### DIFF
--- a/framegear/components/Frame/Frame.tsx
+++ b/framegear/components/Frame/Frame.tsx
@@ -70,6 +70,7 @@ function ErrorFrame() {
   // TODO: implement -- decide how to handle
   // - simply show an error?
   // - best effort rendering of what they do have?
+  // - maybe just ValidFrame with a red border?
   return <PlaceholderFrame />;
 }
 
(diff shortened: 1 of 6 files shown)


framegear/components/Frame/Frame.tsx
```diff
@@ -14,40 +15,14 @@ export function Frame() {
 
   const latestFrame = results[results.length - 1];
 
-  if (!latestFrame.isValid) {
```
cnasc, line 17: Removed this because right now it's more useful to fully render the frame even if it is not 100% correct

Zizzamia (approved):

PR #243, opened 2024-03-12
fix: `button.target` is not dependent on `button.action`

### LINKED ISSUE #242: Bug: `button.target` is only recognized if `button.action` is explicitly set

1. Create frame metadata with a button that has a `target` set, but does not explicitly set `action`
2. Note that the button's target is not set

Per [the spec](https://docs.farcaster.xyz/reference/frames/spec#optional-properties) ([also](https://docs.farcaster.xyz/reference/frames/spec#handling-clicks)): `target` should not require `action` to be set explicitly, as `action` defaults to `"post"`.

_No response_

cnasc:
**What changed? Why?**
Resolves #242 

Pass `button.target` through even if `button.action` is not set.

**Notes to reviewers**

**How has it been tested?**
Unit tests

### DIFF
--- a/.changeset/angry-fishes-greet.md
+++ b/.changeset/angry-fishes-greet.md
@@ -1,5 +1,5 @@
 ---
-"@coinbase/onchainkit": patch
+'@coinbase/onchainkit': patch
 ---
 
 fix: `button.target` is not dependent on `button.action` (#243)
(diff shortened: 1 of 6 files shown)


Zizzamia:
Nice, also for this to be completed as PR:
1. run `yarn changeset` with the proper description
2. is any docs that needs to be updated or improved from this change?
3. this is a bump `patch` version, make sure to update the `version` file as well.

cnasc:
Added changeset and version update. No docs impacted by this change (if anything, https://onchainkit.xyz/frame/types#framebuttonmetadata is now more correct than it was)

Zizzamia (approved):

PR #251, opened 2024-03-13
chore: update contributing

kyhyco:
**What changed? Why?**
Update contributing to point to onchainkit/issues

**Notes to reviewers**

**How has it been tested?**


Zizzamia (approved):
Nice catch!

PR #259, opened 2024-03-19
feat: attest avatar and name in identity kit

kyhyco:
**What changed? Why?**
- add attestation to `Avatar` and `Name` components using EAS attestation

| Name | Avatar |
| --- | --- |
| Screenshot 2024-03-18 at 9 01 28 PM | Screenshot 2024-03-18 at 9 01 48 PM |

**Notes to reviewers**
This PR is not in an ideal state but let's use this PR to discuss bunch of stuff...

**How has it been tested?**
locally


package.json
```diff
@@ -123,9 +125,9 @@
         "types": "./esm/identity/index.d.ts",
         "module": "./esm/identity/index.js",
         "import": "./esm/identity/index.js",
-        "default": "./lib/identity/index.js"
+        "default": "./esm/identity/index.js"
```
kyhyco, line 128: I was able to make onchainkit work with vocs with this change.

site/docs/pages/identity/avatar.mdx
```diff
@@ -3,6 +3,17 @@
...
+
+import { Avatar } from '@coinbase/onchainkit/identity';
+import { App } from './App';
+
+<App>
+  <Avatar address="0x1234567890abcdef1234567890abcdef12345678" />
+  <Avatar address="0x02feeb0AdE57b6adEEdE5A4EEea6Cf8c21BeB6B1" />
+  <Avatar address="0x02feeb0AdE57b6adEEdE5A4EEea6Cf8c21BeB6B1" displayBadge />
+</App>
+
```
kyhyco, line 16: Testing that this works. Will make this part better with explanation.

site/package.json
```diff
@@ -8,8 +8,11 @@
     "preview": "vocs preview"
   },
   "dependencies": {
-    "@coinbase/onchainkit": "0.9.10",
+    "@coinbase/onchainkit": "portal:./..",
```
kyhyco, line 11: Will take a look at how we generate the site and make sure this will work.

package.json
```diff
@@ -14,7 +14,8 @@
     "test:coverage": "jest . --coverage ",
     "release:check": "changeset status --verbose --since=origin/main",
     "release:publish": "yarn install && yarn build && changeset publish",
-    "release:version": "changeset version && yarn install --immutable"
+    "release:version": "changeset version && yarn install --immutable",
+    "watch": "packemon watch"
```
kyhyco, line 18: We can run `yarn watch` along with `cd site && yarn dev` to get hot reload with doc site

src/identity/components/useVerified.ts
```diff
@@ -0,0 +1,22 @@
...
+import type { Address } from 'viem';
+
+import { getEASAttestations } from '../getEASAttestations';
+
+export function useVerified(address: Address) {
+  const [verified, setVerified] = useState(false);
+
+  useEffect(() => {
+    const fetchData = async () => {
+      const attestation = await getEASAttestations(address, base);
```
kyhyco, line 12: We can pass in the `options` to `getEASAttestations` here. Do we want to introduce an `IdentityProvider` to get the `options` to `getEASAttestations`?

package.json
kyhyco: cc @wespickett

.gitignore
```diff
@@ -52,6 +52,7 @@ package-lock.json
 !.yarn/plugins
 !.yarn/sdks
 !.yarn/versions
+install-state.gz
```
Zizzamia, line 55: I believe we want this check-in. Right @wespickett ?

package.json
Zizzamia: I understand the idea, but not a fan of just call it `watch` and being at the root. The question here should be, do we want to test the `src` or the `dist`?

package.json
Zizzamia: I think that's a massive breaking changes. And we should avoid on doing that.

.gitignore
wespickett: @Zizzamia No I don't think so, I think it should be ignored. But also shouldn't this be `.yarn/install-state.gz`? And there is already a rule (`.yarn/*`) that should ignore `.yarn/install-state.gz`

Zizzamia: @kyhyco let's do a solo PR for this change, so we can that merged.

package.json
Zizzamia: Let's consider this breaking change only at v1 cut discussion, which will be in late May.

src/identity/index.ts
```diff
@@ -5,6 +5,7 @@ export { getAvatar } from './core/getAvatar';
 export { getName } from './core/getName';
 export { useAvatar } from './hooks/useAvatar';
 export { useName } from './hooks/useName';
+export { IdentityProvider } from './hooks/useIdentity';
```
Zizzamia, line 8: Ok love this, althought part of me thinks it should be more explicit, maybe `OnchainKitIdentityProvider`, but I think folks can probably customize that..... let's see for now is fine.

src/identity/hooks/useVerified.ts
```diff
@@ -0,0 +1,26 @@
+import { useEffect, useState } from 'react';
+import { base } from 'viem/chains';
+import type { Address } from 'viem';
+
+import { getEASAttestations } from '../getEASAttestations';
+import { useIdentity } from './useIdentity';
+
+export function useVerified(address: Address) {
```
Zizzamia, line 8: Maybe it will be better
```
useVerified({
  address: Address,
  chain?: ChainType
});
```

And when not definied chain is Base

src/identity/hooks/useVerified.ts
```diff
@@ -0,0 +1,26 @@
...
+
+      if (found !== undefined) {
+        setVerified(true);
+      }
+    };
+
+    fetchData();
+  }, [address]);
+
+  return verified;
```
Zizzamia, line 25: so `useVerified` return if an address is verified, and the schema comes from the `identity.schemaId`.

But in the same app, I can see we use multiple schemas. So I think we should be more explicit in the params of the hook what schema we are checking.

Zizzamia: `useVerified` as name feels a bit too vague

src/identity/hooks/useIdentity.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+import { getEASAttestations } from '../getEASAttestations';
+
+type IdentityContextType = {
+  schemaId: string;
+};
+
+const IdentityContext = createContext<IdentityContextType | null>(null);
+
+function checkSchemaId(hash: string): hash is `0x${string}` {
```
Zizzamia, line 13: Let's have this in his own file.

src/identity/hooks/useIdentity.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return /^0x[a-fA-F0-9]{64}$/.test(hash);
+}
+
+type SchemaId = `0x${string}`;
```
Zizzamia, line 17: Let's move types in the types file.

src/identity/components/internal/WithNameBadge.tsx
```diff
@@ -0,0 +1,35 @@
+import type { Address } from 'viem';
```
Zizzamia, line 1: Avoid create a sub folder `internal`, let's keep the code as flat as possible. It will make it easier to mantain.

src/identity/components/internal/WithNameBadge.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { useVerified } from '../../hooks/useVerified';
+import { Badge } from './Badge';
+
+type DisplayBadgeProps = {
+  children: React.ReactNode;
+  address: Address;
+};
+
+function DisplayBadge({ children, address }: DisplayBadgeProps) {
```
Zizzamia, line 10: We are going to have multiple badges, so if a badge is assigned to a schema, than we should probably define a schema in `DisplayBadge`

src/identity/components/Avatar.tsx
```diff
@@ -1,13 +1,19 @@
...
+import { useVerified } from '../hooks/useVerified';
+import { WithAvatarBadge } from './internal/WithAvatarBadge';
 
 type AvatarProps = {
   address: Address;
   className?: string;
   loadingComponent?: JSX.Element;
   defaultComponent?: JSX.Element;
   props?: React.ImgHTMLAttributes<HTMLImageElement>;
+  displayBadge?: boolean;
```
Zizzamia, line 16: I am not super sure `displayBadge` is the right name. I mean is not bad. @kyhyco what's your take?

src/identity/components/internal/WithAvatarBadge.tsx
```diff
@@ -0,0 +1,58 @@
+import type { Address } from 'viem';
+import { useVerified } from '../../hooks/useVerified';
+import { Badge } from './Badge';
+
+type DisplayBadgeProps = {
+  children: React.ReactNode;
+  address: Address;
+};
+
+function DisplayBadge({ children, address }: DisplayBadgeProps) {
```
Zizzamia, line 10: Let's have two files so it's easier to test and mantain

src/identity/components/Avatar.tsx
kyhyco: some other options:
- `displayAttestation`
- `displayAttested`
- `showBadge`
- `isAttested`

src/identity/components/internal/WithAvatarBadge.tsx
kyhyco: Let me rename the component here. These two should live together.

src/identity/components/Avatar.tsx
kyhyco: How about just `attest`?

src/identity/index.ts
Zizzamia: or, if all OnchainKit components will have some shared `config`, should we instead do `OnchainKitProvider`?

What do you think @kyhyco @cnasc ?

src/identity/types.ts
```diff
@@ -59,3 +59,5 @@ export type GetAvatarReturnType = string | null;
  * Note: exported as public Type
  */
 export type GetNameReturnType = string | null;
+
+export type SchemaId = `0x${string}`;
```
Zizzamia, line 63: What's the difference with `EASSchemaUid`? if they are the same let's just use `EASSchemaUid`

src/identity/hooks/useAttestAddress.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import { getEASAttestations } from '../getEASAttestations';
+import { useIdentity } from './useIdentity';
+import { attestationMapping } from '../attestationMapping';
+
+type UseAttestAddress = {
+  address: Address;
+  chain?: Chain;
+};
+
```
Zizzamia, line 13: @kyhyco what's our goal with `useAttestAddress`, can you share it out loud please. I am trying to think on a better name and better return type.

src/identity/components/Avatar.tsx
```diff
@@ -1,13 +1,18 @@
...
-import type { Address } from 'viem';
+import { WithAvatarBadge } from './WithAvatarBadge';
 
 type AvatarProps = {
   address: Address;
   className?: string;
   loadingComponent?: JSX.Element;
   defaultComponent?: JSX.Element;
   props?: React.ImgHTMLAttributes<HTMLImageElement>;
+  attest?: boolean;
```
Zizzamia, line 15: Maybe `showAttestationBadge` or `displayAttestationBadge` is better

verb + utility + component

@kyhyco what do you think?

kyhyco: changed to `attest`

src/identity/attestationMapping.tsx
```diff
@@ -0,0 +1,49 @@
...
+      chain,
+      config,
+    }: {
+      address: Address;
+      chain: Chain;
+      config: unknown;
+    }) => boolean;
+  }
+>;
+
```
Zizzamia, line 24: Helping understand, why do we need a mapping? it feels more complexity to me.

What am I missing?

src/identity/checkAddress.ts
```diff
@@ -0,0 +1,3 @@
+export function checkAddress(hash: string, length: number): hash is `0x${string}` {
```
Zizzamia, line 1: `checkAddress` as name feels very vague, what does this exactly?

src/identity/components/Avatar.tsx
Zizzamia: `attest` feels even more confusing....

src/identity/hooks/useAttestAddress.ts
kyhyco: `useAttestAddress` will return `null` if user is not attested. Return `eas` or `nouns` (whatever attests the user) if attested.

src/identity/components/Avatar.tsx
kyhyco: `showAttestation`?

src/identity/checkAddress.ts
kyhyco: Maybe `checkHashLength` is more appropriate. Giving early errors to the devs if they used an incorrect schemaId

src/identity/attestationMapping.tsx
kyhyco: We want to change two things depending on the attestation method:
1. Badge component
2. Business logic to determine if user is attested on a given method

kyhyco: We could have something like: 
```
const mapping = {
  eas: {
    Badge: EasBadge,
    predicate: () => {...}
  },
  nouns: {
    Badge: NounsBadge,
    predicate: () => {...}
  }
}
```

I am assuming non-eas attestations will have different ways to check attestations of a given address

src/identity/index.ts
kyhyco: `OnchainKitProvider` sounds good to me. Let me tinker with it.

src/identity/hooks/useAttestAddress.ts
kyhyco: changed to `useAttestation`

src/identity/attestationMapping.tsx
Zizzamia: @kyhyco just for context, EAS doesn't need this kind of mapping. As attestation are connected to a particular Wallet Address, and not to a namespace like ENS or Nouns.

So I have a good feeling this mapping is not necessary.

src/identity/components/Avatar.tsx
```diff
@@ -1,13 +1,18 @@
...
-import type { Address } from 'viem';
+import { WithAvatarBadge } from './WithAvatarBadge';
 
 type AvatarProps = {
   address: Address;
   className?: string;
   loadingComponent?: JSX.Element;
   defaultComponent?: JSX.Element;
   props?: React.ImgHTMLAttributes<HTMLImageElement>;
+  showAttestation?: boolean;
```
Zizzamia, line 15: yeah I think `showAttestation` makes sense.

src/identity/attestationMapping.tsx
kyhyco: deleted the mapping file

src/identity/checkHashLength.ts
```diff
@@ -0,0 +1,3 @@
+export function checkHashLength(hash: string, length: number): hash is `0x${string}` {
```
Zizzamia, line 1: Can you add JS docs on this please.

Also should this maybe live in `src/utils` ?

src/identity/components/WithAvatarBadge.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { useAttestation } from '../hooks/useAttestation';
+import { Badge } from './Badge';
+
+type WithAvatarBadgeInnerProps = {
+  children: React.ReactNode;
+  address: Address;
+};
+
+function WithAvatarBadgeInner({ children, address }: WithAvatarBadgeInnerProps) {
+  const attestation = useAttestation(address);
```
Zizzamia, line 11: I like `useAttestation` a lot, athought I am a bit confused on how it works. Let's revisit this in the next PR.

src/identity/components/WithAvatarBadge.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <div style={{ position: 'relative', width: '32px', height: '32px' }} data-testid="inner">
+      {children}
+      {attestation === 'eas' && (
```
Zizzamia, line 16: Why are we checking `eas`? Attestation are only `eas`. what am I missing?

Zizzamia:
That's dope, and we can merge this and do a quick follow up to get into what is a more flexible API for those components.

Like expand the config to support
```ts
type OnchainKitProviderProps = {
  identity: {
    address: Address;
    attestationList: {
      schemaId: EASSchemaUid;
      chain: Chain;
    }[];
    attestationDefaultId: EASSchemaUid;
  };
  children: ReactNode;
};
```

which will allow us to have a default experience like 
```ts
<OnchainKitProvider identity={}>
  <Avatar displayBadge />
  <Name displayBadge />
</OnchainKitProvider>
```

and more custom experience like 
```ts
<OnchainKitProvider identity={}>
  <Avatar address={address} displayBadge />
  <Name address={address} displayBadge />
</OnchainKitProvider>
```

and
```ts
<OnchainKitProvider identity={}>
  <Avatar address={address} attestationId={attestationId} displayBadge />
  <Name address={address} attestationId={attestationId} displayBadge />
</OnchainKitProvider>
```

Zizzamia:
Great job @kyhyco 👏

PR #285, opened 2024-04-12
feat: init Open Frame spec support

Zizzamia:
**What changed? Why?**
We have begun adhering to the Open Frames Spec (https://github.com/open-frames/standard) to ensure that OnchainKit is prepared for the upcoming applications that will utilize it.

**Notes to reviewers**

**How has it been tested?**


site/docs/pages/frame/types.mdx
```diff
@@ -84,9 +84,13 @@ type FrameMetadataReact = FrameMetadataType & {
 
 ```ts
 type FrameMetadataType = {
+  accepts?: {
```
daria-github, line 87: love this approach! 👏

Issue #319, opened 2024-05-02
Feature Request: Framegear URL field is too short


briandoyle81CB:
### Describe the solution you'd like

It should show the entire url.  Some frames depend on long urls or query params that aren't visible when testing.

MukulKolpe:
Hey @briandoyle81CB, @Zizzamia can I work on this issue?

Zizzamia:
@MukulKolpe all yours :)

cnasc:
I wonder if instead of growing the URL field like in #321, it would make sense to add the URL the frame response came from the the validation results? Thinking of this for two reasons:

1. Chances are when debugging the frame, you're going to look at the validation results. If you're already focused on that part of the screen, it's a better UX to have the relevant info there.
2. I think when we expand the input itself, the layout looks a little weird

@briandoyle81CB @bansalgautam thoughts on this?

bansalgautam:
Yeah, I think that would provide a better UX. I tried it, here is the result
![image](https://github.com/coinbase/onchainkit/assets/108223901/27c8e896-5846-455c-aa3d-a2f8706b7422)

cnasc:
@bansalgautam nice! I would just follow the format of the other cells with the key on top then the value on the bottom but other than that LGTM

cpcramer:
Frames have been deprecated.

PR #321, opened 2024-05-03
feat: Made Framegear URL field autogrow

bansalgautam:
**What changed? Why?**
This has been changed as per request by @briandoyle81CB referencing issue #319 

**Notes to reviewers**
Change can be better understood from following video

https://github.com/coinbase/onchainkit/assets/108223901/dc78b7e0-c690-46de-b0f7-402a9ed1bc51

**How has it been tested?**

### DIFF
--- a/framegear/components/FrameInput/FrameInput.tsx
+++ b/framegear/components/FrameInput/FrameInput.tsx
@@ -1,44 +1,27 @@
 import { fetchFrame } from '@/utils/fetchFrame';
 import { frameResultsAtom } from '@/utils/store';
 import { useAtom } from 'jotai';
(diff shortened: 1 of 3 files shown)


Zizzamia:
I like that we have a textarea. But probably we could reduce the space of the "fetch" button, so the link gets more with.

@bansalgautam any other design improvement we could do? So the experience feels more smooth.

bansalgautam:
Yes, we can just go with our orignal design and show the URL in frame validations as suggested by @cnasc [here](https://github.com/coinbase/onchainkit/issues/319#issuecomment-2093510992)

Zizzamia (approved):
Amazing, thank you for the help.

PR #379, opened 2024-05-21
feat: add CDP api key and rpc url to `OnchainKitConfig` and `OnchainKitProvider`

0xAlec:
**What changed? Why?**
- add CDP API key to config and provider
- generate RPC URL automatically based on the `chain` and `apiKey` config
- add tests

**Notes to reviewers**

**How has it been tested?**
`yarn test`


src/OnchainKitConfig.test.ts
```diff
@@ -11,24 +11,40 @@ describe('OnchainKitConfig', () => {
     expect(setOnchainKitConfig).toBeDefined();
   });
 
+  it('should require an api key', () => {
+    const chain = baseSepolia;
+    expect(() => setOnchainKitConfig({ chain })).toThrow(
+      'API Key is required to use OnchainKit - get your API key from the Coinbase Developer Platform: https://portal.cdp.coinbase.com/products/templates',
```
0xAlec, line 17: the URL will be changed once we add the API Key token by itself

src/types.ts
```diff
@@ -7,6 +7,8 @@ import { EASSchemaUid } from './identity/types';
  */
 export type OnchainKitConfig = {
   address: Address | null; // Address is optional as we may not have an address for new users
+  apiKey: string; // ApiKey is required to use CDP APIs
+  rpcUrl: string | null; // RPC URL is automatically set based on the chain and API Key provided in the configuration
```
Zizzamia, line 11: nit. let's keep the key alphabetically order.

src/types.ts
```diff
@@ -7,6 +7,8 @@ import { EASSchemaUid } from './identity/types';
  */
 export type OnchainKitConfig = {
   address: Address | null; // Address is optional as we may not have an address for new users
+  apiKey: string; // ApiKey is required to use CDP APIs
```
Zizzamia, line 10: People reading CDP, will have not clue what is it. Let's probablly spell it out.

Zizzamia: also, do we think we might need more explicit here? like `coinbaseApiKey`? this is more an open question. what do we think?

src/types.ts
Zizzamia: Who sets this automitaclly?

src/types.ts
```diff
@@ -20,6 +22,7 @@ export type OnchainKitContextType = OnchainKitConfig;
 
 export type OnchainKitProviderProps = {
   address?: Address;
+  apiKey: string;
```
Zizzamia, line 25: should this be optional? Not all components or utilities will need an ApiKey, and some folks could use OnchainKit without the need of this.

src/OnchainKitProvider.tsx
```diff
@@ -10,16 +10,24 @@ export const OnchainKitContext = createContext<OnchainKitContextType>(ONCHAIN_KI
...
+  apiKey,
   chain,
   children,
   schemaId,
 }: OnchainKitProviderProps) {
   if (schemaId && !checkHashLength(schemaId, 64)) {
     throw Error('EAS schemaId must be 64 characters prefixed with "0x"');
   }
+  if (!apiKey) {
+    throw Error(
```
Zizzamia, line 22: is it? I don't think so!!!

ApiKey is required only in a few utilities, and for those we should alert missing the apiKey.

OnchainKit will always be bigger than what the Api offers.

src/OnchainKitProvider.tsx
```diff
@@ -10,16 +10,24 @@ export const OnchainKitContext = createContext<OnchainKitContextType>(ONCHAIN_KI
...
+      'API Key is required to use OnchainKit - get your API key from the Coinbase Developer Platform: https://portal.cdp.coinbase.com/products/templates',
+    );
+  }
   const value = useMemo(() => {
     return {
       address: address ?? null,
+      apiKey: apiKey,
+      rpcUrl: `https://api.developer.coinbase.com/rpc/v1/${ONCHAIN_KIT_CONFIG.chain.name.replace(' ', '-').toLowerCase()}/${ONCHAIN_KIT_CONFIG.apiKey}`,
```
Zizzamia, line 30: So this means the `rpcUrl` is always CDP related? 🤔 what if I want something custom, and I want to use OnchainKit with my personal RPC.

Also all like 30 should be `getRpcUrl`, this all interpolation and mixing is enough logic that should be is all function.

src/OnchainKitConfig.ts
```diff
@@ -25,5 +27,13 @@ export const getOnchainKitConfig = <K extends keyof typeof ONCHAIN_KIT_CONFIG>(
  */
 export const setOnchainKitConfig = (properties: SetOnchainKitConfig) => {
   Object.assign(ONCHAIN_KIT_CONFIG, properties);
+  if (!ONCHAIN_KIT_CONFIG.apiKey) {
```
Zizzamia, line 30: wrong assumption, left a comment after on this.

src/OnchainKitConfig.ts
```diff
@@ -25,5 +27,13 @@ export const getOnchainKitConfig = <K extends keyof typeof ONCHAIN_KIT_CONFIG>(
...
+    throw new Error(
+      'API Key is required to use OnchainKit - get your API key from the Coinbase Developer Platform: https://portal.cdp.coinbase.com/products/templates',
+    );
+  }
+  if (properties.apiKey || properties.chain) {
+    ONCHAIN_KIT_CONFIG.rpcUrl = `https://api.developer.coinbase.com/rpc/v1/${ONCHAIN_KIT_CONFIG.chain.name.replace(' ', '-').toLowerCase()}/${ONCHAIN_KIT_CONFIG.apiKey}`;
```
Zizzamia, line 36: mention after this should be his own get function.

src/OnchainKitConfig.ts
0xAlec: removed mandatory api key

src/OnchainKitConfig.ts
0xAlec: added `getRpcUrl`

src/OnchainKitProvider.tsx
0xAlec: added default value for `rpcUrl` (if not provided and if API key is set)

src/OnchainKitProvider.tsx
0xAlec: removed mandatory api key

src/types.ts
0xAlec: removed mandatory api key

src/types.ts
0xAlec: I think `apiKey` makes sense. Don't people already associate OnchainKit with the Coinbase brand? No need to explicitly call out imo

src/OnchainKitConfig.ts
```diff
@@ -5,7 +5,9 @@ import { OnchainKitConfig, SetOnchainKitConfig } from './types';
 // but only acccessed through the get and set functions.
 export const ONCHAIN_KIT_CONFIG: OnchainKitConfig = {
   address: null,
+  apiKey: '',
   chain: baseSepolia,
+  rpcUrl: '',
```
Zizzamia, line 10: I think is best we have `null` as negative. So the type here is either `string | null`.

src/OnchainKitConfig.ts
```diff
@@ -27,3 +29,19 @@ export const setOnchainKitConfig = (properties: SetOnchainKitConfig) => {
   Object.assign(ONCHAIN_KIT_CONFIG, properties);
   return getOnchainKitConfig;
 };
+
+/**
+ * Access the RPC URL for OnchainKit.
+ * Defaults to using Coinbase Developer Platform if a RPC URL is not provided.
+ */
+export const getRPCUrl = () => {
```
Zizzamia, line 37: Let's make `getRPCUrl` his own file.

Zizzamia: Also, I like where this is going.

src/OnchainKitProvider.tsx
```diff
@@ -20,9 +22,15 @@ export function OnchainKitProvider({
   const value = useMemo(() => {
     return {
       address: address ?? null,
+      apiKey: apiKey ?? null,
+      rpcUrl:
+        rpcUrl ??
+        (apiKey
+          ? `https://api.developer.coinbase.com/rpc/v1/${chain.name.replace(' ', '-').toLowerCase()}/${apiKey}`
```
Zizzamia, line 29: We should us the new `getRpcUrl` here.

src/OnchainKitProvider.tsx
```diff
@@ -10,6 +10,8 @@ export const OnchainKitContext = createContext<OnchainKitContextType>(ONCHAIN_KI
  */
 export function OnchainKitProvider({
   address,
+  apiKey,
+  rpcUrl,
```
Zizzamia, line 14: let's keep the alphabetical order

src/types.ts
```diff
@@ -7,6 +7,8 @@ import { EASSchemaUid } from './identity/types';
  */
 export type OnchainKitConfig = {
   address: Address | null; // Address is optional as we may not have an address for new users
+  apiKey: string | null; // ApiKey for Coinbase Developer Platform APIs
+  rpcUrl: string | null; // RPC URL for onchain requests. Defaults to using CDP Node if the API Key is set
```
Zizzamia, line 11: alphabetical

src/OnchainKitProvider.tsx
0xAlec: does the context also make a call to `setOnchainKitConfig`?

every method i am implementing depends on the `ONCHAIN_KIT_CONFIG` object, I'm not sure if this configuration is passed by the provider/context

0xAlec: basically if I use `getRpcUrl` here it will error out because the `ONCHAIN_KIT_CONFIG` object hasn't been set yet

src/RPC.ts
```diff
@@ -0,0 +1,17 @@
+import { ONCHAIN_KIT_CONFIG } from './OnchainKitConfig';
+
+/**
+ * Access the RPC URL for OnchainKit.
+ * Defaults to using Coinbase Developer Platform if a RPC URL is not provided.
+ */
+export const getRPCUrl = () => {
```
Zizzamia, line 7: the file should be named `getRPCUrl.ts`, as in this way it's explicit what it does. Have it call RPC just creates confusion.

Zizzamia:
@0xAlec the test converage drop, make sure to have it back to 100%

Zizzamia (approved):
Great work!!!

PR #381, opened 2024-05-22
feat: scaffold Swap API request and types 

0xAlec:
**What changed? Why?**
- add types for Swap API
- add `swapAPIRequest` and JSONRPC helper functions
- add tests

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/queries/request.ts
+++ b/src/queries/request.ts
@@ -1,0 +1,66 @@
+import { getRPCUrl } from '../getRPCUrl';
+
+export type JSONRPCRequest<T> = {
+  jsonrpc: string;
+  method: string;
+  params: T[];
+  id: number;
+};
+
+export type JSONRPCError = {
+  code: number;
+  message: string;
+};
+
+export type JSONRPCResult<T> = {
+  jsonrpc: string;
+  result: T;
+  error?: JSONRPCError;
+  id: number;
+};
+
+const POST_METHOD = 'POST';
+const JSON_HEADERS = {
+  'Content-Type': 'application/json',
+};
+const JSON_RPC_VERSION = '2.0';
+
+/**
+ * Builds a JSON-RPC request body.
+ *
+ * @param method - The method name.
+ * @param params - The parameters for the method.
+ * @returns The JSON-RPC request body.
+ * @template T - The type of the parameters.
+ */
+export function buildRequestBody<T>(method: string, params: T[]): JSONRPCRequest<T> {
+  return {
+    jsonrpc: JSON_RPC_VERSION,
+    method: method,
+    params: params,
+    id: 1,
+  };
+}
+
+/**
+ * Sends a JSON-RPC request to configured RPC URL. Defaults to using the Coinbase Developer Platform Node.
+ *
+ * @param body - The JSON-RPC request body.
+ * @returns A promise that resolves to the JSON-RPC response.
+ * @throws If an error occurs while sending the request.
+ */
+export async function sendRequest<T, V>(body: JSONRPCRequest<T>): Promise<JSONRPCResult<V>> {
+  try {
+    const url = getRPCUrl();
+    const response = await fetch(url, {
+      method: POST_METHOD,
+      body: JSON.stringify(body),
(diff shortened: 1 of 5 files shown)


src/trade/types.ts
```diff
@@ -0,0 +1,80 @@
+import { Address } from 'viem';
+
+/**
+ * Note: exported as public Type
+ */
+export type Token = {
```
Zizzamia, line 6: all types make sure they follow alphabetical `key` order

src/trade/types.ts
```diff
@@ -0,0 +1,80 @@
...
+ */
+export type Trade = {
+  tx: Transaction;
+  fee: Fee;
+  approveTx?: Transaction;
+  chainId: string;
+};
+
+/**
+ * Note: exported as public Type
```
Zizzamia, line 57: Let's add the comments only when we actually exported, right now you are not exporting them.

src/queries/request.ts
```diff
@@ -0,0 +1,66 @@
+import { getRPCUrl } from '../getRPCUrl';
+
+export type JSONRPCRequest<T> = {
```
Zizzamia, line 3: All types `key` should be alphabetical.

src/queries/request.ts
```diff
@@ -0,0 +1,66 @@
...
+    const url = getRPCUrl();
+    const response = await fetch(url, {
+      method: POST_METHOD,
+      body: JSON.stringify(body),
+      headers: JSON_HEADERS,
+    });
+    const data: JSONRPCResult<V> = await response.json();
+    return data;
+  } catch (error) {
+    console.log(`Error sending request: ${(error as Error).message}`);
```
Zizzamia, line 63: Longs should always start with the function calling the log, so we do in this case
`sendRequest: ~~~`

This will help better understand what's going on.

src/queries/request.ts
```diff
@@ -0,0 +1,66 @@
...
+    throw error;
```
Zizzamia, line 64: Should we really throw an error? it seems quite aggressive.

A library should a way to buble up errors, and then let the user decide what to do with it. This throw is kind of hidden from the kowledge of the devloper, and a soft failure which it is, instead is threated life a hard failure. Which breaks trust between the developer and the library.

src/queries/swap.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @param method - The method to be called in the API.
+ * @param params - The parameters to be passed in the API request.
+ * @returns A promise that resolves to the result of the API request.
+ * @throws An error if the API request fails.
+ */
+export async function swapAPIRequest<T, V>(method: string, params: T[]): Promise<V> {
+  const body = buildRequestBody<T>(method, params);
+  const res = await sendRequest<T, V>(body);
+  if (res.error) {
+    throw new Error(res.error.message);
```
Zizzamia, line 15: Not a fan of all those throw errors, we need a better way to buble up and handling failure inside the library.

Zizzamia: I think there is a meta question to ask which is, what do we want to expose to the developer and what's the expeirence we want them to have.

src/queries/request.ts
0xAlec: viem throws for a jsonrpc error:
https://viem.sh/docs/error-handling

src/queries/swap.ts
0xAlec: viem throws for a jsonrpc error:
https://viem.sh/docs/error-handling

0xAlec: we can add a custom error type `SwapAPIError`

src/queries/request.ts
0xAlec: added

src/queries/swap.ts
0xAlec: this specific `throw` would be for a JSONRPC error (i.e. the request success and there is an `error` field on the response)

Zizzamia (approved):
Amazing

PR #383, opened 2024-05-22
feat: TokenChip component

kyhyco:
**What changed? Why?**
- add `TokenChip` component

**Notes to reviewers**

**How has it been tested?**
locally

### DIFF
--- a/site/docs/pages/token/types.mdx
+++ b/site/docs/pages/token/types.mdx
@@ -1,6 +1,30 @@
 ---
 title: Token Kit Types
-deescription: Glossary of Types in Token Kit.
+description: Glossary of Types in Token Kit.
 ---
 
 # Types [Glossary of Types in Token Kit.]
+
+## `Token`
+
+```ts
+type Token = {
+  name: string; // Token full name e.g. Ether
+  currencyCode: string; // Token symbol e.g. ETH
+  imageURL: string; // Link to the token image
+  blockchain: string; // Blockchain that the token is on
+  address: Address; // The token address on the given blockchain
+  chainId: number; // Blockchain ID
+  decimals: number; // Decimals for the token
+  uuid: string; // ID on ??
+};
+```
+
+## `TokenChipReact`
+
+```ts
+type TokenChipReact = {
+  token: Token; // Rendered token
+  onClick: (token: Token) => void; // Click handler that will use token prop as the argument
+};
+```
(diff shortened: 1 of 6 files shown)


src/token/types.ts
```diff
@@ -1 +1,19 @@
...
+  blockchain: string;
+  address: Address;
+  chainId: number;
+  decimals: number;
+  uuid: string;
+};
+
+export type TokenChipReact = {
+  token: Token;
+  onClick: (token: Token) => void;
```
Zizzamia, line 18: @kyhyco open question, do we really think `onClick` is need it here? After all someone could just add an `onClick` from the outer component.

src/token/types.ts
```diff
@@ -1 +1,19 @@
...
+  imageURL: string;
+  blockchain: string;
+  address: Address;
+  chainId: number;
+  decimals: number;
+  uuid: string;
+};
+
+export type TokenChipReact = {
+  token: Token;
```
Zizzamia, line 17: ok I like `token` is the entity, and than we can play with that.

src/token/types.ts
```diff
@@ -1 +1,19 @@
 // 🌲☀️🌲
+
+import { Address } from 'viem';
+
+export type Token = {
```
Zizzamia, line 5: let's keep everything alphabetical

src/token/types.ts
```diff
@@ -1 +1,19 @@
...
+  name: string;
+  currencyCode: string;
+  imageURL: string;
+  blockchain: string;
```
Zizzamia, line 9: I don't think we need this. As OnchainKit is mostly meant for ETH at this point.

site/docs/pages/token/types.mdx
```diff
@@ -1,6 +1,30 @@
...
+```ts
+type Token = {
+  name: string; // Token full name e.g. Ether
+  currencyCode: string; // Token symbol e.g. ETH
+  imageURL: string; // Link to the token image
+  blockchain: string; // Blockchain that the token is on
+  address: Address; // The token address on the given blockchain
+  chainId: number; // Blockchain ID
+  decimals: number; // Decimals for the token
+  uuid: string; // ID on ??
```
Zizzamia, line 19: I don't think we need `uuid` at the moment.

src/token/components/TokenChip.tsx
```diff
@@ -0,0 +1,38 @@
+import { CSSProperties } from 'react';
+import { TokenChipReact } from '../types';
+
+const styles = {
```
Zizzamia, line 4: I like a lot how you organized the style.

@abcrane123, let's do something similar for your component as well.

Zizzamia:
@kyhyco thank you starting this, I am going to merge it and polish it with a separete PR.

PR #389, opened 2024-05-26
feat: `getTokens`

0xAlec:
**What changed? Why?**
- refactor `queries/` (delete the `swap.ts` files, duplicate functionality to `sendRequest`)
- add `getTokens` function
- add `RawTokenData` type for raw data response from the Swap API, and map it to OnchainKit's `Token` type
- add `GetTokensOptions`
- add tests

**Notes to reviewers**

**How has it been tested?**

in a `src/examples/tokens.ts` file:

```
import { base } from 'viem/chains';
import { setOnchainKitConfig } from '../OnchainKitConfig';
import { getTokens } from '../token/getTokens';

// Initialize the OnchainKit SDK
setOnchainKitConfig({ apiKey: 'redacted', chain: base });

const tokens = await getTokens({ limit: '1', search: 'degen' });
console.log(tokens);
```

```
onchainkit % npx tsx src/examples/swap.ts
[
  {
    address: '0x4ed4e862860bed51a9570b96d89af5e1b0efefed',
    chainId: 8453,
    decimals: 18,
    image: 'https://d3r81g40ycuhqg.cloudfront.net/wallet/wais/3b/bf/3bbf118b5e6dc2f9e7fc607a6e7526647b4ba8f0bea87125f971446d57b296d2-MDNmNjY0MmEtNGFiZi00N2I0LWIwMTItMDUyMzg2ZDZhMWNm',
    name: 'DEGEN',
    symbol: 'DEGEN'
  }
]
```


src/token/getTokens.ts
```diff
@@ -0,0 +1,57 @@
...
+    limit: '50',
+    page: '1',
+  };
+
+  const filters = { ...defaultFilter, ...options };
+
+  try {
+    const res = await sendRequest<GetTokensOptions, RawTokenData[]>(ListSwapAssets, [filters]);
+    if (res.error) {
+      throw new Error(`getTokens: ${res.error.message}`);
```
0xAlec, line 41: this error would be a JSON-RPC error (request was successful and server returned a response, but e.g. a parameter was invalid)

src/token/getTokens.ts
```diff
@@ -0,0 +1,57 @@
...
+      address: token.address,
+      chainId: token.chainId,
+      decimals: token.decimals,
+      image: token.imageURL,
+      name: token.name,
+      symbol: token.currencyCode,
+    }));
+  } catch (error) {
+    console.log(`getTokens: error retrieving tokens: ${(error as Error).message}`);
+    throw error;
```
0xAlec, line 55: this error would be for internal server errors, e.g. remote server is down, or `fetch` fails due to whatever reasons

src/token/getTokens.ts
0xAlec: one thing to consider is if we want to throw an error, or return a `JSONRPCError` object

instead of using `try/catch`, users can try to cast the response to this object and handle the error appropriately
```
import { JSONRPCError } from '@onchainkit/queries/request';

const resp = await getTokens({ limit: '1', search: 'degen' });

if (resp as JSONRPCError) {
  console.error('error getting tokens', resp);
} else {
  console.log(resp);
}
```

src/token/getTokens.ts
```diff
@@ -0,0 +1,57 @@
...
+    page: '1',
+  };
+
+  const filters = { ...defaultFilter, ...options };
+
+  try {
+    const res = await sendRequest<GetTokensOptions, RawTokenData[]>(ListSwapAssets, [filters]);
+
+    if (res.error) {
+      return res.error;
```
0xAlec, line 42: one thing to consider is if we want to throw an error, or return a `JSONRPCError` object

instead of using `try/catch`, users can try to cast the response to this object and handle the error appropriately
```
import { JSONRPCError } from '@onchainkit/queries/request';

const resp = await getTokens({ limit: '1', search: 'degen' });

if (resp as JSONRPCError) {
  console.error('error getting tokens', resp);
} else {
  console.log('tokens:', resp);
}
```

0xAlec: this is for the case where the server responds with JSON-RPC, but there is an issue with the request (e.g. unauthenticated, over rate limit, invalid param, etc.)

vs. if the `fetch` call fails due to the server being down, etc. in which it will throw an error instead

src/definitions/swap.ts
```diff
@@ -0,0 +1 @@
+export const ListSwapAssets = 'cdp_listSwapAssets';
```
Zizzamia, line 1: oh nit, love how we can keep definitions clean

src/token/getTokens.ts
```diff
@@ -0,0 +1,57 @@
+import { ListSwapAssets } from '../definitions/swap';
+import { Token, RawTokenData, GetTokensOptions } from './types';
+import { sendRequest, JSONRPCError } from '../queries/request';
+
+/**
+ * Retrieves a list of tokens on Base.
+ * 
+ * @param {GetTokensOptions} options - An optional object that can be used to filter the list of tokens.
```
Zizzamia, line 8: Btw, to avoid duplication of docs, and kind of over work on writing them. We can avoid write the @parm here and just use the `site/docs/pages` as the only place where people should read the docs.

I think it's fair to still keep a description here, just for general context. But all the types live in the main doc site.

src/token/getTokens.ts
```diff
@@ -0,0 +1,57 @@
...
+//     address: '0x4ed4e862860bed51a9570b96d89af5e1b0efefed',
+//     chainId: 8453,
+//     decimals: 18,
+//     image: 'https://d3r81g40ycuhqg.cloudfront.net/wallet/wais/3b/bf/3bbf118b5e6dc2f9e7fc607a6e7526647b4ba8f0bea87125f971446d57b296d2-MDNmNjY0MmEtNGFiZi00N2I0LWIwMTItMDUyMzg2ZDZhMWNm',
+//     name: 'DEGEN',
+//     symbol: 'DEGEN'
+//   }
+// ]
+*/
+export async function getTokens(options?: GetTokensOptions): Promise<Token[] | JSONRPCError> {
```
Zizzamia, line 29: I like this a lot, very clean on how `options` are used and the response is either A or B.

src/token/getTokens.ts
Zizzamia: I don't think we should expose `import { JSONRPCError } from '@onchainkit/queries/request';` even thought might be the most Eng-Correct way, it doesn't feel simple enough.

I feel like we should expose a more error specific type for each API. so something
```
import { GetTokensAPIError } from '@coinbase/onchainkit/token'
```

so in some way even if it's a bit reduant if fills more explicit.

src/token/index.ts
```diff
@@ -1,2 +1,3 @@
 // 🌲☀️🌲
-export type { Token, TokenChipReact } from './types';
+export type { Token, TokenChipReact, GetTokensOptions } from './types';
```
Zizzamia, line 2: Types should always be the last exported.

src/token/types.ts
```diff
@@ -13,9 +13,32 @@ export type Token = {
   symbol: string; // A ticker symbol or shorthand, up to 11 characters
 };
 
+// The raw response from the Swap API
+export type RawTokenData = {
```
Zizzamia, line 17: `RawTokenData` I feel like this can be a bit more explicit as name.

Zizzamia: also should be alphabetic the type.

src/token/types.ts
```diff
@@ -13,9 +13,32 @@ export type Token = {
...
+  name: string;
+  address: Address;
+  currencyCode: string;
+  decimals: number;
+  imageURL: string;
+  blockchain: string;
+  aggregators: string[];
+  swappable: boolean;
+  unverified: boolean;
+  chainId: number;
```
Zizzamia, line 27: How `aggregators`, `swappable`, and  `unverified` work? And should we expose in the Token type?

src/token/types.ts
```diff
@@ -13,9 +13,32 @@ export type Token = {
...
   token: Token;
 };
+
+/**
+ * Note: exported as public Type
+ */
+export type GetTokensOptions = {
+  limit?: string;
+  search?: string;
+  page?: string;
```
Zizzamia, line 43: Is search a filter for both `address`, `name` and `symbol`?

src/token/types.ts
0xAlec: renamed to `LegacyTokenData`

0xAlec: and sorted

src/token/types.ts
0xAlec: `aggregators` can be ignored, `swappable` is whether the asset can be swapped, and `unverified` - believe there is some sort of verification process for tokens, I'm unsure if this is on the Coinbase side.

I can ask the Wallet team for more info

src/token/getTokens.ts
0xAlec: how about this?

```
import { getTokens } from '../token/index';

const resp = await getTokens({ limit: '1', search: 'degen' });

if ('error' in resp) {
  console.error('error getting tokens', (resp as GetTokensError).error);
} else {
  console.log(resp);
}
```

```
alec@TJ03GKYH46 onchainkit % npx tsx src/examples/swap.ts
error getting tokens 'unauthorized - invalid key'
alec@TJ03GKYH46 onchainkit % npx tsx src/examples/swap.ts
[
  {
    address: '0x4ed4e862860bed51a9570b96d89af5e1b0efefed',
    chainId: 8453,
    decimals: 18,
    image: 'https://d3r81g40ycuhqg.cloudfront.net/wallet/wais/3b/bf/3bbf118b5e6dc2f9e7fc607a6e7526647b4ba8f0bea87125f971446d57b296d2-MDNmNjY0MmEtNGFiZi00N2I0LWIwMTItMDUyMzg2ZDZhMWNm',
    name: 'DEGEN',
    symbol: 'DEGEN'
  }
]
```

0xAlec: https://github.com/microsoft/TypeScript/issues/10485

src/token/types.ts
Zizzamia: should `swappable` potentially being exposed as `isSwappable: boolean`? it seems a useful data point.

0xAlec: yes, but let's leave that for a future PR. let me double check with wallet team **_when_** assets are not `swappable` and we can add that to the documentation. i would assume most assets are by default, swappable

src/token/getTokens.ts
Zizzamia: yeah I like this, we can get the Frontend Fellowship share some thoughts on our types, but overall this is better on what we had.

src/token/types.ts
Zizzamia: dope

src/token/types.ts
0xAlec: yea, looks like it works

```
const resp = await getTokens({ limit: '1', search: '0x4ed4e862860bed51a9570b96d89af5e1b0efefed' });
```

```
onchainkit % npx tsx src/examples/swap.ts
[
  {
    address: '0x4ed4e862860bed51a9570b96d89af5e1b0efefed',
    chainId: 8453,
    decimals: 18,
    image: 'https://d3r81g40ycuhqg.cloudfront.net/wallet/wais/3b/bf/3bbf118b5e6dc2f9e7fc607a6e7526647b4ba8f0bea87125f971446d57b296d2-MDNmNjY0MmEtNGFiZi00N2I0LWIwMTItMDUyMzg2ZDZhMWNm',
    name: 'DEGEN',
    symbol: 'DEGEN'
  }
]
```

src/token/types.ts
Zizzamia: sg

src/token/types.ts
```diff
@@ -19,3 +34,25 @@ export type Token = {
...
+ */
+export type GetTokensError = {
+  code: number;
+  error: string;
+};
+
+/**
+ * Note: exported as public Type
+ */
+export type GetTokensResponse = Token[] | GetTokensError;
```
Zizzamia, line 58: I like this

Zizzamia (approved):

PR #403, opened 2024-05-29
feat: Add isBase Utility

cpcramer:
**What changed? Why?**
Add isBase utility for sponsored paymaster requests.
- Add isBase unit tests.
- Add isBase options and response types.
- Add and update docs
Screenshot 2024-05-29 at 5 00 23 PM

Screenshot 2024-05-29 at 5 00 33 PM

Screenshot 2024-05-29 at 5 00 43 PM

**Notes to reviewers**

**How has it been tested?**
Unit tests.


src/wallet/utils/validChainId.ts
```diff
@@ -0,0 +1,15 @@
+import { validChainIdOptions, validChainIdResponse } from '../types';
+import { baseSepolia, base } from 'viem/chains';
+
+/**
+ * validChainId
+ *  - Checks if the paymaster operations chain id is valid
+ *  - Only allows the base and baseSepolia chain ids
+ */
+export function validChainId({ chainId }: validChainIdOptions): validChainIdResponse {
```
cpcramer, line 9: Curious about your thoughts of naming convention. 

Planning for `validChainId`, `validEntryPoint`, etc. 

We might want to add `sponsor` or `paymaster` such a `validSponsoredChainId`, `validSponsoredEntryPoint`, etc.

Zizzamia: If something returns "something" a verb is usually a best practice, so in this case I will just call it `isBase`.

src/wallet/utils/validChainId.ts
```diff
@@ -0,0 +1,15 @@
...
+  if (chainId !== baseSepolia.id && chainId !== base.id) {
+    return { isValid: false, error: 'Invalid chain id', code: 1 };
```
Zizzamia, line 11: @0xAlec @cpcramer  there is something here we should probably standardize on how `code` should be interepreted.

I can easily imagine how the `code` has to be unique across all OnchainKit. So if that is the case `code` should be a `string`. And maybe somehting like `W_ERR_1` where the first character is about the Kit prefix, and than there is a number.

Or how others do?

yarn.lock
```diff
@@ -29,143 +29,124 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@babel/code-frame@npm:^7.0.0, @babel/code-frame@npm:^7.10.4, @babel/code-frame@npm:^7.12.13, @babel/code-frame@npm:^7.23.5, @babel/code-frame@npm:^7.24.1, @babel/code-frame@npm:^7.24.2":
-  version: 7.24.2
-  resolution: "@babel/code-frame@npm:7.24.2"
+"@babel/code-frame@npm:^7.0.0, @babel/code-frame@npm:^7.10.4, @babel/code-frame@npm:^7.12.13, @babel/code-frame@npm:^7.24.6":
```
Zizzamia, line 32: Why this `lock` changed?

src/wallet/types.ts
```diff
@@ -4,3 +4,9 @@
 export type ConnectAccountReact = {
   children?: React.ReactNode; // Children can be utilized to display customized content when the wallet is connected.
 };
+
+export type isBaseOptions = {
+  chainId: number;
+};
+
+export type isBaseResponse = { isValid: true } | { isValid: false; error: string; code: string };
```
Zizzamia, line 12: Do we think the wording `isValid` is right here? 🤔

Also, I wonder if we are over thinking this one, and should just be return a boolean, that's it.

src/wallet/utils/isBase.ts
```diff
@@ -0,0 +1,15 @@
+import { isBaseOptions, isBaseResponse } from '../types';
+import { baseSepolia, base } from 'viem/chains';
+
+/**
+ * isBase
+ *  - Checks if the paymaster operations chain id is valid
+ *  - Only allows the base and baseSepolia chain ids
+ */
+export function isBase({ chainId }: isBaseOptions): isBaseResponse {
```
Zizzamia, line 9: Write a doc page as well in the `site` folder please.

And add a changeset patch as well.

src/wallet/types.ts
cpcramer: Well this is my docs example with the current working. I think it's pretty clean! 

```
import { isBase } from '@coinbase/onchainkit/wallet';

const isBaseResponse = await isBase({ chainId: '8453' });

if isBaseResponse.isValid {
  console.log('Chain id is a valid base or base sepolia chain id');
}
```

src/wallet/utils/isBase.ts
cpcramer: Added docs page and changeset!

site/docs/pages/wallet/is-base.mdx
```diff
@@ -0,0 +1,35 @@
+# `isBase`
```
Zizzamia, line 1: Let's move this to `docs/pages/is-base.mdx`

site/docs/pages/wallet/is-base.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+The `isBase` utility is designed to verify if the chain id is a valid base or base sepolia chain id.
```
Zizzamia, line 3: Base and Base Sepolia are name, so let's have them Uppercase

site/docs/pages/wallet/is-base.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+## Usage
+
+:::code-group
+
+```tsx [code]
+import { isBase } from '@coinbase/onchainkit/wallet';
```
Zizzamia, line 10: let's do 
```
import { isBase } from '@coinbase/onchainkit';
```
instead

site/docs/pages/wallet/is-base.mdx
```diff
@@ -0,0 +1,35 @@
...
+
+const isBaseResponse = await isBase({ chainId: '8453' });
+
+if isBaseResponse.isValid {
```
Zizzamia, line 14:
```
if isBaseResponse.isValid {
```
this is not real code

src/wallet/types.ts
Zizzamia: Yeah I think for this one we should be even simpler. Let's just return a `boolean`.

site/docs/pages/wallet/is-base.mdx
cpcramer: What do you mean? Updated to:

```
import { isBase } from '@coinbase/onchainkit';

const isValid = await isBase({ chainId: '8453' });

if (isValid) {
  console.log('The chainId is valid.');
} else {
  console.log('Invalid chainId.');
}
```

site/docs/pages/is-base.mdx
```diff
@@ -0,0 +1,33 @@
...
+if (isValid) {
+  console.log('The chainId is valid.');
+} else {
+  console.log('Invalid chainId.');
+}
+```
+
+```ts [return value]
+true;
+```
```
cpcramer, line 23: It feels like this can be improved...

site/docs/pages/is-base.mdx
```diff
@@ -0,0 +1,33 @@
...
+The `isBase` utility is designed to verify if the chain id is a valid Base or Base Sepolia chain id.
+
+## Usage
+
+:::code-group
+
+```tsx [code]
+import { isBase } from '@coinbase/onchainkit';
+
+const isValid = await isBase({ chainId: '8453' });
```
Zizzamia, line 12: Btw `isBase` does not have a promise 😉

Zizzamia (approved):
Great work, I am going to merge this and do some polish.

PR #409, opened 2024-05-29
feat: Add Check Entrypoint

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


site/docs/pages/is-base.mdx
```diff
@@ -9,7 +9,7 @@ The `isBase` utility is designed to verify if the chain id is a valid Base or Ba
 ```tsx [code]
 import { isBase } from '@coinbase/onchainkit';
 
-const isValid = await isBase({ chainId: '8453' });
+const isValid = isBase({ chainId: '8453' });
```
Zizzamia, line 12: you want to rebase, I already fix this

src/wallet/isEntrypoint.ts
```diff
@@ -0,0 +1,10 @@
+import { ENTRYPOINT_ADDRESS_V06 } from 'permissionless';
+import { isEntrypointOptions, isEntrypointResponse } from './types';
+
+export function isEntrypoint({ entrypoint }: isEntrypointOptions): isEntrypointResponse {
```
Zizzamia, line 4: I am not sure this is the right wording for this function. `is` `Entrypoint` is way too generic.

What are we actually trying to do here?

site/docs/pages/is-base.mdx
cpcramer: Thanks! rebased 👍

src/wallet/isEntrypoint.ts
Zizzamia: What is Entrypoint anyway?

cpcramer: We are trying to validate that the specified entrypoint contract is correct. It should be "version 6"

cpcramer: `isEntrypointContractV6`?

Zizzamia: Entrypoint is the address for something, and back looking to `willSponsor` where the `entrypoint` is coming from?

Is that coming from Paymaster service? or from where?

cpcramer: The entrypoint is specified on the client side when they initialize the `paymasterClient`

site/docs/pages/is-entrypoint.mdx
```diff
@@ -0,0 +1,34 @@
+# `isValidAAEntrypoint`
+
+The `isValidAAEntrypoint` utility is designed to verify if the entrypoint address is a valid v6 entrypoint.
```
Zizzamia, line 3: I think we explain a bit more here what's going.

Like what kind of address we are checking. And where this method should be used.

site/docs/pages/is-entrypoint.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+## Usage
+
+:::code-group
+
+```tsx [code]
+import { isValidAAEntrypoint } from '@coinbase/onchainkit';
+import { ENTRYPOINT_ADDRESS_V06 } from 'permissionless';
```
Zizzamia, line 11: In this example I think is fair to avoid showing this line.

So you could instead have something
```
const AAImplementationAddress = '0x123`; 
const isValid = isValidAAEntrypoint({ entrypoint: AAImplementationAddress });
```
to showcase what you are checking.

otherwise the example feels like
is true === to true :D

site/docs/pages/is-entrypoint.mdx
```diff
@@ -0,0 +1,34 @@
...
+
+```ts [return value]
+true;
+```
+
+:::
+
+## Returns
+
+[`isValidAAEntrypointtResponse`](/types#isvalidaaentrypointtresponse)
```
Zizzamia, line 30: As the response is just a boolean we don't need a custom type, just write boolean like we did in 
https://onchainkit.xyz/is-base#returns

site/sidebar.ts
```diff
@@ -13,6 +13,10 @@ export const sidebar = [
             text: 'isBase',
             link: '/is-base',
           },
+          {
```
Zizzamia, line 16: So for this one, as we are talking about Account Abstraction, and AA is highly connected with Wallet and Paymaster, let's have this under the Walletkit

site/docs/pages/is-entrypoint.mdx
cpcramer: Updated to:

```
The `isValidAAEntrypoint` utility is designed to verify if the target account abstraction singleton address.
This should be used to validate the entrypoint before sponsoring a transaction.
```

site/sidebar.ts
cpcramer: Updated

site/docs/pages/is-entrypoint.mdx
cpcramer: Updated

site/docs/pages/is-entrypoint.mdx
cpcramer: Good point. Updated to:

import { isValidAAEntrypoint } from '@coinbase/onchainkit';

```
const AAImplementationAddress = '0x123`;
const isValid = isValidAAEntrypoint({ entrypoint: AAImplementationAddress });

if (isValid) {
  console.log('The entrypoint is valid.');
} else {
  console.log('Invalid entrypoint.');
}
```

Zizzamia (approved):
Nice!!!

I will take care of some polishing.

PR #414, opened 2024-05-31
feat: TokenChip.css

kyhyco:
**What changed? Why?**
- add TokenChip.css to handle css modifiers (`:hover`)

| Normal | Hover | Active |
| --- | --- | --- |
| Screenshot 2024-05-30 at 5 20 41 PM | Screenshot 2024-05-30 at 5 20 45 PM | Screenshot 2024-05-31 at 4 28 34 PM |

**Notes to reviewers**

**How has it been tested?**
locally

### DIFF
--- a/src/token/components/TokenChip.css
+++ b/src/token/components/TokenChip.css
@@ -1,0 +1,25 @@
+.token-chip__button {
+  border-radius: 16px;
+  padding: 4px 12px 4px 4px;
+  display: flex;
+  align-items: center;
+  background: #eef0f3;
+  width: fit-content;
+}
+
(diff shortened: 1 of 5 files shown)


src/token/components/TokenChip.css
```diff
@@ -0,0 +1,25 @@
+.token-chip__button {
```
Zizzamia, line 1: Should we maybe prefix those with `ock`?

`ock_token-chip__button`?

kyhyco: yes converting

Zizzamia (approved):

PR #415, opened 2024-05-31
feat: internal components

kyhyco:
**What changed? Why?**
- add `src/internal/components` to store internal components shared across OnchainKit that should not be exposed to the user

**Notes to reviewers**

**How has it been tested?**
locally


Zizzamia:
Closing this one, as we end up shipping other PRs.

PR #421, opened 2024-05-31
feat: TokenSearch

kyhyco:
**What changed? Why?**
- add TokenSearch component that enable search with debounce delay

Screenshot 2024-05-31 at 3 00 31 PM

**Notes to reviewers**

**How has it been tested?**
locally


site/docs/pages/token/token-search.mdx
```diff
@@ -0,0 +1,53 @@
...
+:::code-group
+
+```tsx [code]
+import { TokenSearch } from '@coinbase/onchainkit/token';
+
+const tokens = [
+  // ...
+];
+
+<TokenSearch tokens={tokens} onSearch={(tokens) => {}} />; // [!code focus]
```
Zizzamia, line 22: A tricky thing here, is that we should make it easier to do both:
1. API calls of different tokens
2. Use the data that is found

site/docs/pages/token/token-search.mdx
```diff
@@ -0,0 +1,62 @@
...
+      name.toLowerCase().includes(v) ||
+      symbol.toLowerCase().includes(v) ||
+      address.toLowerCase().includes(v)
+    );
+  });
+
+  // Do something with filteredTokens here
+}, [tokens]);
+
+<TokenSearch onChange={handleChange} delayMs={200} />; // [!code focus]
```
kyhyco, line 47: @Zizzamia how about this?

Zizzamia: Oh very interesting. I think we can show the example in two code examples.

Zizzamia (approved):
Dope!!! Great job

PR #438, opened 2024-06-03
feat: TokenImage and TokenSelector components

kyhyco:
**What changed? Why?**

- add TokenImage to render token image
Screenshot 2024-06-03 at 4 28 17 PM

- add TokenSelector to render a button displaying token info or placeholder text
Screenshot 2024-06-03 at 4 29 08 PM

**Notes to reviewers**

**How has it been tested?**
locally

### DIFF
--- a/site/docs/pages/token/introduction.mdx
+++ b/site/docs/pages/token/introduction.mdx
@@ -15,8 +15,10 @@ TokenKit enables you to effortlessly create user experiences using Token compone
 The available components are:
 
 - [`<TokenChip />`](/token/token-chip): Small button component for a given token.
+- [`<TokenImage />`](/token/token-image): Image component for token image.
 - [`<TokenRow />`](/token/token-row): Row component for a given token.
 - [`<TokenSearch />`](/token/token-search): Search component to search by name, symbol and address for a given list of tokens.
+- [`<TokenSelector />`](/token/token-selector): Stylized button component to display token info or placeholder text.
 
 :::code-group
 
--- a/site/docs/pages/token/token-image.mdx
+++ b/site/docs/pages/token/token-image.mdx
@@ -62,12 +62,12 @@ The `TokenImage` component is an image that crops token image to a circle with a
 
 ```tsx [code]
 <TokenImage
-  src="https://dynamic-assets.coinbase.com/dbb4b4983bde81309ddab83eb598358eb44375b930b94687ebe38bc22e52c3b2125258ffb8477a5ef22e33d6bd72e32a506c391caa13af64c00e46613c3e5806/asset_icons/4113b082d21cc5fab17fc8f2d19fb996165bcce635e6900f7fc2d57c4ef33ae9.png"
+  src={null}
   size={16}
 />
 
 <TokenImage
-  src="https://dynamic-assets.coinbase.com/dbb4b4983bde81309ddab83eb598358eb44375b930b94687ebe38bc22e52c3b2125258ffb8477a5ef22e33d6bd72e32a506c391caa13af64c00e46613c3e5806/asset_icons/4113b082d21cc5fab17fc8f2d19fb996165bcce635e6900f7fc2d57c4ef33ae9.png"
+  src={null}
   size={24}
 />
 ```
--- a/site/docs/pages/token/token-image.mdx
+++ b/site/docs/pages/token/token-image.mdx
@@ -1,0 +1,110 @@
+import { TokenImage } from '../../../../src/token/components/TokenImage';
+import App from '../App';
+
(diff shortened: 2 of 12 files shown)


site/docs/pages/token/introduction.mdx
```diff
@@ -15,8 +15,10 @@ TokenKit enables you to effortlessly create user experiences using Token compone
 The available components are:
 
 - [`<TokenChip />`](/token/token-chip): Small button component for a given token.
+- [`<TokenImage />`](/token/token-image): Image component for token image.
 - [`<TokenRow />`](/token/token-row): Row component for a given token.
 - [`<TokenSearch />`](/token/token-search): Search component to search by name, symbol and address for a given list of tokens.
+- [`<TokenSelector />`](/token/token-selector): Stylized button component to display token info or placeholder text.
```
Zizzamia, line 21: Can TokenChip maybe have a selector option 🤔, that's a very temptive idea, but not sure if will make things more complex.

site/docs/pages/token/token-image.mdx
```diff
@@ -0,0 +1,110 @@
...
+  </div>
+</App>
+
+`TokenImage` with null as src
+
+:::code-group
+
+```tsx [code]
+<TokenImage
+  src="https://dynamic-assets.coinbase.com/dbb4b4983bde81309ddab83eb598358eb44375b930b94687ebe38bc22e52c3b2125258ffb8477a5ef22e33d6bd72e32a506c391caa13af64c00e46613c3e5806/asset_icons/4113b082d21cc5fab17fc8f2d19fb996165bcce635e6900f7fc2d57c4ef33ae9.png"
```
Zizzamia, line 65: Wait, where is the `null`? 

Also, we should point out to Shelley that we might need a more cute default image.

Also ideally we should give a way to customize the color....

I guess the custom bg, let's do that later, and see also how Sid think the theme story.

site/docs/pages/token/token-selector.mdx
```diff
@@ -0,0 +1,109 @@
+import { TokenSelector } from '../../../../src/token/components/TokenSelector';
+import App from '../App';
+
+# `<TokenSelector />`
+
+### Warning: component under development, may change in coming weeks
+
+`TokenSelector` component is stylized button component that displays the token info or placeholder text.
```
Zizzamia, line 8: I read this
```
stylized button component that displays the token info or placeholder text.
```
and I am like, "you just described the TokenChip", I understand the selector might have more power, but so far I don't see it. 🤔

src/token/components/TokenSelector.tsx
```diff
@@ -0,0 +1,58 @@
...
+    lineHeight: '1.5',
+    fontWeight: '500',
+    marginLeft: '4px',
+  },
+  caret: {
+    marginLeft: '8px',
+  },
+};
+
+export function TokenSelector({ token, onClick }: TokenSelectorReact) {
```
Zizzamia, line 40: I am very tempted to merge TokenChip into TokenSelector..... what do we think? How other design system offer this? 🤔

abcrane123: i had the same thought but the functionality is so different between the two

Zizzamia: TokenSelector to me feels more a super set of a TokenChip. I guess it's cleaner as component. But at the same time they both end up delegating the `onClick` experience to the dev, to decide.

So in the end the question we ask is not even if you want or not `onClick` but more, do you want a Selector image yes or no.

Ok so TokenChip is pretty much a tag component (https://ant.design/components/tag), where TokenSelector is a select component (https://ant.design/components/select).

I think at first they for sure are very similar, BUT, if we end up building the Tag experience, and Select experience like other Design System do. 

Then yeah, I think keep them split makes-sense for the long-term.

kyhyco: Had similar thoughts. I did initially extend TokenChip with postfix icon but it didn't feel right.

I could see a world where there is multiple options for TokenSelector. Pop up the big token selector component with chips, row and search or just a simple dropdown.

src/token/components/TokenImage.tsx
```diff
@@ -0,0 +1,31 @@
+import { useMemo } from 'react';
+import { TokenImageReact } from '../types';
+
+export function TokenImage({ src, size = 24 }: TokenImageReact) {
+  const styles = useMemo(() => {
+    return {
```
Zizzamia, line 6: @abcrane123 @kyhyco how we style is something we need to figure out soon.

Having React styling can be painful in terms of perf.

Zizzamia (approved):

PR #440, opened 2024-06-04
feat: useSelectSwapTokens hook

kyhyco:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/swap/components/useSelectSwapTokens.ts
+++ b/src/swap/components/useSelectSwapTokens.ts
@@ -1,0 +1,43 @@
+import { useCallback, useState } from 'react';
+import { Token } from '../../token';
+
+type UseSelectSwapTokensHook = {
+  userTokens: Token[];
+  swappableTokens: Token[];
+};
+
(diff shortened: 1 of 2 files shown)


src/swap/types.ts
```diff
@@ -47,3 +47,10 @@ export type TrendingToken = Token & {
   numOfBuys: number;
   numOfSells: number;
 };
+
+export type UseSwap = {
```
kyhyco, line 51: Maybe `UseSwapHook` is more appropriate?

src/swap/components/useSelectSwapTokens.ts
```diff
@@ -0,0 +1,43 @@
+import { useCallback, useState } from 'react';
+import { Token } from '../../token';
+
+type UseSelectSwapTokensHook = {
+  userTokens: Token[];
+  swappableTokens: Token[];
+};
+
+export function useSelectSwapTokens({ userTokens, swappableTokens }: UseSelectSwapTokensHook) {
```
kyhyco, line 9: thinking of keeping this internal for now

PR #450, opened 2024-06-05
docs: make provider page clearer

roushou:
**What changed? Why?**

This PR makes the `/onchainkit-provider` documentation page clearer by:
- highlighting the correct lines
- fixing a typo: `chain="base"` -> `chain={base}`
- removing unnecessary comment line

**Notes to reviewers**

Running `prettier` impacts those changes while it should not

**How has it been tested?**

Launched locally with `yarn dev`


Zizzamia:
Oh yeah format, does that at times. But we still need it :D 

What's your advice in solving this? Btw love those little changes, really appreciate them.

roushou:
Happy to help :+1: 

To correct myself, it seems that's actually how prettier works, I just didn't expect it.

Closest issue I found is this one https://github.com/prettier/prettier/issues/807 which is still open.

It didn't find anything in the [mdx](https://mdxjs.com/packages/mdx/) parser options that could help.

Looking at how some other projects handle this, it seems that they actually don't.
- [vocs](https://github.com/wevm/vocs) (used under-the-hood) uses [biome](https://biomejs.dev) which AFAIK doesn't support md/mdx files
- mdx [ignores](https://github.com/mdx-js/mdx/blob/f639f9a5a3641a30a8936d48e76eed8c2239fdbe/.prettierignore) md/mdx files

My 2cents would be to follow them. The documentation would be clearer this way.

PR #453, opened 2024-06-05
feat: postcss integration + TokenChip.css

kyhyco:
**What changed? Why?**
- add postcss in tailwindcss to support `@import` pattern
- convert TokenChip to css

**Notes to reviewers**

**How has it been tested?**
locally


.gitignore
```diff
@@ -7,7 +7,7 @@ lib
 tmp
 /out-tsc
 site/vocs.config.tsx.timestamp*
-src/styles.css
+styles.css
```
Zizzamia, line 10: to work well, this has to be in `src/`

package.json
```diff
@@ -5,8 +5,8 @@
   "repository": "https://github.com/coinbase/onchainkit.git",
   "license": "MIT",
   "scripts": {
-    "build": "packemon build --addEngines --addFiles --declaration && npx packemon validate --no-license --no-people --no-repo && tailwindcss -i ./src/index.css -o ./src/styles.css --minify",
-    "watch": "packemon watch",
+    "build": "packemon build --addEngines --addFiles --declaration && npx packemon validate --no-license --no-people --no-repo && tailwindcss -i ./src/index.css -o ./src/styles.css --postcss --minify",
```
Zizzamia, line 8: what postcss does?

kyhyco: this will allow us to import files at the root index.css

src/index.css
```diff
@@ -1,9 +1,6 @@
...
 
-@layer components {
-  .ock-label {
-    @apply text-base font-medium leading-4 text-black;
-  }
-}
+@import 'tailwindcss/components';
+@import './token/components/TokenChip.css';
+
+@import 'tailwindcss/utilities';
```
Zizzamia, line 6: I don't think we should include `components` and `utility`, maybe not even `base`.

If we want to avoid to have OnchainKit css override other apps css.

site/docs/pages/token/token-chip.mdx
```diff
@@ -59,3 +55,9 @@ const token = {
 ## Props
 
 [`TokenChipReact`](/token/types#tokenchipreact)
+
+## CSS classes
+
+- `ock-tokenchip-button`
```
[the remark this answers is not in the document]
kyhyco: Should we just link to the css file?

package.json
kyhyco: actually... might not need postcss. Let me play around with it

Zizzamia: ok

kyhyco: We don't need postcss and just use native css import. Removed postcss

site/docs/pages/token/token-chip.mdx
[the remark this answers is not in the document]
Zizzamia: I will say show the code as codeblock

src/index.css
kyhyco: removed

PR #462, opened 2024-06-06
docs: add `bun` usage

roushou:
**What changed? Why?**

Add `bun` usage to code groups. Feels good to have it for bun users.

**Notes to reviewers**

Looks like this

image

**How has it been tested?**

Launched docs locally with `yarn dev`

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,9 @@ npm install @coinbase/onchainkit viem@2.x
 
 # Use PNPM
 pnpm add @coinbase/onchainkit viem@2.x
+
+# Use BUN
+bun add @coinbase/onchainkit viem@2.x
 ```
 
 ## Components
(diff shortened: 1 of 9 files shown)


Zizzamia:
Out of curiosity, have you tested it works well with all OnchainKit stuff?

roushou:
My usage of OnchainKit with so far is:

- Having the Smart Wallet working
- Displaying identity components correctly (both core and react)

Still a lot to try out but this library seems to be mainly a wrapper around Viem, Wagmi and TanStack Query.

They all include Bun in their documentation:

- Viem: https://viem.sh/docs/getting-started#installation
- Wagmi: https://wagmi.sh/vue/installation#package-manager
- TanStack: https://tanstack.com/query/latest/docs/framework/react/installation

So I'm kinda shortcutting.

Only thing I'm not sure of is xmtp since I never used it.

Zizzamia (approved):
Love the way you describe your thinking. Sounds good, let's merge it.

PR #463, opened 2024-06-06
docs: fix `getAvatar(...)` documentation

roushou:
**What changed? Why?**

This fixes the documentation of `getAvatar(...)` as it accepts a `string` and not `{ ensName: string }`.
https://github.com/coinbase/onchainkit/blob/172faf809c87ff85077a013649f8e255a6255e9f/src/identity/core/getAvatar.ts#L5

**Notes to reviewers**

FWIW using a `string` is not very explicit since it could be an `address` too. Same with the wording `getAvatar()`

What do you think of following Viem?

`getAvatar(ensName: string)` -> `getEnsAvatar({ name: string })`

**How has it been tested?**

### DIFF
--- a/site/docs/pages/identity/get-avatar.mdx
+++ b/site/docs/pages/identity/get-avatar.mdx
@@ -15,7 +15,7 @@ Supported providers:
 ```tsx
 import { getAvatar } from '@coinbase/onchainkit/identity';
 
-const avatar = await getAvatar({ ensName: 'vitalik.eth' });
+const avatar = await getAvatar('vitalik.eth');
 ```
 
 ## Returns


Zizzamia:
So the conversation here is a bit bigger.

I think the docs is what we want, but we forgot to update the code.

We want this
```
getAvatar({ ensName: string })
```
because the idea is that you can retrieve the Avatar from different kind of identities, like ENS, Nouns or any new one.

So yeah for now works like Viem behind the scene, but that's more a temporary solution till we implement our final vision.

Similar to `getName` that is 
```
getName = async ({ address, showAddress }: GetName)
```
and using your or suggestion, maybe what we need it
```
getAvatar = async ({ name: string }: GetAvatar)
```


@roushou what do you think? Btw great catch

roushou:
Makes more sense :+1: 

Your wording is more appropriate so let's go this way `getAvatar({ name: string })`.

Closing this PR since it's not what we want. And let's follow this one https://github.com/coinbase/onchainkit/pull/464

PR #464, opened 2024-06-06
feat: standardize `getAvatar()`

roushou:
**What changed? Why?**

This PR makes `getAvatar()` more explicit by:
- changing the wording to `getEnsAvatar()`
- accepting `{ name: string }` in parameter instead of `string`

The previous `getAvatar()` is still there but deprecated to prevent hard breaks for existing users.

FWIW there is a personal opinion involved as names/avatars are "traditionally" not the same as those from ENS. Maybe it's intentional but I think that following both Viem and Wagmi may be a good idea so users coming from there have the same API.

**Notes to reviewers**

This change implies to also make other components follow the same path such as `getName`, `useAvatar`, `useName`.
Will make according changes in separate PR(s) depending on this one going through or not. Just need to make sure to **not** include this change in a release before.

**How has it been tested?**

I checked only the documentation website. My laptop goes OOM when running tests.


Zizzamia:
Just replied here https://github.com/coinbase/onchainkit/pull/463, but I think we want to release a `minor` change, that has a breaking changes.

And do
```
getAvatar = async ({ name: string }: GetAvatar)
```

Overall our goal is not to replicate what Viem has, but instead provide a extra layer that gives more functionality. And in this case in particular is about fetching identity from a new Resolver the we will potentially offer, that gives you a choice on what idenity system you want to use (ENS, Nouns, etc..).

roushou:
Makes sense. Replied in the other PR https://github.com/coinbase/onchainkit/pull/463#issuecomment-2153148783.

Let's go for `getAvatar({ name: string })`

roushou:
Made some changes @Zizzamia

Zizzamia (approved):
Approving this, and probably changing `name` to `ensName` to make it more clear.

PR #465, opened 2024-06-06
getquote

0xAlec:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/definitions/swap.ts
+++ b/src/definitions/swap.ts
@@ -1,1 +1,2 @@
 export const ListSwapAssets = 'cdp_listSwapAssets';
+export const GetSwapQuote = 'cdp_getSwapQuote';
--- a/src/swap/core/getQuote.ts
+++ b/src/swap/core/getQuote.ts
@@ -1,0 +1,30 @@
+import { GetSwapQuote } from '../../definitions/swap';
+import { sendRequest } from '../../queries/request';
+import { GetQuoteResponse, GetQuoteParams, Quote, SwapError } from '../types';
+
+/**
+ * Retrieves a quote for a swap from Token A to Token B.
+ */
+export async function getQuote(params: GetQuoteParams): Promise<GetQuoteResponse> {
+  // Default parameters
+  const defaultParams = {
+    amountReference: 'from',
+  };
+
+  params = { ...defaultParams, ...params };
+
+  try {
+    const res = await sendRequest<GetQuoteParams, Quote>(GetSwapQuote, [params]);
+
+    if (res.error) {
+      return {
+        code: res.error.code,
+        error: res.error.message,
+      } as SwapError;
+    }
+
+    return res.result;
+  } catch (error) {
+    throw new Error(`getQuote: ${error}`);
+  }
+}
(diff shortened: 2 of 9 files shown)


src/token/types.ts
```diff
@@ -56,7 +56,7 @@ export type GetTokensOptions = {
  * Note: exported as public Type
  */
 export type Token = {
-  address: Address; // The address of the token contract
+  address: Address | ''; // The address of the token contract
```
0xAlec, line 59: this is since ETH has `''` as the `Address` field

src/definitions/swap.ts
```diff
@@ -1 +1,2 @@
 export const ListSwapAssets = 'cdp_listSwapAssets';
+export const GetSwapQuote = 'cdp_getSwapQuote';
```
Zizzamia, line 2: Side point, we might want to use the UPPER_CASE standard for those const, and keep them alphabetical order. So do.

```
export const CDP_GET_SWAP_QUOTE = 'cdp_getSwapQuote';
export const CDP_LIST_SWAP_ASSETS = 'cdp_listSwapAssets';
```

src/token/types.ts
kyhyco: is this always going to be true?

src/swap/types.ts
```diff
@@ -1,23 +1,68 @@
 import { Address } from 'viem';
-import { Token } from '../token/types';
+import { LegacyTokenData, Token } from '../token/types';
 
 export type AddressOrETH = Address | 'ETH';
 
+/**
+ * Note: exported as public Type
+ */
+export type GetQuoteParams = {
```
Zizzamia, line 9: For all types, let's add a comment that explain each line, similar to what we did here https://github.com/coinbase/onchainkit/blob/main/src/token/types.ts#L50

And keep all types alphabetical ordered.

src/swap/types.ts
```diff
@@ -1,23 +1,68 @@
...
+  from: Token;
+  to: Token;
+  amount: string;
```
Zizzamia, line 12: What's the difference between `amount` and `amountReference`?

src/swap/core/getQuote.ts
```diff
@@ -0,0 +1,61 @@
...
+    amountReference: 'from',
+  };
+
+  params = { ...defaultParams, ...params };
+
+  // Only pass in the address as a parameter to the request
+  // Defaults to ETH if no address is provided
+  const addressParams: GetQuoteParamsWithAddress = {
+    from: params.from.address || 'ETH',
+    to: params.to.address || 'ETH',
```
Zizzamia, line 28: Why the type can be wither an `Address` or a `string`? I guess the API works like this?

src/token/types.ts
Zizzamia: When we say ETH, are we talking about ETH Base or ETH (L1)?

0xAlec: either, native ETH does not have an address on any network

also this PR isn't ready yet, will be doing some refactoring once i make some upstream changes to the api

Zizzamia: awesome

PR #491, opened 2024-06-08
chore: add ability to declare types for environment variables

roushou:
**What changed? Why?**

Add ability to declare types for environment variables. Less error-prone, easier refactoring.

It's also self-contained which means your LSP should apply it only to the directory the file is created at i.e. it's only applied to the `site` directory here so you don't have conflicts in other packages where other declarations could be defined.

**Notes to reviewers**

https://github.com/coinbase/onchainkit/assets/155655425/70973e26-d51e-48af-b7bd-008edfacb039

**How has it been tested?**


Zizzamia (approved):
Oh sweet

PR #499, opened 2024-06-09
chore: bump github actions

roushou:
**What changed? Why?**

It's nice to be up-to-date. AFAIK there is no breaking changes between v3 and v4 for both `actions/setup-node` and `actions/checkout` besides using Node.js v20 as the default runtime instead of v16 which Node.js dropped out from their long time support.

https://github.com/actions/setup-node/compare/v3...v4.0.0#diff-1243c5424efaaa19bd8e813c5e6f6da46316e63761421b3e5f5c8ced9a36e6b6R36

https://github.com/actions/checkout/compare/v4.0.0...v4.1.0#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R15

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -2,7 +2,7 @@ name: Setup
 runs:
   using: composite
   steps:
-    - uses: actions/setup-node@v3
+    - uses: actions/setup-node@v4
       with:
         node-version-file: '.nvmrc'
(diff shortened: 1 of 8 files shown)


Zizzamia (approved):
Niiiice

PR #506, opened 2024-06-10
feat: Inter fonts

kyhyco:
**What changed? Why?**
- Apply `Inter` font to doc site

Example
| Old | New |
| --- | --- |
|  Screenshot 2024-06-10 at 2 25 00 PM | Screenshot 2024-06-13 at 2 39 42 PM |

**Notes to reviewers**
Tried out several approaches:
- importing `@fontsource/inter`: messes up the css import path and will require creating a separate package just to handle fonts if we go this route
- including Inter font files in OnchainKit: this could be an alternative if we don't want to directly hit google fonts
- using `@tailwind/components` to create custom font components: much better to create a react component

- Use case for tailwind and non-tailwind users is supported in this PR

**How has it been tested?**
locally


site/docs/pages/App.tsx
```diff
@@ -32,7 +32,9 @@ export default function App({ children }: { children: ReactNode }) {
   return (
     <WagmiProvider config={wagmiConfig}>
       <QueryClientProvider client={queryClient}>
-        <div style={{ display: 'flex', flexDirection: 'column' }}>{children}</div>
+        <div style={{ display: 'flex', flexDirection: 'column' }} className="ock-docsite-content">
```
Zizzamia, line 35: Do we want all our demo to shocase Inter?

kyhyco: Great question. Let's check in with design tomorrow

mindapivessa:
@kyhyco I don't think the font applied is Inter? 

Here's Inter: https://fonts.google.com/specimen/Inter

Screenshot 2024-06-11 at 12 26 39 PM

src/text/TextBody.tsx
```diff
@@ -0,0 +1,9 @@
+import { ReactNode } from 'react';
+
```
Zizzamia, line 2: Why `src/text`?

Zizzamia: Oh I see, you are trying to create an internal Design System to use for building all those pieces.

got it!

kyhyco: We could also do something like `src/internal/text`

abcrane123 (approved):
love this !

Zizzamia (approved):

PR #509, opened 2024-06-11
🚧 Test getAddressType

cpcramer:
**What changed? Why?**
Takes in an address and returns if it's an EOA, Smart Wallet, or Smart Contract

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/checkAddressType.ts
+++ b/src/checkAddressType.ts
@@ -1,53 +1,97 @@
-import type { Address, Hex, BlockTag } from 'viem';
-import type { PublicClient } from 'viem';
-import { decodeAbiParameters } from 'viem';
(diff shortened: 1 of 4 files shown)


src/checkAddressType.ts
```diff
@@ -0,0 +1,78 @@
...
+  address: string;
+  client: PublicClient;
+};
+
+type CheckAddressTypeResponse = {
+  type: 'EOA' | 'Coinbase Smart Wallet' | 'Smart Contract';
+  error?: string;
+};
+
+export async function checkAddressType({
```
Zizzamia, line 21: I will probably call it `getAddressType`

PR #514, opened 2024-06-11
feat: refactor documentation

roushou:
This PR is a first step to make the documentation more readable and maintainable. There's a lot of subjectivity so feel free to discard if it doesn't align with your taste. The concerns I have with the current documentation:

- Things are packed and unorganized which make harder to know where to look
- There are a lot of superfluous content like code blocks showing return values `true`, `1000` etc...
- Showing the HTML generated is not necessary following the discussions on this PR https://github.com/coinbase/onchainkit/pull/500
- There are a lot of copy paste and often with some parts tied together which make it hard to maintain

The main reference I have is [Shadcn](https://ui.shadcn.com/) because I find it simple to read. I find [Radix UI](https://www.radix-ui.com/primitives/docs/components/avatar) documentation much better because I don't have to switch between tabs but I don't know how to replicate it using Vocs. One common point is that things have enough spacing between them.

There are still some work so feel free to add your changes. Or merge it if you prefer to do it incrementally.


Zizzamia:
One quick thought, I am not a fan of Demo being hidden by a tab. Mostly because it adds an extra click when seen.

Even thought they feel cleaner, I like see demo components as I scroll, without the need to click.

roushou:
Makes sense. Leaving this PR open and will close if there's no positive feedback in the following days.

roushou:
Closing since nobody seems up for it

PR #543, opened 2024-06-13
chore: Add token amount formatter

abcrane123:
**What changed? Why?**
- format amount returned from `getSwapQuote`

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/swap/utils.ts
+++ b/src/swap/utils.ts
@@ -17,3 +17,11 @@ export function isSwapError(response: unknown): response is SwapError {
     response !== null && typeof response === 'object' && 'error' in response
   );
 }
+
+export function formatTokenAmount(amount: string, decimals: number) {
+  // Convert the string amount to a number using decimals value
+  const numberAmount = Number(amount) / Math.pow(10, decimals);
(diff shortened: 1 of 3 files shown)


src/swap/utils.ts
```diff
@@ -17,3 +17,11 @@ export function isSwapError(response: unknown): response is SwapError {
     response !== null && typeof response === 'object' && 'error' in response
   );
 }
+
+export function formatTokenAmount(amount: string, decimals: number) {
+  // Convert the string amount to a number using decimals value
+  const numberAmount = Number(amount) / Math.pow(10, decimals);
+  // Round to a maximum of 11 significant digits
+  const roundedAmount = Number(numberAmount.toPrecision(11));
```
abcrane123, line 25: i round to 11 digits here so that the amount will fit in the input but let me know if there is another standard we should follow

Zizzamia: Let's merge this, but I would like the team feedback on what's best. In particular from Jordan Frankfurt and Conner.

Zizzamia (approved):

PR #560, opened 2024-06-14
Add example apps for local testing

ilikesymmetry:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


Zizzamia:
We are using Vocs for local testing, which is the `site` folder. We can close this one. 

If something is not quite working with Vocs, I would rather figure out how to improve `site`.

Also for general examples. This can be forked https://github.com/Zizzamia/an-onchain-app-in-100-components and you can spin up a new Repo to test on your own.

And one final thing, we want to be careful to keep this repo very heighly focus on the library, and code that needs to be maintained and kept super clean.

If we want to test Examples ides, BOAT repo is for that potentially, and we can talk how to restructured better that repo.

PR #562, opened 2024-06-14
chore: Update isWalletASmartWallet to isWalletACoinbaseSmartWallet

cpcramer:
**What changed? Why?**
Update `isWalletASmartWallet` naming to be `isWalletACoinbaseSmartWallet` so we explicitly state what the function is doing. 
**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/.changeset/ninety-windows-cheer.md
+++ b/.changeset/ninety-windows-cheer.md
@@ -1,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
+
+**chore**: Update isWalletASmartWallet name to be isWalletACoinbaseSmartWallet. We want to explicitly state that this is checking for Coinbase Smart Wallets. By @cpcramer #562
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -67,13 +67,20 @@
...
   "publishConfig": {
     "access": "public"
   },
-  "files": ["esm/**/*", "lib/**/*", "src/", "src/**/*"],
+  "files": [
+    "esm/**/*",
+    "lib/**/*",
+    "src/",
+    "src/**/*"
+  ],
```
cpcramer, line 83: Auto formatting from new CI lint

lukasrosario:
if we want this util would maybe go with `isAddressACoinbaseSmartWallet` instead, wallet is just the client, what we're really checking is the address

lukasrosario:
hmm actually this accepts a user op. so maybe something like `isCoinbaseSmartWalletUserOp` ?

ultimately i still think we should avoid utilities like these if possible and lean on capabilities. so i'd challenge this by asking if there is a specific use case for this that cant be answered by querying capabilities

Zizzamia (approved):
Approving and merging this for now. But yeah we need to re-think the capabilities approach, and how to make it simpler for folks to use them.

PR #563, opened 2024-06-14
feat: Add debounce to SwapAmountInput

kyhyco:
**What changed? Why?**
- Refactored `TextInput`
  - handles debounce
  - add callback to validate input
  - move to `src/internal/form`
- Update `SwapAmountInput` to use `TextInput`

**Notes to reviewers**

**How has it been tested?**
locally


src/internal/form/TextInput.tsx
```diff
@@ -0,0 +1,55 @@
...
+          onChange(value);
+        }
+      }
+    },
+    [onChange, handleDebounce, delayMs, setValue, inputValidator],
+  );
+
+  return (
+    <input
+      data-testid="ockTextInput_Search"
```
Zizzamia, line 47: We should probably update the `_Search` part.

Zizzamia: As it is used to insert also numbers for Swap Input.

Zizzamia (approved):

PR #577, opened 2024-06-14
fix: TokenSelectButton hit detection

kyhyco:
**What changed? Why?**
- fix TokenSelectButton not detecting click event when svg element is clicked

**Notes to reviewers**

**How has it been tested?**
locally

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -6,7 +6,8 @@
   "formatter": {
     "enabled": true,
     "indentWidth": 2,
-    "indentStyle": "space"
+    "indentStyle": "space",
+    "ignore": ["package.json"]
   },
   "javascript": {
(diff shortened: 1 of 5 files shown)


biome.json
```diff
@@ -6,7 +6,8 @@
   "formatter": {
     "enabled": true,
     "indentWidth": 2,
-    "indentStyle": "space"
+    "indentStyle": "space",
+    "ignore": ["package.json"]
```
kyhyco, line 10: We might be formatting during deploy. Ignoring `package.json` for now.

Zizzamia: Nice

Zizzamia (approved):

PR #604, opened 2024-06-17
feat: Add disable swap button functionality

abcrane123:
**What changed? Why?**
- add optional `disabled` prop to `SwapButton` so users can control whether or not swap is swappable

Screenshot 2024-06-17 at 1 18 12 PM

Screenshot 2024-06-17 at 1 21 48 PM

**Notes to reviewers**

**How has it been tested?**
locally

### DIFF
--- a/site/docs/pages/swap/swap.mdx
+++ b/site/docs/pages/swap/swap.mdx
@@ -72,6 +72,9 @@ export default function SwapComponents() {
     });
   }, [sendTransaction]);
 
+  // Disable SwapButton if additional checks required
+  const isDisabled = false;
+
   return ({ address ? (
     <Swap address={address}> // [!code focus]
       <SwapAmountInput // [!code focus]
@@ -85,7 +88,7 @@ export default function SwapComponents() {
         token={USDCToken} // [!code focus]
         type="to" // [!code focus]
       /> // [!code focus]
-      <SwapButton onSubmit={onSubmit} /> // [!code focus]
+      <SwapButton onSubmit={onSubmit} disabled={isDisabled} /> // [!code focus]
       <SwapMessage /> // [!code focus]
     </Swap> // [!code focus]
   ) : (
@@ -103,4 +106,4 @@ export default function SwapComponents() {
 - [`SwapReact`](/swap/types#swapreact)
 - [`SwapAmountInput`](/swap/types#swapamountinput)
 - [`SwapToggleButton`](/swap/types#swaptogglebutton)
-- [`SwapButton`](/swap/types#swapbutton)
+- [`SwapButtonReact`](/swap/types#swapbuttonreact)
--- a/src/swap/components/SwapButton.tsx
+++ b/src/swap/components/SwapButton.tsx
@@ -6,7 +6,7 @@ import { buildSwapTransaction } from '../core/buildSwapTransaction';
 import { cn } from '../../utils/cn';
 import { isSwapError } from '../core/isSwapError';
 
-export function SwapButton({ onSubmit }: SwapButtonReact) {
+export function SwapButton({ disabled, onSubmit }: SwapButtonReact) {
   const { address, fromAmount, fromToken, toToken, setError } =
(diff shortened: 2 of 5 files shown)


site/docs/pages/swap/swap.mdx
```diff
@@ -90,7 +93,7 @@ export default function SwapComponents() {
         token={USDCToken} // [!code focus]
         type="to" // [!code focus]
       /> // [!code focus]
-      <SwapButton onError={onError} onSubmit={onSubmit} /> // [!code focus]
+      <SwapButton onError={onError} onSubmit={onSubmit} disabled={isDisabled} /> // [!code focus]
```
kyhyco, line 96: We might want to handle the disable state internally and not expose it externally.

The `SwapButton` should only be enabled if 
- `fromAmount`, `fromToken`, `toToken` is set
- AND user has the available balance on their wallet

kyhyco (approved):

site/docs/pages/swap/swap.mdx
kyhyco: After discussion, we need to expose it externally to explicitly disable the button

0xAlec: shouldn't the developer have maximum control of the component?

e.g. `shadcn` is a library we want to base `OnchainKit` off, and they pass in `disabled` as a prop to their `Button` component ([example](https://ui.shadcn.com/docs/components/button#loading))

ilikesymmetry (approved):
I think we should enable `SwapAmountInput` to use a `disabled` flag too. Target use case is I'm a memecoin community who wants to make it easy for people to buy my token. I would fix the `"to"` swap token to my memecoin and disable it so that people only have to worry about choosing their input token and its amount.

site/docs/pages/swap/swap.mdx
ilikesymmetry: I +1 exposing control to disable how it is currently written! We can apply additional automatic disabled states for things like missing inputs, but enabling apps to have more control is our strategy seemingly!

kyhyco: All those points are valid point! We are carefully considering what should be exposed and what shouldn't. `disable` prop on the `SwapButton` is very reasonable and we should do it. :)

kyhyco:
@ilikesymmetry you can already do that :D

```typescript
<Swap>
  <SwapAmountInput token={{ eth info }} />
  <SwapAmountInput token={{ memecoin info }} />
</Swap>
```

this will lock-in the selected tokens

kyhyco (approved):

0xAlec (approved):

kyhyco:
@abcrane123 could you add `yarn changeset`?

kyhyco (approved):

site/docs/pages/swap/swap.mdx
```diff
@@ -103,4 +106,4 @@ export default function SwapComponents() {
 - [`SwapReact`](/swap/types#swapreact)
 - [`SwapAmountInput`](/swap/types#swapamountinput)
 - [`SwapToggleButton`](/swap/types#swaptogglebutton)
-- [`SwapButton`](/swap/types#swapbutton)
+- [`SwapButtonReact`](/swap/types#swapbuttonreact)
```
Zizzamia, line 109: oh thank you for fixing this

src/swap/components/SwapButton.tsx
```diff
@@ -6,7 +6,7 @@ import { buildSwapTransaction } from '../core/buildSwapTransaction';
 import { cn } from '../../utils/cn';
 import { isSwapError } from '../core/isSwapError';
 
-export function SwapButton({ onSubmit }: SwapButtonReact) {
+export function SwapButton({ disabled, onSubmit }: SwapButtonReact) {
```
Zizzamia, line 9: Should this have the default `false`? I understand it works as `undefined`, but doesn't seem correct.

PR #605, opened 2024-06-17
feat: docs for `buildSwapTransaction` + cleanup

0xAlec:
**What changed? Why?**
- fix hyperlinks in `getSwapQuote`
- add docs for `buildSwapTransaction`
- add docs for missing types

**Notes to reviewers**
image

image

**How has it been tested?**
locally

### DIFF
--- a/.changeset/pretty-camels-allow.md
+++ b/.changeset/pretty-camels-allow.md
@@ -1,0 +1,5 @@
+---
+"@coinbase/onchainkit": minor
+---
+
+feat: docs for `buildSwapTransaction` + cleanup by @0xAlec #605
--- a/site/docs/pages/swap/build-swap-transaction.mdx
+++ b/site/docs/pages/swap/build-swap-transaction.mdx
@@ -1,0 +1,117 @@
+# `buildSwapTransaction`
+
+The `buildSwapTransaction` function is used to get an unsigned transaction for a swap between two Tokens.
+
+## Usage
+
+:::code-group
+
+```tsx [code]
+import { buildSwapTransaction } from '@coinbase/onchainkit/swap';
+
+const fromToken: Token = {
+  name: 'ETH',
+  address: '',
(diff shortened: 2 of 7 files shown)


kyhyco (approved):

abcrane123 (approved):

.changeset/pretty-camels-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@coinbase/onchainkit": minor
+---
+
+feat: docs for `buildSwapTransaction` + cleanup by @0xAlec #605
```
kyhyco, line 5: @0xAlec nit:
```
- **feat**: ...
```

src/swap/types.ts
```diff
@@ -27,6 +27,9 @@ export type BuildSwapTransactionParams = GetSwapQuoteParams & {
   fromAddress: Address; // The address of the user
 };
 
+/**
+ * Note: exported as public Type
+ */
 export type Fee = {
   amount: string; // The amount of the fee
   baseAsset: Token; // The base asset for the fee
```
Zizzamia, line 35: @0xAlec, let's change this to `token` instead of `baseAsset`, as in OnchainKit we call `asset`, `token`.

site/docs/pages/swap/build-swap-transaction.mdx
```diff
@@ -0,0 +1,117 @@
+# `buildSwapTransaction`
+
+The `buildSwapTransaction` function is used to get an unsigned transaction for a swap between two Tokens.
+
```
Zizzamia, line 4: @0xAlec let's add more info here, like how to use `buildSwapTransaction` with `sendTransaction` from Wagmi. Also we should link `transaction` to `Transaction` the type.

So in our docs when we talk about entity we link the Types, so it makes sense on what are those.

Same for `Token`.

PR #611, opened 2024-06-17
feat: adjust `SwapMessage` component

abcrane123:
**What changed? Why?**
- add swap messages
- add swap error codes

Screenshot 2024-06-20 at 4 38 25 PM

**Notes to reviewers**

**How has it been tested?**
locally


site/docs/components/SwapComponents.tsx
```diff
@@ -109,7 +108,6 @@ export default function SwapComponents() {
               type="to"
             />
             <SwapButton onSubmit={onSubmit} />
-            <SwapMessage />
```
Zizzamia, line 112: Without exposing the component, than will be harder for the developer to customize the style.

src/swap/core/getSwapMessage.ts
```diff
@@ -0,0 +1,41 @@
+import type { GetSwapMessageParams } from '../types';
+
+export enum SwapMessage {
+  INCOMPLETE_FIELD = 'Complete the fields to continue',
+  INSUFFICIENT_BALANCE = 'Insufficient balance',
+  LOW_LIQUIDITY = 'Liquidity too low for the token"',
+}
```
Zizzamia, line 7: Great use of ENUM.

site/docs/components/SwapComponents.tsx
abcrane123: yeah that is a good point, what do you think @kyhyco ?

kyhyco: sgtm

src/swap/types.ts
```diff
@@ -162,7 +183,7 @@ export type SwapReact = {
  * Note: exported as public Type
  */
 export type SwapError = {
-  code: number; // The error code
+  code?: number; // The error code
```
Zizzamia, line 186: Can we make `code` not optional? I like the idea that with every error we have a code we can track? what do you think?

src/swap/types.ts
```diff
@@ -116,18 +127,28 @@ export type SwapContextType = {
...
+  setIsLoading: (l: SwapLoadingState) => void;
   setToAmount: (a: string) => void;
   setToToken: (t: Token) => void;
   toAmount: string;
   toToken?: Token;
 };
 
+export type SwapLoadingState = {
+  fromTokenBalance?: boolean;
+  swapQuote?: boolean;
```
Zizzamia, line 148: When working with booleans, it's good practice to either start the variable with `is` or `has`.

So when reading the code it's easier to know the potential type without always triple checking.

src/swap/core/getSwapMessage.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function getSwapMessage({
+  error,
+  fromAmount,
+  fromToken,
+  fromTokenBalance,
+  isLoading,
+  toAmount,
+  toToken,
+}: GetSwapMessageParams) {
```
Zizzamia, line 20: Very clean, I like this direction.

Zizzamia: Also it's going to make it super easy to test, and mantain.

src/swap/components/SwapAmountInput.tsx
```diff
@@ -121,6 +127,60 @@ export function SwapAmountInput({
...
+    setIsLoading,
+    token,
+    type,
+  ]);
+
+  useEffect(() => {
+    if (
+      token?.symbol !== 'ETH' &&
+      balanceResponse?.error &&
+      error?.error !== balanceResponse?.error?.shortMessage
```
Zizzamia, line 154: Just a thought, but this seems could be a small utility that can be shared between the two `useEffect`
```
token?.symbol !== 'ETH' &&
balanceResponse?.error &&
error?.error !== balanceResponse?.error?.shortMessage
```

src/swap/components/SwapAmountInput.tsx
```diff
@@ -121,6 +127,60 @@ export function SwapAmountInput({
     }
   }, [token, setToken]);
 
+  useEffect(() => {
```
Zizzamia, line 130: Just for readibility for the code let's add a line of comment to each `useEffect`, so when reading this later we don't forget what's are all about.

src/swap/types.ts
abcrane123: yeah i was wondering how to handle this best. we discussed changing the code to a string `SWAP_ERROR_<CODE>` which might be a good idea. the errors returned from the balance endpoints do not have a number code.

Zizzamia: My take here is that standardize API will take weeks, where can standardize the frontend in days.

So let's lead the way on how we think about code error strings, and than work backwards with the team to have the API reflect those as well. cc @0xAlec

src/swap/components/SwapAmountInput.tsx
abcrane123: this logic was moved to `Swap`

src/swap/components/SwapAmountInput.tsx
abcrane123: this logic was removed

src/swap/components/SwapAmountInput.tsx
```diff
@@ -125,14 +127,17 @@ export function SwapAmountInput({
       </div>
       <div className="flex w-full items-center justify-between">
         <TextInput
-          className="w-full border-[none] bg-transparent font-display text-[2.5rem] text-gray-900 leading-none outline-none"
+          className={cn(
+            'w-full border-[none] bg-transparent font-display text-[2.5rem]',
+            'leading-none outline-none',
+            hasInsufficientBalance ? 'text-[red]' : 'text-gray-900',
```
kyhyco, line 133: Use the new `color` theme object here.

`color` options are:
```
export const color = {
  inverse: 'text-inverse',
  foreground: 'text-foreground',
  foregroundMuted: 'text-foreground-muted',
  error: 'text-error',
  primary: 'text-primary',
  success: 'text-success',
  warning: 'text-warning',
  disabled: 'text-disabled',
} as const;
```

```suggestion
            hasInsufficientBalance ? color.error : color.foreground,
```

src/swap/components/Swap.tsx
```diff
@@ -9,15 +9,21 @@ import { isSwapError } from '../core/isSwapError';
...
+  SwapLoadingState,
+  SwapReact,
+} from '../types';
 import type { Token } from '../../token';
 import type { UseBalanceReturnType, UseReadContractReturnType } from 'wagmi';
 import { background, cn, text } from '../../styles/theme';
 
 export function Swap({ address, children, title = 'Swap' }: SwapReact) {
-  const [error, setError] = useState<SwapError>();
+  const [swapErrorState, setSwapErrorState] = useState<SwapErrorState>();
```
Zizzamia, line 26: love the care for types

src/swap/constants.ts
```diff
@@ -0,0 +1,6 @@
+export const LOW_LIQUIDITY_ERROR_CODE = 'SWAP_QUOTE_LOW_LIQUIDITY_ERROR';
+export const GENERAL_SWAP_ERROR_CODE = 'SWAP_ERROR';
+export const GENERAL_SWAP_QUOTE_ERROR_CODE = 'SWAP_QUOTE_ERROR';
+export const GENERAL_SWAP_BALANCE_ERROR_CODE = 'SWAP_BALANCE_ERROR';
+export const UNCAUGHT_SWAP_QUOTE_ERROR_CODE = 'UNCAUGHT_SWAP_QUOTE_ERROR';
+export const UNCAUGHT_SWAP_ERROR_CODE = 'UNCAUGHT_SWAP_ERROR';
```
Zizzamia, line 6: Should ERROR_CODE being a post-fix or a pre-fix.
mmmmmm

I know in English it's more correct as post-fix, but in reading code might be not as easy. Let's leave it like this for now. But I have a feeling all const error, might be better as error code pre-fix. 

OR

we can create a mapping system like we are doing for themeing. And just create a `const errorCode = {}` which can be autocomplete when coding.

src/swap/core/getSwapErrorCode.ts
```diff
@@ -0,0 +1,37 @@
+import {
+  GENERAL_SWAP_ERROR_CODE,
+  GENERAL_SWAP_BALANCE_ERROR_CODE,
+  GENERAL_SWAP_QUOTE_ERROR_CODE,
+  LOW_LIQUIDITY_ERROR_CODE,
+  UNCAUGHT_SWAP_ERROR_CODE,
+  UNCAUGHT_SWAP_QUOTE_ERROR_CODE,
+} from '../constants';
+
+export function getSwapErrorCode(
```
Zizzamia, line 10: Next PR, we should find the right place to document those error in Swap.

Maybe as we have the `types` page, we start having the `errors` page for each Module. Which will help inform the type of error components can have for each module.

src/swap/core/getSwapMessage.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export enum SwapMessage {
+  BALANCE_ERROR = 'Error fetching token balance',
+  INCOMPLETE_FIELD = 'Complete the fields to continue',
+  INSUFFICIENT_BALANCE = 'Insufficient balance',
+  IS_LOADING = 'Loading...',
+  LOW_LIQUIDITY = 'Liquidity too low for the token',
+}
+
+export function getSwapMessage({
```
Zizzamia, line 12: Oh we are missing the `test` file. Something for next PR.

Zizzamia (approved):

PR #612, opened 2024-06-17
chore: fix minor issues

abcrane123:
**What changed? Why?**
- add default value for `disabled` prop
- handle 0n value for token balance
- handle edge case (display `0` instead of `0.` for rounded amount)
**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/utils/getRoundedAmount.ts
+++ b/src/utils/getRoundedAmount.ts
@@ -1,10 +1,7 @@
 export function getRoundedAmount(balance: string, fractionDigits: number) {
-  const parsedBalance = Number.parseFloat(balance);
-  const roundedBalance = Number(parsedBalance)
-    ?.toFixed(fractionDigits)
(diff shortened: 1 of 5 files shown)


src/utils/getRoundedAmount.ts
```diff
@@ -1,4 +1,10 @@
 export function getRoundedAmount(balance: string, fractionDigits: number) {
   const parsedBalance = Number.parseFloat(balance);
-  return Number(parsedBalance)?.toFixed(fractionDigits).replace(/0+$/, '');
+  const roundedBalance = Number(parsedBalance)
+    ?.toFixed(fractionDigits)
+    .replace(/0+$/, '');
+  if (roundedBalance === '0.') {
```
abcrane123, line 6: wondering if there is a cleaner way to handle this

Zizzamia (approved):

PR #613, opened 2024-06-17
docs: force doc site to connect to base chain

kyhyco:
**What changed? Why?**
- force doc site to connect to base chain

**Notes to reviewers**

**How has it been tested?**
locally

### DIFF
--- a/site/docs/components/App.tsx
+++ b/site/docs/components/App.tsx
@@ -18,6 +18,9 @@ const wagmiConfig = createConfig({
     coinbaseWallet({
       appChainIds: [base.id],
       appName: 'onchainkit',
+      options: {
+        chainId: base.id,
+      },
     }),
   ],
   ssr: true,
--- a/.changeset/happy-avocados-matter.md
+++ b/.changeset/happy-avocados-matter.md
@@ -1,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
+
+- **docs**: force doc site to connect to base. By @kyhyco #613


0xAlec (approved):

site/docs/components/App.tsx
```diff
@@ -18,6 +18,9 @@ const wagmiConfig = createConfig({
     coinbaseWallet({
       appChainIds: [base.id],
       appName: 'onchainkit',
+      options: {
+        chainId: base.id,
+      },
```
kyhyco, line 23: Should we be doing this using `useConnect` and handles this in `OnchainKitProvider`?

@ilikesymmetry

ilikesymmetry: To clarify, the goal is for users to have base selected automatically as their chain for the Swap demo?

One thing to consider is that we generally should demo on testnets, so `baseSepolia`, but maybe that doesn't make sense here because people won't know what tokens to search for. Does our API even support testnet if we wanted?

I'm not familiar with the cbwallet-specific options, but if that works we should just roll with it. ConnectKit handles connected-chain-validation for devs which is really nice, so it may be worth peeping their solution. I think pairing [`useAccount`](https://wagmi.sh/vue/api/composables/useAccount#usage) to get the account's connected chain and [`useSwitchChain`](https://wagmi.sh/react/api/hooks/useSwitchChain#useswitchchain) to prompt chain switching will be our best bet for custom-rolling in the general case.

PR #615, opened 2024-06-17
feat: swap title

kyhyco:
**What changed? Why?**
- allow users to set the title for Swap component

**Notes to reviewers**

**How has it been tested?**
locally


Zizzamia:
@kyhyco let's take a minute to add a test for this case.

src/swap/components/Swap.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    useBalance: jest.fn(),
+    useReadContract: jest.fn(),
+  };
+});
+
+jest.mock('../../internal/text', () => {
+  return {
+    TextTitle3: ({ children }: { children: string }) => <div>{children}</div>,
+  };
+});
```
kyhyco, line 20: I will fix this in a follow up PR. Getting a bit messy with resolving external dependencies like `clsx`

src/swap/components/Swap.tsx
```diff
@@ -76,6 +78,7 @@ export function Swap({ address, children }: SwapReact) {
     [fromToken, toToken],
   );
 
+  /* istanbul ignore next */
```
kyhyco, line 81: I will add tests for these in a follow up PR

Zizzamia (approved):
Niiiice

PR #616, opened 2024-06-18
chore: upgrade packemon to v4

kyhyco:
**What changed? Why?**
- upgrade node to v20
- upgrade packemon to v4

**Notes to reviewers**
packemon v4 supports ESM only

**How has it been tested?**
locally


.nvmrc
```diff
@@ -1 +1 @@
-18.17
+20.11
```
kyhyco, line 1: Do we need to update the GHA?

Zizzamia:
I am very cold feet on upgrading `packemon to v4`. Last time we tried and changed the all `lib` story to `esm` it broke a ton of stuff on Vercel.

I am going to pause this PR, and try to merge it and test it in July when things are a bit more chill. Right now we can not effort to spend a day or even an hour more on Node / Build issues.

Zizzamia (approved):

PR #617, opened 2024-06-18
feat: :ock typescript import

kyhyco:
**What changed? Why?**
Followed advice from Miles on how to augment packemon. 

**Notes to reviewers**
Verified build artifact for `TokenRow` component against old packemon settings. The build artifact is the same for both old and new packemon approach.

**How has it been tested?**
locally


package.json
```diff
@@ -93,55 +94,55 @@
       "types": "./esm/index.d.ts",
       "module": "./esm/index.js",
       "import": "./esm/index.js",
-      "default": "./lib/index.js"
+      "default": "./esm/index.js"
```
Zizzamia, line 97: I don't feel confortable about this change. We will pause this till July.

Left more context here
https://github.com/coinbase/onchainkit/pull/616

Zizzamia:
Closing this one for now.

PR #624, opened 2024-06-18
fix: Identity matching Name component

Zizzamia:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/identity/components/Identity.tsx
```diff
@@ -17,7 +17,7 @@ export function Identity({ address, children }: IdentityReact) {
       // @ts-ignore
       avatar: childrenArray.filter(({ type }) => type === Avatar),
       // @ts-ignore
-      names: childrenArray.filter(({ type }) => type !== Name),
+      names: childrenArray.filter(({ type }) => type === Name),
```
Zizzamia, line 20: Added more tests to avoid this typo in the future.

PR #630, opened 2024-06-18
feat: Add swap loading state

abcrane123:
**What changed? Why?**
- refactor token balance api calls
- add swap loading state (input is disabled if quote is loading)

**Notes to reviewers**

**How has it been tested?**
locally


src/swap/components/Swap.tsx
```diff
@@ -1,4 +1,4 @@
-import { Children, useCallback, useMemo, useState } from 'react';
+import { Children, useCallback, useEffect, useMemo, useState } from 'react';
```
kyhyco, line 1: nit:
```suggestion
import { Children, useCallback, useMemo, useState } from 'react';
```

kyhyco (approved):

src/swap/components/Swap.tsx
```diff
@@ -8,15 +8,89 @@ import { getSwapQuote } from '../core/getSwapQuote';
...
+      isToQuoteLoading: false,
+    });
+
+  // returns ETH balance
+  const ethBalanceResponse: UseBalanceReturnType = useBalance({
+    address,
+  });
+
+  // returns erc20 token balance
+  const fromTokenBalanceResponse: UseReadContractReturnType = useReadContract({
```
Zizzamia, line 37: We should probably abstract this into his own Hook. As we are going to reuse it soon.

abcrane123: agreed

PR #632, opened 2024-06-18
feat: add theming to all Text components

kyhyco:
**What changed? Why?**
- add css variables
- extend tailwind config to generate new text CSS classes
- add `TextAs` component which is the text primitive component
- apply new text component through out the code base
- tailwindcss integration guide

**Notes to reviewers**

Follow up PRs:
- Identity components + className override to text components
- background color theme

**How has it been tested?**
locally


src/internal/text/TextAs.tsx
```diff
@@ -0,0 +1,25 @@
...
+} & TextReact;
+
+export function TextAs({
+  as,
+  children,
+  className,
+  variant = 'foreground',
+  ...props
+}: TextAsReact) {
+  const textColor = textColorMap[variant];
```
ilikesymmetry, line 18: Kind of confused why we use the variant to switch `textColor` versus just passing the color through the `className`? I can see why we'd want to make an association between high-level emphasis names and color styling, but I feel like we should start with the simplest approach first and then refactor to find more optimal abstractions. Jumping to an abstraction beyond classNames introduces more vocab all devs need to keep in context.

I've never built a component library, but I urge us to second check any abstractions that do not have an absolute necessity. Could you explain to me where effort this saved by internal and external devs by using this variant system in comparison to just taking colors via className?

src/internal/text/TextBody.tsx
```diff
@@ -1,13 +1,12 @@
...
-};
-
-export function TextBody({ children }: TextBodyReact) {
+export function TextBody(props: TextReact) {
   return (
-    <span className="text-base text-black text-sans leading-normal">
-      {children}
-    </span>
+    <TextAs
+      as="span"
```
ilikesymmetry, line 7: For example, this I consider a step backwards in internal DevX. Every frontend developer immediately knows what a `span` is, but it takes me clicking in to the `TextAs` component, reading it, and zooming out to understand this custom syntax. It also takes more keystrokes to express the same thing ("this is a span").

src/internal/text/textColorMap.ts
ilikesymmetry, line 1: This has code smell to me because the mapping isn't expressing any new information. It's a mapping from a string `key` to `text-${key}`. Can I get more context on the use-case for this?

src/swap/components/SwapMessage.tsx
```diff
@@ -16,7 +16,7 @@ export function SwapMessage() {
   }, [error]);
   return (
     <div className="flex">
-      <TextMessage>{message}</TextMessage>
+      <TextLabel2 data-testid="ockSwapMessage_Message">{message}</TextLabel2>
```
ilikesymmetry, line 19: What is a `TextLabel2` and how is it different from a `TextLabel`? I would rather express variants of similar ideas by name, not arbitrary number.

src/token/components/TokenRow.tsx
```diff
@@ -23,9 +23,7 @@ export const TokenRow = memo(function TokenRow({
         <span className="flex flex-col items-start">
           <TextHeadline>{token.name}</TextHeadline>
           {!hideSymbol && (
-            <span className="font-normal text-[#5B616E] leading-normal">
-              {token.symbol}
-            </span>
+            <TextBody variant="foreground-muted">{token.symbol}</TextBody>
```
ilikesymmetry, line 26: I do not consider this an improvement in DevX. I can read and immediately know what the former implementation looks like, but the new version requires me to look into `TextBody` and our variant mapping to figure out what this looks like.

ilikesymmetry (commented):
Have a handful of questions on the context behind these tradeoffs!

src/internal/text/TextAs.tsx
kyhyco: `Text` components will never be exposed directly to the end users. These are internal components.

We want to display all possible values for color in a given `Text` component. Right now, you will get an autocompletion dialog when modify any `Text` component.

Sure, we could just have `className` but then you will need to remember all the possible class names specifically associated with text colors and that's a pain.

src/internal/text/textColorMap.ts
kyhyco: explained above

src/token/components/TokenRow.tsx
kyhyco: The point is that you do not have to read what `TextBody` is.

kyhyco: That is to say, we can rename `variant` to `color` instead to make it clear what it affects.

src/internal/text/TextAs.tsx
kyhyco: Btw, the `className` will be added in the follow up when I fix up the Identity components.

src/internal/text/TextBody.tsx
Zizzamia: @ilikesymmetry I feel the same, but let me reflect very quickly on what's going on here.

I think here the goal is to optimize the internal use of TextBody, not to expose `TextAs`.

Internal components are a bit more about effeciency, and in some cases sacrificies usability. Which means do more with less.

The real question here to ask is, are we ever going to use `as="span"` as feature? like have `h3` or `h4` in it? So far we don't. But this could be easily be work that helps us get there if need it.

So if we are thinking to use `as` feature relativelly soon, that it makes sense keep it. If not, than we should just use `span`. 

@kyhyco what's your answear here.

src/internal/text/index.ts
```diff
@@ -2,5 +2,4 @@ export { TextBody } from './TextBody';
 export { TextHeadline } from './TextHeadline';
 export { TextLabel1 } from './TextLabel1';
 export { TextLabel2 } from './TextLabel2';
-export { TextMessage } from './TextMessage';
```
Zizzamia, line 5: hehehe yeah I agree, I wasn't quite sure to have it. Was more a quick decion.

Thank you for correcting this.

src/internal/text/textColorMap.ts
```diff
@@ -0,0 +1,10 @@
+export const textColorMap = {
```
Zizzamia, line 1: nit. let's to UPPER_CASE for const.

so it will be `TEXT_COLOR_MAP`

src/swap/components/SwapMessage.tsx
Zizzamia: +1 on @kyhyco point

Zizzamia: @ilikesymmetry, overall what we are optimizing here, it's time to build a component by having clear aligment on how internal Figma primitives are named and than used inside the internal code.

In this case, we want those name to feel more Figma vibes, because they come from the last few years of experience from the internal UI System team in how they approach large complex design systems.

So, something you will see, it's how we expose public components that feel right for our audience, but we build internal components that are more optimized for us as Eng and Design team to work well and fast.

Zizzamia (approved):

Zizzamia:
@kyhyco there a  couple of notes that we can follow up later, but for now I am going to merge this PR as it is a great step forward for us.

src/internal/text/TextBody.tsx
kyhyco: Here are the alternative:
```javascript
export function TextBody(props: TextReact) {
  const color = useTextColor(color);
  const className = cn('font-sans text-base leading-normal', color);

   // class assignment behavior from next PR
   const color = useTextColor(color);
   const className= cn('font-sans text-base leading-normal', color, props.className);

    // no text color assignment
    const className=cn('font-sans text-base leading-normal', props.className);

  return (
    <span className={className} {...props} />
  )
}
```

The primary reason for `TextAs` is to consolidate the color assignment logic across multiple text components without having to copy paste it every time.`TextAs` is a super simple component as well. 

Here are the benefits of this whole approach:
- For potentially addressing accessibility, it would be nice to address that in a single line PR by exposing `as` on `TextReact` type.

- For DX, when I look at a UX design, I look for the text name + color. I would love for the autocompletion to kick in for selecting the component and color.

- For maintainability, you only need to modify `TextAs` and `TextReact` without copy pasting new business logic. We also guarantee that share logic will be in `TextAs`.

- For adding new Text component, I think this is neat:
  - copy paste any regular text component
  - change name
  - set default tag
  - update tailwind styles

Yes, I agree `span` is much nicer to read on initial reading. But you don't get the benefits of centralized logic in a single file. That is also nice.

src/internal/text/TextAs.tsx
ilikesymmetry: My main point is that instead of remembering a className, I have to remember a variant name so to me it's no saved effort and therefore an abstraction that doesn't need to exist. I'm all for easing required context for developers, but I feel like this doesn't achieve that.

src/token/components/TokenRow.tsx
ilikesymmetry: I still don't think that autocomplete of a color prop is worth adding the abstraction versus just doing the raw className. It feels like more juice than the squeeze and we have to maintain this moving forward. I'll lean on your judgement, but my opinion is to be ruthless with adding new abstractions.

kyhyco: Oh you better know I hate meaningless abstractions haha

kyhyco: I would say that once you have component name + color autocompletion _and_ you have minimal abstraction between understanding designs and converting them to code, you don't want to go back. I promise you.

src/internal/text/TextAs.tsx
Zizzamia: `variant name` are a common practice of Design Systems. As when building large project where you need to collaborate with many designers and engs, it's easier to align on `variant name` that atomic classnames.

A `variant name` can last years, a `classname` can last only a few months, till a new design decision will happen.

src/internal/text/TextBody.tsx
Zizzamia: Great context, thank you @kyhyco 
Ok so based on that it makes sense to keep this implementation.

src/token/components/TokenRow.tsx
Zizzamia: My summary here is that abstraction are not need it as much in this case when the team is small, and also the codebase is small.

But as soon as you get 3+ eng and 2+ designers, and 20+ pages with 100+ components. Any meaninful re-design or design standardization, it's faster and easier when abstraction like this one are in place.

Coinbase had not this kind of abstraction in 2019 and back than was a pain align on desing or any meaniful refactoring. When we adopted a more mature Design System in 2020, that allowed us to re-shape 100s of pages in just weeks and also keep high level of consistency between all those pages.

So again, a team with just 3~4 people, with a small codebase, they probably don't need this. Anything larger, they will appreciate this, as it will increase quality and speed overall.

src/swap/components/SwapMessage.tsx
kyhyco: To drive it home here, there are meanings associated with new color names (i.e. `foreground`, `forgroundMuted`). And that's what we will see for the properties in figma. The color value is _part_ of the new color names. This helps designers and engineers communicate in a higher level.

I can say "can we change the text color here from foreground to forgroundMuted"? Instead of requesting for specific color values like "red-900" and "blue-500".

Additionally, what are values can we use for text color? Tailwind comes with a lot of colors and we only use a handful. Letting you know what they are is super nice.

PR #633, opened 2024-06-18
feat: Add swap loading state and spinner to SwapButton

abcrane123:
**What changed? Why?**
- add swap loading state to SwapButton with spinner
- add dependencies to context

Screenshot 2024-06-18 at 4 51 15 PM

**Notes to reviewers**

**How has it been tested?**
locally


src/swap/components/Swap.tsx
```diff
@@ -213,6 +213,11 @@ export function Swap({ address, children, title = 'Swap' }: SwapReact) {
     roundedFromTokenBalance,
     roundedToTokenBalance,
     swapQuoteLoadingState,
+    setFromAmount,
+    setFromToken,
+    setSwapQuoteLoadingState,
+    setToAmount,
+    setToToken,
```
abcrane123, line 220: if we don't add these the `fromAmount` and `toAmount` changes aren't reflected in `SwapButton`

abcrane123: but it is causing lint to fail

src/swap/components/SwapButton.tsx
```diff
@@ -30,18 +38,35 @@ export function SwapButton({ disabled = false, onSubmit }: SwapButtonReact) {
...
+    !fromToken ||
+    !toAmount ||
+    !toToken ||
+    disabled ||
+    swapQuoteLoadingState?.isFromQuoteLoading ||
+    swapQuoteLoadingState?.isToQuoteLoading;
+
+  const isLoading =
+    swapQuoteLoadingState?.isFromQuoteLoading ||
+    swapQuoteLoadingState?.isToQuoteLoading;
```
kyhyco, line 52: we want loading for `onSubmit` and not on `getSwapQuote`

abcrane123: oh woops lol

abcrane123: fixed !

src/swap/types.ts
```diff
@@ -134,14 +134,16 @@ export type SwapContextType = {
   setFromToken: (t: Token) => void;
   setToAmount: (a: string) => void;
   setToToken: (t: Token) => void;
-  swapQuoteLoadingState: SwapQuoteLoadingState;
+  setSwapLoadingState: (s: SwapLoadingState) => void;
+  swapLoadingState: SwapLoadingState;
```
abcrane123, line 138: wondering if these should be optional

src/swap/components/Swap.tsx
Zizzamia: For those edge cases, just add 
```
/* biome-ignore lint: [add reason] */
```

Zizzamia: 5% of the time Linting rules are not efficient, and for those cases we ignore them and document the reason

src/swap/types.ts
Zizzamia: what's your thinking?

abcrane123: hmm i'm torn because on one hand requiring it ensures loading states are handled correctly but on the other hand making it optional allows more flexibility for users if they want to exclude

src/swap/components/Swap.tsx
```diff
@@ -100,9 +100,9 @@ export function Swap({ address, children, title = 'Swap' }: SwapReact) {
         return;
       }
       try {
```
Zizzamia, line 102: @abcrane123 follow up PR. This is getting harder to read, let's add comments to explain for each loading state what's going on.

src/swap/components/Swap.tsx
```diff
@@ -139,8 +139,8 @@ export function Swap({ address, children, title = 'Swap' }: SwapReact) {
...
```
Zizzamia, line 141: @abcrane123 follow up PR. Same here, add line comments to explain what's going on.

src/swap/types.ts
Zizzamia: we can make optional later if folks ask for it.

Zizzamia (approved):

PR #642, opened 2024-06-20
feat: add new background, pressable and color themes

kyhyco:
**What changed? Why?**
- add new background, pressable and color themes

**Notes to reviewers**
Address new style guide in a later PR

**How has it been tested?**
locally


src/styles/theme.ts
```diff
@@ -12,7 +12,42 @@ export const text = {
...
+export const color = {
+  inverse: 'text-inverse',
+  foreground: 'text-foreground',
+  foregroundMuted: 'text-foreground-muted',
+  error: 'text-error',
+  primary: 'text-primary',
+  success: 'text-success',
+  warning: 'text-warning',
+  disabled: 'text-disabled',
+} as const;
```
kyhyco, line 53: These might seem unnecessary. But this really improves the readability of styles where a className maybe a very long string.

For now, I don't foresee additional style mappers.

package.json
```diff
@@ -88,6 +88,7 @@
     "./package.json": "./package.json",
     "./styles.css": "./src/styles.css",
     "./tailwind.css": "./src/tailwind.css",
+    "./theme": "./esm/styles/theme.js",
```
kyhyco, line 91: this is gonna be super cool where we expose these themes for users to do basic style editing.

src/styles/theme.ts
Zizzamia: awesome

PR #673, opened 2024-06-21
chore: refactor balance hooks

abcrane123:
**What changed? Why?**
- refactor balance api calls into reusable hooks

**Notes to reviewers**

**How has it been tested?**
locally

### DIFF
--- a/src/swap/components/Swap.tsx
+++ b/src/swap/components/Swap.tsx
@@ -7,11 +7,10 @@ import { SwapContext } from '../context';
 import { getSwapQuote } from '../core/getSwapQuote';
 import { isSwapError } from '../core/isSwapError';
 import { formatTokenAmount } from '../../utils/formatTokenAmount';
-import { useBalance, useReadContract } from 'wagmi';
-import { getTokenBalances } from '../core/getTokenBalances';
 import { getTokenBalanceErrorState } from '../core/getTokenBalanceErrorState';
-import { erc20Abi } from 'viem';
-import type { Address } from 'viem';
+import { background, cn, text } from '../../styles/theme';
+import { useGetETHBalance } from '../core/useGetETHBalance';
+import { useGetTokenBalance } from '../core/useGetTokenBalance';
 import type {
   SwapError,
   SwapErrorState,
@@ -19,8 +18,6 @@ import type {
   SwapReact,
 } from '../types';
 import type { Token } from '../../token';
-import type { UseBalanceReturnType, UseReadContractReturnType } from 'wagmi';
-import { background, cn, text } from '../../styles/theme';
 
 export function Swap({ address, children, title = 'Swap' }: SwapReact) {
   const [swapErrorState, setSwapErrorState] = useState<SwapErrorState>();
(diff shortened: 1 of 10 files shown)


src/swap/core/useGetETHBalance.ts
```diff
@@ -0,0 +1,26 @@
+import { useMemo } from 'react';
+import { useBalance } from 'wagmi';
+import { formatUnits } from 'viem';
+import { getRoundedAmount } from '../../utils/getRoundedAmount';
+import type { Address } from 'viem';
+
+const ETH_DECIMALS = 18;
+export function useGetETHBalance(address: Address) {
```
Zizzamia, line 8: I think this hook should live in the `wallet` module. My thinking here is that we are checking the Balance of a Wallet.

And Swap uses this utility. Similar how Swap uses Token utilities.

src/swap/core/useGetTokenBalance.ts
```diff
@@ -0,0 +1,36 @@
+import { useMemo } from 'react';
+import { useReadContract } from 'wagmi';
+import { erc20Abi, formatUnits } from 'viem';
+import { getRoundedAmount } from '../../utils/getRoundedAmount';
+import type { Address } from 'viem';
+import type { Token } from '../../token';
+import type { UseReadContractReturnType } from 'wagmi';
+
+export function useGetTokenBalance(address: Address, token?: Token) {
```
Zizzamia, line 9: Same, let's have this in Wallet.

I see Wallet as accessing all info related to a particular Account.

Zizzamia: Btw I really like the simple API you wrote as `useGetTokenBalance(address: Address, token?: Token)`, very nice.

Btw why `token` is optional?

abcrane123: `token` is optional because `token` and `swappableTokens` are optional props for `SwapAmountInput` so there is a chance that `token` is undefined in `Swap` component ( typescript gets upset )

src/swap/core/useGetETHBalance.ts
abcrane123: done !

src/swap/core/useGetTokenBalance.ts
Zizzamia: Ok so, small question. Does `useGetTokenBalance` works without passing a Token?

If not, we should make it mandatory, if does, than optional is fine.

Ideally the Hooks type should reflect how the Hook works, and than if in case in Swap we don't have a Token, than we use conditinals to cover that case in the Swap component.

abcrane123: it does not work without a `token` (it will be disabled). since it is a hook, i don't think we are able to conditionally call it. do you have another work around in mind ?

src/swap/components/Swap.tsx
```diff
@@ -34,91 +30,65 @@ export function Swap({ address, children, title = 'Swap' }: SwapReact) {
...
-    toToken,
-    toTokenBalanceResponse,
-  ]);
+  const {
+    convertedBalance: convertedToBalance,
+    roundedBalance: roundedToBalance,
+    error: toBalanceError,
+  } = useGetTokenBalance(address, toToken);
 
+  /* istanbul ignore next */
```
abcrane123, line 51: the coverage report was getting upset about these lines so i added the ignore statement here. i'm not sure of a way to test those values since they don't render anywhere but let me know if someone has an idea of how to test these lines. 
Screenshot 2024-06-21 at 1 11 34 PM

kyhyco (approved):

PR #676, opened 2024-06-21
chore: Add tests and comments

abcrane123:
**What changed? Why?**
- add test for `Spinner` component
- add comments to amount change handlers

**Notes to reviewers**

**How has it been tested?**
locally

### DIFF
--- a/src/swap/components/Swap.tsx
+++ b/src/swap/components/Swap.tsx
@@ -101,6 +101,8 @@ export function Swap({ address, children, title = 'Swap' }: SwapReact) {
         return;
       }
       try {
+        /* when fromAmount changes we fetch quote for toAmount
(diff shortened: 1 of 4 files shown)


kyhyco (approved):

src/swap/components/Swap.tsx
```diff
@@ -140,6 +145,8 @@ export function Swap({ address, children, title = 'Swap' }: SwapReact) {
         return;
       }
       try {
+        /* when toAmount changes we fetch quote for fromAmount
```
Zizzamia, line 148: Small nit for future comments.

Use this kind of comments
```
//
```
inside function

and this kind
```
/**
 *
 */
```
at the begining of a function.

Issue #684, opened 2024-06-22
Bug: ConnectAccount (esm)


tuler:
### Describe the bug and the steps to reproduce it

```
ModuleBuildError: Module build failed (from ../../node_modules/.pnpm/next@14.2.4_@babel+core@7.24.7_react-dom@18.3.1_react@18.3.1__react@18.3.1/node_modules/next/dist/build/webpack/loaders/next-swc-loader.js):
Error: 
  × The "use client" directive must be placed before other expressions. Move it to the top of the file to resolve this issue.
   ╭─[/<my_project>/node_modules/.pnpm/@coinbase+onchainkit@0.22.1_@tanstack+query-core@5.45.0_@types+react@18.3.3_@xmtp+frames-vali_z233vvjdr7prezdn7muxchujsq/node_modules/@coinbase/onchainkit/esm/wallet/components/ConnectAccount.js:1:1]
 1 │ import { useCallback } from 'react';
 2 │ import { useAccount, useConnect, useDisconnect } from 'wagmi';
 3 │ import { jsxs, jsx } from 'react/jsx-runtime';
 4 │ 'use client';
   · ─────────────
 5 │ /**
 6 │  * ConnectAccount
 7 │  *  - Connects to the wallet
   ╰────
```

### What's the expected behavior?

Connect button to render

### What version of the libraries are you using?

0.22.1

PirosB3:
Having same exact issue

kyhyco:
How do you reproduce this error?

kyhyco:
`ConnectAccount` needs to be used in a client component. Looks like you are trying to render the component on the server.

narteysarso:
@kyhyco  I have the same issue with  version: 0.22.1. 

I am using it in an client side component. 

This is my code:
```
"use client";

import {Identity, Avatar, Name, Badge, Address} from "@coinbase/onchainkit/identity";

const Profile = () => {
    return(
        <Identity
        address="0x838aD0EAE54F99F1926dA7C3b6bFbF617389B4D9"
        schemaId="0xf8b05c79f090979bf4a80270aba232dff11a10d9ca55c4f88de95317970f0de9"
      >
        <Avatar />
        <Name>
          <Badge />
        </Name>
        <Address />
      </Identity>
    )
}

export default Profile;

```

My opinion is that version 0.22.1 has `'use client'` directive added to the code. However, the directive is wrongly placed in the code. 'use client' directive should be the first line in the code.

[...]

PR #691, opened 2024-06-24
feat: migrate from packemon to tsup

roushou:
This PR uses [tsup](https://github.com/egoist/tsup) for bundling instead of packemon. This should fix the bundling issues I came across recently.

- server components: https://github.com/coinbase/onchainkit/issues/684
- esm: https://github.com/coinbase/onchainkit/pull/646

TBH I have no knowledge of packemon so there might be easier solutions, I'm just introducing a tool that does the job fine for me, and maybe it will here too. An objective point though is that `tsup` is widely used, should be much more battle-tested and have more resources available.

tsup supports:
- esm / cjs
- declaration types
- code splitting
- tree-shaking
- and more

It also keeps `use client` directives so this should resolve the issue related to server components. There's a caveat though where code splitting might interfere and not keep directives in chunk files.

Two solutions for that:
- disable code splitting
- add `use client` directives where components are exported i.e. `index.ts` like here https://github.com/radix-ui/primitives/blob/16d6e8409abeacb2b2023ab3db1c0193434f404f/packages/react/avatar/src/index.ts#L1

The second one is better but it will define all components as client components. Generally this is not ideal but this is fine for OnchainKit since everything is done on the client.

Another thing to consider, dev dependencies are inlined when bundling and not dependencies. So if you want OnchainKit to work without users having to install an external lib then you should put it in dev dependencies.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -44,7 +44,7 @@
       ".changeset",
       ".next",
       "coverage",
-      "esm",
+      "dist",
       "framegear",
       "site/docs/dist/"
     ]
--- a/package.json
+++ b/package.json
@@ -5,7 +5,9 @@
   "repository": "https://github.com/coinbase/onchainkit.git",
   "license": "MIT",
   "scripts": {
-    "build": "packemon build --addEngines --addFiles --declaration && npx packemon validate --no-license --no-people --no-repo && tailwindcss -i ./src/styles/index.css -o ./src/tailwind.css --minify && tailwindcss -i ./src/styles/index-with-tailwind.css -o ./src/styles.css --minify",
+    "build": "yarn build:ts && yarn build:css",
+    "build:ts": "tsup",
+    "build:css": "tailwindcss -i ./src/styles/index.css -o ./src/tailwind.css --minify && tailwindcss -i ./src/styles/index-with-tailwind.css -o ./src/styles.css --minify",
     "ci:format": "biome ci --linter-enabled=false --organize-imports-enabled=false",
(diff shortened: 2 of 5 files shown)


kyhyco:
@roushou My preference is to keep using packemon. It's built by one of our past coworker and we use it internally everywhere.

If packemon doesn't work, we can push updates upstream to packemon and the maintainer is _very_ active.

If we need further help on packemon, we can internally pull our packemon experts to help out as well.

Overall, I don't see a huge benefit of replacing packemon with tsup at the moment.

----------------------------------------

For [tree-shaking](https://knip.dev/), we can use knip to remove any unused code. Works super well.

Zizzamia:
@kyhyco before introducing any breaking changes, please wait my +1

roushou:
> Another thing to consider, dev dependencies are inlined when bundling and not dependencies. So if you want OnchainKit to work without users having to install an external lib then you should put it in dev dependencies.

For clarity, I was referring to the case where `tsup` is used for bundling. AFAIK other bundlers (esbuild, rollup, ...) work the other way by inlining dependencies instead of dev dependencies.

> We should discuss this one further. I too think that viem, wagmi, etc. should be a dev dependencies. There could be a potential conflict between version that the user is using vs what we bundle with.

IMO it would be better to not inline Viem/Wagmi/... but declare them as peer dependencies. Given the wide range of features those libraries cover it might be hard to cover them all so users might still reach out to them and cause conflicts.

roushou:
Makes sense to still use packemon. Letting this PR open in case you change your mind. Otherwise feel free to close it.

roushou:
Closing since it's been a while

PR #720, opened 2024-06-25
feat: ConnectWallet component

kyhyco:
**What changed? Why?**
Screenshot 2024-06-25 at 12 57 22 PM

| No ENS Avatar and Name | No ENS Avatar with ENS Name | ENS Avatar + ENS Name |
| --- | --- | --- |
| Screenshot 2024-06-25 at 12 59 41 PM | Screenshot 2024-06-25 at 12 51 36 PM | Screenshot 2024-06-25 at 12 52 39 PM |

Component API
```javascript
<Wallet>
  <ConnectWallet>
    <Avatar className="h-6 w-6" />
    <Name />
  </ConnectWallet>
  <WalletDropdown>
    <IdentityLayout>
      <Avatar />
      <Name>
        <Badge />
      </Name>
      <Address className={color.foregroundMuted} />
    </IdentityLayout>
    <WalletDisconnect />
    <GoToWalletDashboard />
  </WalletDropdown>
</Wallet>
```

**Notes to reviewers**

**How has it been tested?**


mindapivessa:
UI:
- Update the address color to `foreground-muted`/`Gray 600`
Screenshot 2024-06-25 at 4 02 44 PM

- Increase top and bottom padding by 8px each.

kyhyco:
@mindapivessa 

Screenshot 2024-06-25 at 2 07 01 PM

mindapivessa:
@kyhyco can we remove top padding and only add the bottom. I think you just added the top

src/wallet/components/ConnectAccount.tsx
```diff
@@ -1,56 +0,0 @@
...
-import { useAccount, useConnect, useDisconnect } from 'wagmi';
-import type { ConnectAccountReact } from '../types';
-
-/**
- * ConnectAccount
- *  - Connects to the wallet
- *  - Disconnects from the wallet
- *  - Displays the wallet network
- */
-export function ConnectAccount({ children }: ConnectAccountReact) {
```
Zizzamia, line 11: This, will cause a breaking changes. 🤔

Which can be tricky as ConnectAccount is already popular as component. 🤔

Let's avoid this breaking changes for now. So we still keep `ConnectAccount` in the codebase, BUT we remove it from the documentation. So we gently showcase the new one instead, and then we will kill `ConnectAccount` in a three of weeks. After the new one has been already around.

So we avoid a drastic change, but we show the new first, we hide the old, and then when people forgot about the old we kill it.

abcrane123 (approved):
looks good to me !

src/identity/components/IdentityProvider.tsx
```diff
@@ -0,0 +1,34 @@
...
+  return useContext(IdentityContext);
+}
+
+type IdentityProvider = {
+  address?: Address;
+  children: ReactNode;
+  schemaId?: Address | null;
+};
+
+export function IdentityProvider(props: IdentityProvider) {
```
kyhyco, line 20: Colocating all context related functions.

src/internal/hooks/useValue.ts
```diff
@@ -0,0 +1,5 @@
+import { useMemo } from 'react';
+
+export function useValue<T>(object: T): T {
+  return useMemo(() => object, [object]);
+}
```
kyhyco, line 5: this gets rid of the huge `useMemo` block in context providers

src/identity/index.ts
```diff
@@ -4,6 +4,7 @@ export { Avatar } from './components/Avatar';
 export { Badge } from './components/Badge';
 export { Name } from './components/Name';
 export { Identity } from './components/Identity';
+export { IdentityLayout } from './components/IdentityLayout';
```
Zizzamia, line 7: I feel like we should just offer one solution that combines Provider and Layout, and just keep it as `<Identity />`.

Having them split, can be done for internal purposes, but for external, I think we should keep just one. cc @kyhyco

src/internal/hooks/useValue.ts
Zizzamia: ok nice

src/wallet/components/ConnectAccount.test.tsx
```diff
@@ -1,87 +0,0 @@
...
-import { useAccount, useConnect, useDisconnect } from 'wagmi';
-import { ConnectAccount } from './ConnectAccount';
-
-jest.mock('wagmi', () => ({
-  useAccount: jest.fn(),
-  useConnect: jest.fn(),
-  useDisconnect: jest.fn(),
-}));
-
-describe('ConnectAccount Component', () => {
```
Zizzamia, line 16: Should we keep this for now? and deprecate the component and test later.

src/wallet/components/GoToWalletDashboard.tsx
```diff
@@ -0,0 +1,16 @@
+import { cn, pressable, text } from '../../styles/theme';
+import { walletSvg } from './walletSvg';
+
+export function GoToWalletDashboard() {
+  return (
+    <a
```
Zizzamia, line 6: I don't think this should be his own component. But instead have something like
```ts
<WalletDropdownItem>
  // add your own links or text
</WalletDropdownItem>
```

Mostly because I want us to teach how to create many links inside a dropdown, and avoid to force them having a special component for this.

src/wallet/components/WalletDisconnect.tsx
```diff
@@ -0,0 +1,28 @@
+import { useConnect, useDisconnect } from 'wagmi';
+import { cn, pressable, text } from '../../styles/theme';
+import { disconnectSvg } from './disconnectSvg';
+import { useCallback } from 'react';
+
+export function WalletDisconnect() {
```
Zizzamia, line 6: Let's have this as `WalletDropdownDisconnet` so it makes sense this is an Item inside the Dropdown.

Otherwise it feels like a random button I can fit anywhere even outside the dropdownn.

src/wallet/components/WalletDropdown.tsx
```diff
@@ -1,15 +1,28 @@
...
+  const { isOpen } = useWalletContext();
+
+  const { address } = useAccount();
+
+  if (!isOpen || !address) {
+    return null;
+  }
+
+  return (
+    <IdentityProvider address={address}>
```
Zizzamia, line 17: smart

src/wallet/components/walletSvg.tsx
```diff
@@ -0,0 +1,18 @@
+export const walletSvg = (
```
Zizzamia, line 1: I guess we could keep a custom component just for this special case, but ideally we still need one to create new links for the dropdown.

Zizzamia:
@kyhyco I like this direction a lot, and I like how you made it super easy to include IdentityProvider.
```ts
<Wallet>
  <ConnectWallet>
    <Avatar className="h-6 w-6" />
    <Name />
  </ConnectWallet>
  <WalletDropdown>
    <IdentityLayout>
      <Avatar />
      <Name>
        <Badge />
      </Name>
      <Address className={color.foregroundMuted} />
    </IdentityLayout>
    <WalletDisconnect />
    <GoToWalletDashboard />
  </WalletDropdown>
</Wallet>
```

So there are only 3 points I would like to reflect as follow up PR.
1. Should we ship `<IdentityLayout />`, I feel like it removes the magic. I feel like that's a special case of a WalletDropdownItem, so should this instead being named `<WalletDropodownIdentity />` ?
2.  `<WalletDisconnect />` let's probably name it `<WalletDropdownDisconnect />` so it's clear is part of the dropdown
3. we need a component that a developer can safely extend the dropdown, with new Links. Maybe `<WalletDropdownLink />`
4. if we want to keep `<GoToWalletDashboard />` to make it super easy for this particular case, than let's try to have it with a name normalized for this case so `<WalletDropdownDashboardLink />`

Zizzamia (approved):

PR #729, opened 2024-06-27
feat: EthBalance component

kyhyco:
**What changed? Why?**
- Add `EthBalance` component
- `getRoundedAmount` renders correctly when amount is greater than 0 but less than the fractionDigit
- Add `mock` and `silenceError` testing helper functions
- Add `findComponent` helper function

**Notes to reviewers**

**How has it been tested?**
locally

### DIFF
--- a/src/identity/components/Name.test.tsx
+++ b/src/identity/components/Name.test.tsx
@@ -10,6 +10,7 @@ import { useAttestations } from '../hooks/useAttestations';
 import { Name } from './Name';
 import { Badge } from './Badge';
 import { useIdentityContext } from './IdentityProvider';
+import { silenceError } from '../../internal/testing';
 
 jest.mock('../hooks/useName', () => ({
   useName: jest.fn(),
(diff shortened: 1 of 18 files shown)


src/identity/components/Name.test.tsx
```diff
@@ -43,11 +44,13 @@ describe('OnchainAddress', () => {
     (useIdentityContext as jest.Mock).mockReturnValue({
       schemaId: '0x123',
     });
+    const restore = silenceError();
```
kyhyco, line 47: In jest test, if a test is expected to throw add these in between:
```
const restore = silenceError();
expect(...).toThrow(...);
restore();
```

This will makes the false console.error go away in test results cc @Zizzamia @abcrane123 @cpcramer @ilikesymmetry

Zizzamia: nice

src/utils/getRoundedAmount.ts
```diff
@@ -3,5 +3,13 @@ export function getRoundedAmount(balance: string, fractionDigits: number) {
     return balance;
   }
   const parsedBalance = Number.parseFloat(balance);
-  return Number(parsedBalance)?.toFixed(fractionDigits).replace(/0+$/, '');
+  const result = Number(parsedBalance)
+    ?.toFixed(fractionDigits)
+    .replace(/0+$/, '');
+
+  if (parsedBalance > 0 && Number.parseFloat(result) === 0) {
```
Zizzamia, line 10: nit. add a line of comment explaining this case. I know it has a test, which is great. But have a line of comment helps remember what this part is covering for.

PR #747, opened 2024-06-28
feat: Add isSliced option to the Name and Address component

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/identity/components/Address.tsx
+++ b/src/identity/components/Address.tsx
@@ -3,7 +3,11 @@ import { getSlicedAddress } from '../getSlicedAddress';
 import { cn, text } from '../../styles/theme';
 import type { AddressReact } from '../types';
 
-export function Address({ address = null, className }: AddressReact) {
+export function Address({
+  address = null,
+  isSliced = true,
+  className,
+}: AddressReact) {
   const { address: contextAddress } = useIdentityContext();
   if (!contextAddress && !address) {
     throw new Error(
@@ -11,9 +15,11 @@ export function Address({ address = null, className }: AddressReact) {
     );
   }
 
+  const displayAddress = contextAddress ?? address;
+
   return (
     <span data-testid="ockAddress" className={cn(text.label2, className)}>
-      {getSlicedAddress(contextAddress ?? address)}
+      {isSliced ? getSlicedAddress(displayAddress) : displayAddress}
     </span>
   );
 }
--- a/src/identity/components/Name.tsx
+++ b/src/identity/components/Name.tsx
@@ -12,6 +12,7 @@ import { cn, text } from '../../styles/theme';
  */
 export function Name({
   address = null,
+  isSliced = true,
   className,
   children,
   ...props
@@ -23,8 +24,10 @@ export function Name({
     );
   }
 
+  const displayAddress = contextAddress ?? address;
+
   const { data: name, isLoading } = useName({
-    address: contextAddress ?? address,
+    address: displayAddress,
   });
 
   const badge = useMemo(() => {
@@ -37,6 +40,10 @@ export function Name({
     return <span className={className} />;
   }
 
+  const formattedAddress = isSliced
+    ? getSlicedAddress(displayAddress)
+    : displayAddress;
+
   // It displays the ENS name if available; otherwise, it shows either a sliced version of the address
   // or the full address, based on the 'sliced' prop. By default, 'sliced' is set to true.
   return (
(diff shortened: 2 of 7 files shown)


src/identity/components/Address.tsx
```diff
@@ -3,17 +3,23 @@ import { getSlicedAddress } from '../getSlicedAddress';
...
+  className,
+}: AddressReact) {
   const { address: contextAddress } = useIdentityContext();
   if (!contextAddress && !address) {
     throw new Error(
       'Address: an Ethereum address must be provided to the Identity or Address component.',
     );
   }
 
+  const displayAddress = contextAddress ?? address;
```
Zizzamia, line 18: The name `display` makes me think this is something we are showcasing. But is not quite correct, as this is more an `address` we use to then do other things.

Same story in the other file.

So what's actually is this `address`? After all this is the `address` that identifies the user using this component. Or provided address.

What name should we pick here instead of `displayAddress`?

And again whatever we pick will be used in the other file as well.

src/identity/types.ts
```diff
@@ -12,6 +12,7 @@ import type { Address, Chain } from 'viem';
  */
 export type AddressReact = {
   address?: Address | null; // The Ethereum address to render.
+  isSliced?: boolean; // Determines if the displayed address should be sliced.
   className?: string; // Optional className override for top span element.
```
Zizzamia, line 16: always `alphabetical`, it keeps the code easy to read

src/identity/types.ts
```diff
@@ -141,7 +142,7 @@ export type IdentityReact = {
 export type NameReact = {
   address?: Address | null; // Ethereum address to be displayed.
   className?: string; // Optional className override for top span element.
-  sliced?: boolean; // Determines if the address should be sliced when no ENS name is available.
+  isSliced?: boolean; // Determines if the displayed address should be sliced.
```
Zizzamia, line 145: fix it as `alphabetical`

src/identity/components/Address.tsx
cpcramer: `effectiveAddress` - the address being effectively used by the component.

`currentAddress`- The current address in use, highlighting that this is the active address for the component.

`resolvedAddress` - Final resolved address being used by the component, whether it comes from the context or is provided directly as a prop.

cpcramer: I'm going to update to `resolvedAddress`, I believe that it's the most appropriate naming here.

src/identity/core/getName.ts
```diff
@@ -14,5 +13,5 @@ export const getName = async ({
     address,
   });
 
-  return ensName ?? getSlicedAddress(address);
+  return ensName ?? null;
```
Zizzamia, line 16: This introduce a breaking change that we should document in the Changeset.

src/identity/components/Name.tsx
```diff
@@ -23,8 +24,10 @@ export function Name({
     );
   }
 
+  const resolvedAddress = contextAddress ?? address;
```
Zizzamia, line 27: why the word `resolved`? I think is best just to call it `accountAddress`.

src/identity/components/Name.tsx
```diff
@@ -46,11 +53,9 @@ export function Name({
         className={cn(text.headline, className)}
         {...props}
       >
-        {name ?? getSlicedAddress(contextAddress ?? address)}
+        {name ?? formattedAddress}
```
Zizzamia, line 56: If `<Name>` returns either a `name` or a `formattedAddress`, why the `getName` utility doesn't do the same?

A component is just a visualizzation of an utility. So should `getName` return `formattedAddress`?

src/identity/components/Name.tsx
cpcramer: Because that's the address that we settle on between `contextAddress` and `address`. 

I'll update to `accountAddress`

src/identity/core/getName.ts
cpcramer: updated the changeset

src/identity/components/Name.tsx
cpcramer: That's a good point that the component should reflect the utility.

The utility fetches the `ENS` name for a given address. If the this lookup returns `null`, I'd expect the function to also return `null`. 

I would not expect `isName` to return something other than `null` or the `ENS` name. 

If we want to have `isName` return a formatted address, then we also must pass the `isSliced` boolean to the `isName` function, which I feel would be strange.

src/identity/types.ts
```diff
@@ -140,9 +141,8 @@ export type IdentityReact = {
  */
 export type NameReact = {
   address?: Address | null; // Ethereum address to be displayed.
-  className?: string; // Optional className override for top span element.
-  sliced?: boolean; // Determines if the address should be sliced when no ENS name is available.
   children?: ReactNode; // Optional attestation by passing Badge component as its children
+  className?: string; // Optional className override for top span element.
```
cpcramer, line 145: Keeping this alphabetic ordering change in.

Zizzamia (approved):

PR #761, opened 2024-07-08
feat: [Swap] ERC-20 Approval Flow

0xAlec:
**What changed? Why?**
- fix ERC20 approval and swap flow
- remove `onSubmit` and add callback hooks for `onError` and `onSuccess`
- add transaction states for swaps
- add some error codes and add `getSwapError` for returning the error message used in `SwapMessage`

**Notes to reviewers**

**How has it been tested?**
- see video on Slack

made through Swap component:
https://basescan.org/tx/0x780891a40187ade80d9bbd7384f7389d33b27b55a7f97460558f49016546a644


src/swap/components/SwapButton.test.tsx
```diff
@@ -70,22 +70,6 @@ describe('SwapButton', () => {
     expect(button).toBeDisabled();
   });
 
-  it('calls handleSubmit with onSubmit when clicked', () => {
```
Zizzamia, line 73: Why remove it?

src/swap/components/SwapButton.tsx
```diff
@@ -3,14 +3,11 @@ import { Spinner } from '../../internal/loading/Spinner';
 import { useSwapContext } from './SwapProvider';
 import type { SwapButtonReact } from '../types';
 
-export function SwapButton({
-  className,
-  disabled = false,
-  onSubmit,
```
Zizzamia, line 9: So what's going on to `onSubmit` going forwar?

src/swap/types.ts
```diff
@@ -141,7 +142,8 @@ export type SwapContextType = {
   from: SwapUnit;
   to: SwapUnit;
   loading: boolean;
-  handleSubmit: (f: (swapTransaction: BuildSwapTransaction) => void) => void;
+  pendingTransaction: boolean;
```
Zizzamia, line 144: Boolean should always start with a prefix like `is` or `has`.

so in this case we want `isPendingTransaction`.

Btw is there a chance where something like `transactionStatus` is a better experience? and `pending` is just a status in it?

src/swap/components/SwapProvider.tsx
```diff
@@ -184,21 +184,74 @@ export function SwapProvider({
           return handleError({ swapError: response });
         }
 
-        onSubmit?.(response);
+        const { transaction, approveTransaction } = response;
+
+        if (approveTransaction?.data) {
```
Zizzamia, line 189: This condition has "a lot" of code.

Same for the "catch" part. How all this part can be exported into an utility fuction, and have stronger unit-test around that function?

src/swap/components/SwapButton.tsx
0xAlec: removed and no longer overridable

src/swap/components/SwapProvider.tsx
```diff
@@ -184,21 +184,74 @@ export function SwapProvider({
...
       }
     },
-    [address, handleError, from.amount, from.token, to.token],
+    [
+      address,
+      config,
+      handleError,
+      from.amount,
+      from.token,
+      sendTransactionAsync,
```
Zizzamia, line 244: What's `sendTransactionAsync`?

src/swap/components/SwapButton.tsx
Zizzamia: if we remove that, we still need a way to listen to `onErrors` and `onSuccess`.

Somewhere we need to give back to the developers the information of the Swap experience.

src/swap/components/SwapProvider.tsx
0xAlec: https://wagmi.sh/react/api/hooks/useSendTransaction#mutate-async

so we can `await` it (to use the transaction hash)

src/swap/types.ts
0xAlec: changed to `isTransactionPending`

src/swap/components/SwapButton.tsx
0xAlec: added hooks for both of these

.changeset/sharp-oranges-kick.md
```diff
@@ -0,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
+
+**feat**: [Swap] ERC-20 Approval Flow by @0xAlec #761
```
Zizzamia, line 5: So the fact we are removing `onSubmit` will create a breaking changes.

We need to make sure we document this in the changeset. Like you see in some cases in the CHANGELOG https://github.com/coinbase/onchainkit/blob/main/CHANGELOG.md

src/swap/components/SwapProvider.tsx
```diff
@@ -10,14 +10,12 @@ import { isSwapError } from '../core/isSwapError';
...
-  BuildSwapTransaction,
-  SwapContextType,
-} from '../types';
+import type { SwapError, SwapErrorState, SwapContextType } from '../types';
 import type { Token } from '../../token';
-import type { Address } from 'viem';
+import type { Address, TransactionReceipt } from 'viem';
+import { useSendTransaction, useConfig, type BaseError } from 'wagmi';
+import { waitForTransactionReceipt } from '@wagmi/core';
+import { USER_REJECTED_ERROR_CODE } from '../constants';
```
Zizzamia, line 18: This all order of imports feels off.

it should be
```
// we do all the external libraries
import { waitForTransactionReceipt } from '@wagmi/core';
import {
  createContext,
  useContext,
  useCallback,
  useMemo,
  useState,
} from 'react';
import { useSendTransaction, useConfig } from 'wagmi';

// than all the ./ level
import { useSwapBalances } from './useSwapBalances';
import { buildSwapTransaction } from '../core/buildSwapTransaction';

// than all the ../ level
import { isSwapError } from '../core/isSwapError';
import { getSwapQuote } from '../core/getSwapQuote';
import { USER_REJECTED_ERROR_CODE } from '../constants';

// than all the ../../
import { formatTokenAmount } from '../../utils/formatTokenAmount';

// than the types
import type { Address, TransactionReceipt } from 'viem';
import type { BaseError } from 'wagmi';
```

and ofcourse not comments and no spaces.

src/swap/components/SwapProvider.tsx
```diff
@@ -95,6 +94,10 @@ export function SwapProvider({
 
   const { from, to } = useFromTo(address);
 
+  const { sendTransactionAsync } = useSendTransaction();
```
Zizzamia, line 97: Can we start adding some line comments on what each thing is for.... otherwise this is going to get hard to mantain.

src/swap/components/SwapProvider.tsx
```diff
@@ -95,6 +94,10 @@ export function SwapProvider({
...
+
+  const config = useConfig();
```
Zizzamia, line 99: Like here
```
// Get wagmi config, to use with the waitForTransactionReceipt
const config = useConfig();
```

so as someone reads the code gets better understanding what's going on in a few months from now.

src/swap/components/SwapProvider.tsx
```diff
@@ -163,7 +166,8 @@ export function SwapProvider({
 
   const handleSubmit = useCallback(
     async function handleSubmit(
-      onSubmit?: (swapTransaction: BuildSwapTransaction) => void,
+      onError?: (error: SwapError) => void,
+      onSuccess?: (txReceipt: TransactionReceipt) => void | Promise<void>,
```
Zizzamia, line 170: asking out loud, should `txReceipt` just be `transactionReceipt`? just for better clarity.

src/swap/components/SwapProvider.tsx
```diff
@@ -173,6 +177,7 @@ export function SwapProvider({
       handleError({ swapError: undefined });
 
       try {
+        /* fetch the transaction */
```
Zizzamia, line 180: Line comments are done by using `//`
The `/*` it's only used as multi-line comments at the beggining of a function.

src/swap/components/SwapProvider.tsx
```diff
@@ -184,21 +189,85 @@ export function SwapProvider({
           return handleError({ swapError: response });
         }
 
-        onSubmit?.(response);
+        const { transaction, approveTransaction } = response;
+
+        /* for swaps from ERC-20 tokens,
```
Zizzamia, line 194: Use `//` to write the comment.

src/swap/components/SwapProvider.tsx
```diff
@@ -184,21 +189,85 @@ export function SwapProvider({
...
+            value: approveTransaction.value,
+            data: approveTransaction.data,
+          });
+          await waitForTransactionReceipt(config, {
+            hash: approveTxHash,
+            confirmations: 1,
+          });
+          setPendingTransaction(false);
+        }
+
```
Zizzamia, line 210: There is a lot of code that miss unit-test in this file. And this code is super critical.
Let's take the time to unpack this Provider in utility functions that can be atomically tested.

src/swap/constants.ts
```diff
@@ -1,6 +1,8 @@
 export const LOW_LIQUIDITY_ERROR_CODE = 'SWAP_QUOTE_LOW_LIQUIDITY_ERROR';
+export const TOO_MANY_REQUESTS_ERROR_CODE = 'TOO_MANY_REQUESTS_ERROR';
```
Zizzamia, line 2: This line is breaking the alphabetical order.

src/swap/core/getSwapErrorCode.ts
```diff
@@ -3,6 +3,7 @@ import {
   GENERAL_SWAP_BALANCE_ERROR_CODE,
   GENERAL_SWAP_QUOTE_ERROR_CODE,
   LOW_LIQUIDITY_ERROR_CODE,
+  TOO_MANY_REQUESTS_ERROR_CODE,
```
Zizzamia, line 6: this like is breaking the alphabetical order.

src/swap/core/getSwapMessage.ts
```diff
@@ -1,12 +1,19 @@
...
 import type { GetSwapMessageParams } from '../types';
 
 export enum SwapMessage {
   BALANCE_ERROR = 'Error fetching token balance',
+  CONFIRM_IN_WALLET = 'Confirm in wallet',
   INCOMPLETE_FIELD = 'Complete the fields to continue',
   INSUFFICIENT_BALANCE = 'Insufficient balance',
   LOW_LIQUIDITY = 'Liquidity too low for the token',
   SWAP_IN_PROGRESS = 'Swap in progress...',
+  TOO_MANY_REQUESTS = 'Too many requests. Please try again later.',
```
Zizzamia, line 15: All of those should be alphabetically ordered.

src/swap/core/getSwapMessage.ts
```diff
@@ -39,14 +52,28 @@ export function getSwapMessage({
   if (!error) {
     return '';
   }
-  // handle invalid params error
+  // handle specific error codes
+  if (
+    Object.values(error)?.find(
```
Zizzamia, line 57: All `Object.values(error)?.find(` could be an utility function. And make this code cleaner.

0xAlec: added `getSwapError`

src/swap/components/SwapProvider.tsx
0xAlec: done

src/swap/components/SwapProvider.tsx
0xAlec: done

src/swap/components/SwapProvider.tsx
0xAlec: added

.changeset/sharp-oranges-kick.md
0xAlec: added to changelog (but not under a version since we haven't released it yet - will let you take that on)

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Changelog
 
+- **feat**: [Swap] ERC-20 Approval Flow by @0xAlec #761. This is a breaking change that removes the `onSubmit` functionality from the `SwapButton` component and adds an approval flow for swapping from ERC-20 tokens.
```
Zizzamia, line 3: @0xAlec don't need to write this, as it is automated. But instead add this info in the `changeset` file.

src/swap/components/SwapButton.tsx
```diff
@@ -5,12 +5,15 @@ import type { SwapButtonReact } from '../types';
 
 export function SwapButton({
   className,
+  onError,
+  onSuccess,
   disabled = false,
```
Zizzamia, line 10: Always keep alphabetical order with props.

CHANGELOG.md
0xAlec: moved to changeset

src/swap/components/SwapButton.tsx
0xAlec: I thought default values should be last? at least in some languages if there is a default value the style guide indicates they have to be last in the ordered parameters, but that may be python

src/swap/components/SwapProvider.tsx
```diff
@@ -184,25 +201,117 @@ export function SwapProvider({
...
+    isTransactionPending,
     handleAmountChange,
     handleToggle,
     handleSubmit,
   });
 
   return <SwapContext.Provider value={value}>{children}</SwapContext.Provider>;
 }
+
+export async function processSwapTransaction({
```
Zizzamia, line 260: Niiice love this, let's have this in `core` as his own file. In this will make this file smaller and easier to mantain.
And ofcourse the tests will be on their own file as well.

src/swap/components/SwapButton.tsx
0xAlec: fixed in new diff

src/swap/constants.ts
Zizzamia: cc @0xAlec

src/swap/core/getSwapError.ts
```diff
@@ -0,0 +1,36 @@
+import { SwapMessage } from './getSwapMessage';
+import {
+  LOW_LIQUIDITY_ERROR_CODE,
+  TOO_MANY_REQUESTS_ERROR_CODE,
+  USER_REJECTED_ERROR_CODE,
+} from '../constants';
+import type { SwapErrorState } from '../types';
+
+export function getSwapError(error: SwapErrorState): string | undefined {
```
Zizzamia, line 9: Niiice

src/swap/components/SwapProvider.tsx
0xAlec: I would push back/disagree with this because imo `core` should be vanillaJS only and `wagmi` is a dependency in this function

0xAlec: Thinking out loud but should we have a `components/utils` for decomposing complex business logic into testable functions, for cases like these? cc @abcrane123

Zizzamia: We should avoid over deep leveling stuff. as it is hard to find them.

Let's do `swap/utils` instead.

Zizzamia (approved):

Issue #758, opened 2024-07-08
Feature Request: add Storybook


roushou:
### Describe the solution you'd like

The development workflow can be improved by using Storybook which provides isolated testing, accessibility checks, interaction testing and more.

Some useful addons:
- [a11y](https://storybook.js.org/addons/@storybook/addon-a11y)
- [source](https://storybook.js.org/addons/@storybook/addon-storysource)
- [visual testing](https://storybook.js.org/addons/@chromatic-com/storybook) (requires Chromatic)
- [figma](https://storybook.js.org/addons/@storybook/addon-designs)

Network requests can be mocked using a [service worker](https://mswjs.io/) (also useful for happy paths). It does require some additional steps to keep mocks in [sync](https://mswjs.io/docs/recipes/keeping-mocks-in-sync).

Also update CI to catch bugs and regressions.

PR #765, opened 2024-07-09
feat: add Storybook

roushou:
close https://github.com/coinbase/onchainkit/issues/758


roushou:
This should be a good starting point.

`yarn watch` should be run alongside `yarn storybook:dev` to get live changes of styles.

Insight:
- Imo `TokenSearch` should allow users to specify the source of tokens to search in. It also requires an api key to be able to do real search.
-  `Avatar` has 2 loading states which is very weird, see story (refresh page on `WithBadge` story`
- `Badge` doesn't render correctly when used individually, see story
- Accessibility checks are context dependent i.e. `TokenSelectDropdown` doesn't have a11y issues because it's built with `div` tags and not `select/options`
- Documentation is auto-generated using JSDoc annotations from components, it would be great to document components if necessary

I was thinking about mocking requests using service workers but I find that it's actually better to do real requests to know if things are broken or not.

Regarding CI, it can be done programmatically using [test-runner](https://storybook.js.org/docs/writing-tests/test-runner) or visually and collaboratively using [Chromatic](https://www.chromatic.com/). I prefer the latter since it's visual but it depends on if you are looking to use Chromatic or not @fakepixels @Zizzamia

src/identity/components/Name.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+      return (
+        <QueryClientProvider client={new QueryClient()}>
+          <Story />
+        </QueryClientProvider>
+      );
+    },
+  ],
+  parameters: {
+    layout: 'centered',
+  },
```
siddharthkul, line 19: Let's move this to `.storybook/preview.ts`? Looks like it is shared among all stories.

src/identity/components/EthBalance.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+  title: 'Identity/EthBalance',
+  component: EthBalance,
+  decorators: [
+    (Story) => {
+      return (
+        <WagmiProvider config={wagmiConfig}>
+          <QueryClientProvider client={new QueryClient()}>
+            <Story />
+          </QueryClientProvider>
+        </WagmiProvider>
```
siddharthkul, line 24: Should we move this to preview.ts as well?

.gitignore
```diff
@@ -57,3 +57,5 @@ package-lock.json
 # VitePress
 docs/.vitepress/cache
 docs/.vitepress/dist
+
+*storybook.log
```
siddharthkul, line 61: I'd also add recommend ignoring `storybook-static/*`

```
# Storybook
*storybook.log
storybook-static/*
```

src/identity/components/Name.stories.tsx
roushou: Good call. Moved all of them inside `.storybook/preview.ts`

.gitignore
roushou: Good call. I though running `npx storybook init` would update correctly `.gitignore`.

src/identity/components/EthBalance.stories.tsx
roushou: I'd favor explicitness because:

- Not every components needs to be wrapped inside
- Some components only need to be wrapped inside `QueryClientProvider`

More verbose but it's clear what's needed when reading stories.

Zizzamia:
@roushou can you get the Build pass, please. So than I can review it.

roushou:
@Zizzamia I can't reproduce locally. It seems to come from this line

https://github.com/coinbase/onchainkit/blob/ab93ca06dd9f922e177a27751d6cd9ce7aa53971/.github/actions/setup/action.yml#L17

Which makes build fail on purpose on CI

https://classic.yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile

src/styles/tailwind-base.css
```diff
@@ -33,3 +33,4 @@
 }
 
 @tailwind utilities;
+@tailwind components;
```
Zizzamia, line 36: Why this? @roushou 

We have not see the need to have those in a Tailwind application.

Zizzamia:
Ok I am going to take care of the Build issue.

src/styles/tailwind-base.css
roushou: Right it's not needed here. I added it because I'm used to have it and I often define custom classes in `@layer components` but there's none in OnchainKit.

PR #777, opened 2024-07-11
docs: Add Wrap Your App with OnchainProviders section

cpcramer:
**What changed? Why?**
Add `Wrap Your App with OnchainProviders` section in the getting started guide. 

This is a follow up from my dogfooding feedback [here](https://coinbase.slack.com/archives/C06NZU31P4N/p1719185284498779) - _"It wasn't clear that I had to wrap my root component in the OnchainProviders - I had to reference the Wagmi docs to remember this step."_

**Notes to reviewers**

**How has it been tested?**
Screenshot 2024-07-11 at 11 31 11 AM

Screenshot 2024-07-11 at 11 31 22 AM

### DIFF
--- a/site/docs/pages/getting-started.mdx
+++ b/site/docs/pages/getting-started.mdx
@@ -157,11 +157,13 @@ export const wagmiConfig = createConfig({
   },
 });
 ```
-
 :::
+
 ### Wrap Your App with OnchainProviders
 
 Wrap your app with the `OnchainProviders` component to enable OnchainKit components and utilities.
+
+:::code-group
 
 ```tsx [layout.tsx]
 import { OnchainProviders } from './OnchainProviders' // [!code focus]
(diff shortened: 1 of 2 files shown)


site/docs/pages/getting-started.mdx
```diff
@@ -159,6 +159,21 @@ export const wagmiConfig = createConfig({
...
+import { OnchainProviders } from './OnchainProviders' // [!code focus]
+
+export default function RootLayout({ children }) {
+  return (
+    <OnchainProviders> // [!code focus]
+      { children }
+    </OnchainProviders> // [!code focus]
+  )
+}  
+```
```
cpcramer, line 176: We could also have the filename be `app.tsx` and the function be called `app`.

cpcramer:
I considered adding an additional `OnchainProviders.tsx` tab. I don't think it's necessary because the developer can reference this code which is directly above this step.

site/docs/pages/getting-started.mdx
```diff
@@ -159,6 +159,21 @@ export const wagmiConfig = createConfig({
 ```
 
 :::
+### Wrap Your App with OnchainProviders
+
+Wrap your app with the `OnchainProviders` component to enable OnchainKit components and utilities.
+
+```tsx [layout.tsx]
+import { OnchainProviders } from './OnchainProviders' // [!code focus]
```
Zizzamia, line 167: What's inside here? Let's have the second tab to show what's inside there.

cpcramer:
> @cpcramer you mean this https://onchainkit.xyz/config/onchainkit-provider? we should for sure improve this page.

Yes! It's just too "far" from the getting started guide. Perhaps we can keep https://onchainkit.xyz/config/onchainkit-provider to use as a deeper dive into the code. 

But I think we need to mention wrapping the developers application with our provider in the getting started page

Zizzamia (approved):

PR #781, opened 2024-07-11
Feat: Add chain to useName

kirkas:
**What changed? Why?**
- **feat**: Added `chain` option to `useName` function for L2 chain name resolution support. 
- **feat**: Added `chain` option to `<Name>` component for L2 chain name resolution support.
- **fix**: Modified `getName` to return a rejected promise instead of an error. 
- **fix**: Disabled `retry` in `getNewReactQueryTestProvider` to run tests faster and avoid timeouts. 

**How has it been tested?**
- locally with 100% passing
- 100% test coverage
- In storybook

image

### DIFF
--- a/src/identity/core/getName.ts
+++ b/src/identity/core/getName.ts
@@ -22,7 +22,7 @@ export const getName = async ({
   const chainSupportsUniversalResolver = chainIsEthereum || chainIsBase;
 
   if (!chainSupportsUniversalResolver) {
-    throw Error(
+    return Promise.reject(
       'ChainId not supported, name resolution is only supported on Ethereum and Base.',
     );
   }
--- a/src/identity/hooks/useName.test.tsx
+++ b/src/identity/hooks/useName.test.tsx
@@ -9,7 +9,6 @@ import { getNewReactQueryTestProvider } from './getNewReactQueryTestProvider';
 import { describe, beforeEach, it, expect, vi, type Mock } from 'vitest';
 import { base, optimism } from 'viem/chains';
 
-
 vi.mock('../../network/client');
 vi.mock('../../network/getChainPublicClient', () => ({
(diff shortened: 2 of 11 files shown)


src/identity/core/getName.ts
```diff
@@ -22,7 +22,7 @@ export const getName = async ({
   const chainSupportsUniversalResolver = chainIsEthereum || chainIsBase;
 
   if (!chainSupportsUniversalResolver) {
-    throw Error(
+    return Promise.reject(
```
Zizzamia, line 25: Oh brilliant

src/identity/hooks/useName.ts
```diff
@@ -1,10 +1,12 @@
 import { useQuery } from '@tanstack/react-query';
-import type { Address } from 'viem';
 import { getName } from '../core/getName';
+import type { Address, Chain } from 'viem';
 import type { GetNameReturnType } from '../types';
+import type { QueryOptions } from '@tanstack/react-query';
 
 type UseNameOptions = {
   address: Address; // The Ethereum address for which the ENS name is to be fetched.
+  chain?: Chain;
```
Zizzamia, line 9: Very much a nit, let's add a line comment here, like we have for address, and let's move this in the `types` folder.

src/identity/hooks/useName.test.tsx
```diff
@@ -56,4 +56,47 @@ describe('useName', () => {
...
+
+    // Wait for the hook to finish fetching the ENS name
+    await waitFor(() => {
+      // Check that the ENS name and loading state are correct
+      expect(result.current.data).toBe(testEnsName);
+      expect(result.current.isLoading).toBe(false);
+    });
+  });
+
+  it('returns error for unsupported chain ', async () => {
```
Zizzamia, line 82: Niiiice tests.

PR #788, opened 2024-07-12
docs: Swap introduction page

cpcramer:
**What changed? Why?**
Add Swap introduction page 

We want to ensure that all of the components in the sidebar in docs are formatted appropriately with an "introduction" page like "Identity" and "Token" components. This updates the Swap component to be consistent with the other formats. 

**Notes to reviewers**

**How has it been tested?**
Screenshot 2024-07-12 at 11 45 27 AM

### DIFF
--- a/.changeset/nice-fireants-rescue.md
+++ b/.changeset/nice-fireants-rescue.md
@@ -1,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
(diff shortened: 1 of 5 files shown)


Zizzamia:
This page is not needed it anymore, as the all concept of Intro was deprecated in favours of Guides.

PR #792, opened 2024-07-12
Feat: getName: Default to ENS name when custom chain name is not available

kirkas:
**What changed? Why?**
- **fix**: Modified `getName` to default to ENS name when custom chain name is not available.

**How has it been tested?**
Add tests, 100% coverage

### DIFF
--- a/src/identity/core/getName.ts
+++ b/src/identity/core/getName.ts
@@ -39,11 +39,11 @@ export const getName = async ({
     });
 
     if (baseName) {
-      return baseName
-    } else {
-      // Default to mainnet
-      client = getChainPublicClient(mainnet)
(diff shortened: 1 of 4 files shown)


src/identity/core/getName.ts
```diff
@@ -27,18 +27,23 @@ export const getName = async ({
...
       functionName: 'name',
       args: [addressReverseNode],
     });
-    return baseEnsName ?? null;
+
+    if (baseName) {
+      return baseName;
+    }
+
+    // Default to mainnet
```
Zizzamia, line 45: Love the comments

Zizzamia (approved):

PR #785, opened 2024-07-12
Feat: add chain to IdentityProvider and Identity

kirkas:
**What changed? Why?**

- **feat**: Added `chain` option to `<IdentityProvider>` for L2 chain name resolution support.
- **feat**: Added `chain` option to `<Identity>` component for L2 chain name resolution support.
- **fix**: Modify `<Name>` to prioritize its own address/chain props over the provider's. 
- **fix**: Modify `<Address>`, `<Avatar>`, `<EthBalance>` to prioritize its own address prop over the provider's. 

**Notes to reviewers**

**How has it been tested?**
Locally 100% test passing / coverage

PR #802, opened 2024-07-15
chore: sort imports

roushou:
This PR makes use of the organize import feature of Biome by:

- adding `biome check --write .` in the scripts.
- simplifying `ci:lint` and `ci:format` scripts to a single `ci` script that uses all of `biome ci` capabilities.
- simplifying `format` and `lint` github workflows to a single `lint` workflow

`biome check` has already been run on the codebase.

Regarding https://github.com/coinbase/onchainkit/pull/785#discussion_r1676286820 @Zizzamia. There are some limitations compared to `prettier-plugin-sort-imports` and `eslint-plugin-import/order` which allow to configure sorting rules in the config. This is currently not possible with Biome. There's an RFC open https://github.com/biomejs/biome/discussions/3015.

For now, grouping can be done by manually adding blank lines around imports https://biomejs.dev/analyzer/import-sorting/#grouped-imports. Imports order is determined by "how far" imported modules are, see https://biomejs.dev/analyzer/import-sorting/#how-imports-are-sorted.


vitest.config.ts
```diff
@@ -20,10 +20,10 @@ export default defineConfig({
       ],
       reportOnFailure: true,
       thresholds: {
-        statements: 95.34,
+        statements: 95.33,
         branches: 97.7,
         functions: 90.5,
-        lines: 95.34,
+        lines: 95.33,
```
roushou, line 26: Decreased thresholds for coverage to pass

Zizzamia:
Could it be possible if we run ESLint after Biome? and use that as a way to organize imports with the settings from Build Onchain Apps?

Also, I kind of like keep Lint and Format split, so when they fails folks know what it did failed.

roushou:
Added back github workflows.

I just had a look at eslint and didn't managed to make it work without big tradeoffs.

BOAT uses a config file that is now deprecated by eslint in favor of "flag config files" (see there https://eslint.org/docs/latest/use/configure/configuration-files) and `eslint-plugin-import` doesn't work with it, there's an open issue https://github.com/import-js/eslint-plugin-import/issues/2556.

The native rule [sort-imports](https://eslint.org/docs/latest/rules/sort-imports) works but doesn't support automatic fix fully. Running with `--fix` flag currently still shows 219 errors that need to be fixed manually. This rule is more simplistic than `eslint-plugin-import` with things like [grouping](https://eslint.org/docs/latest/rules/sort-imports#allowseparatedgroups) that are not as much configurable.

Zizzamia:
Hey @roushou , I end up committing this with separate PRs. Thank you again for starting this thread, it was really useful.

PR #831, opened 2024-07-17
chore: adjust transaction action component

abcrane123:
**What changed? Why?**
- use txn id to get txn hash for chain explorere
- remove action element for in progress status 

**Notes to reviewers**

**How has it been tested?**


src/transaction/components/TransactionProvider.tsx
```diff
@@ -75,6 +84,7 @@ export function TransactionProvider({
     setTransactionId,
     status,
     transactionId,
+    transactionHash: callsStatus?.receipts?.[0]?.transactionHash,
```
Zizzamia, line 87: @abcrane123 let's get an utility for this. something like `getTransactionHash`, and for now have it as `utils` in the transaction module.

src/transaction/components/TransactionProvider.tsx
```diff
@@ -42,6 +43,14 @@ export function TransactionProvider({
     setTransactionId,
   });
 
+  const { data: callsStatus } = useCallsStatus({
```
Zizzamia, line 46: Let's have the habit that for all Wagmi stuff, do have a custom Hook for it. This can also help us to simplify testing going forward. Because custom Hook are easier to Mock and create test cases.

src/transaction/types.ts
```diff
@@ -38,6 +38,7 @@ export type TransactionContextType = {
   setTransactionId: (id: string) => void;
   status?: string;
   transactionId?: string;
+  transactionHash?: string;
```
Zizzamia, line 41: Let's update the doc as well. For this type.

src/transaction/components/TransactionProvider.tsx
abcrane123: i combined in the `useCallsStatus` wrapper hook. let me know if you think i should separate them instead.

.changeset/lazy-shoes-work.md
```diff
@@ -0,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
+
+- **patch**: retrieve txn hash for chain explorer by: @abcrane123 #831
```
Zizzamia, line 5: this is a `feat` :-)

src/transaction/core/useCallsStatus.ts
```diff
@@ -0,0 +1,24 @@
...
+type UseCallsStatusParams = {
+  transactionId: string;
+};
+
+export function useCallsStatus({ transactionId }: UseCallsStatusParams) {
+  try {
+    const { data } = useCallsStatusWagmi({
+      id: transactionId,
+      query: {
+        refetchInterval: (data) =>
```
Zizzamia, line 12: can you write this as 

```
refetchInterval: (data) => {
  return data.state.data?.status === 'CONFIRMED' ? false : 1000,
}
```
it's just a bit more nit

Zizzamia (approved):

Zizzamia (approved):

PR #865, opened 2024-07-23
feat: added toast animation

abcrane123:
**What changed? Why?**
- adjust toast animation
- adjust toast message
- export toast icon component
**Notes to reviewers**

**How has it been tested?**


src/transaction/hooks/useGetTransactionStatus.tsx
```diff
@@ -28,8 +28,10 @@ export function useGetTransactionStatus() {
       //   </a>
```
Zizzamia, line 28: @abcrane123 can you remind me what's the story with this?

src/transaction/hooks/useGetTransactionToast.tsx
```diff
@@ -62,6 +62,10 @@ export function useGetTransactionToast() {
       icon = <Spinner className="px-1.5 py-1.5" />;
       label = 'Transaction in progress';
     }
+    if (status === 'success') {
+      icon = successSVG;
+      label = 'Successful';
```
Zizzamia, line 67: All the SVGs let's move them into `internal/svg/` folder

site/tailwind.config.js
```diff
@@ -30,6 +30,21 @@ export default {
       screens: {
         md: '848px',
       },
+      keyframes: {
```
Zizzamia, line 33: I assume this is for test purposes?

src/transaction/hooks/useGetTransactionStatus.tsx
abcrane123: `useWriteContracts` returns a transaction id that looks something like `0xf595f3f5111f88b30c668946d4430cf74326aadec59fe846…0000000000000000000000000000000000000000000014a34`. i asked connor how i can use this id to get the correct block explorer link for a pending transaction. i pass this id to `useCallsStatus` to get the txn hash which i use to get the "view transaction" block explorer link but from the designs it looks like there is a way to get the explorer link with that transaction id.

site/tailwind.config.js
abcrane123: i wasn't sure where i need to put this to get the fade in animation to work. noticed there is a top level `tailwind.config.js` and this one. let me know if i should delete it.

Zizzamia: ideally those should come from OnchainKit, no?

Zizzamia (approved):

PR #866, opened 2024-07-23
feat: Add EOA account support to Transaction Component

cpcramer:
**What changed? Why?**
Add `EOA` account support to the `Transaction Component`. 
- Add `useWriteContract` wrapper.
- To support both `Smart Wallets` and `EOA` wallets we are defaulting to use `writeContracts` first, and then if that fails with an `EOA` specific error then we fallback to use `writeContract` and iterate through the list of contracts that were intended to be batched.
- Add basic functionality test for the Transaction Provider

Screenshot 2024-07-23 at 6 44 08 PM

**Notes to reviewers**

**How has it been tested?**
Basic functionality unit test :
Screenshot 2024-07-23 at 10 34 14 PM

Tested locally on BaseSepolia. 

EOA Transaction landed onchain https://sepolia.basescan.org/tx/0x950f41d5bc0fe191b3a019d4c00f2ee295c1fb0a77f5a99d2db54334d5c3a300

### DIFF
--- a/src/transaction/components/TransactionProvider.tsx
+++ b/src/transaction/components/TransactionProvider.tsx
@@ -55,6 +55,8 @@ export function TransactionProvider({
   const { transactionHash } = useCallsStatus({ onError, transactionId });
 
   const fallbackToWriteContract = useCallback(async () => {
+    // EOAs don't support batching, so we process contracts individually.
+    // This gracefully handles accidental batching attempts with EOAs.
     for (const contract of contracts) {
       try {
         await writeContract(contract);
@@ -72,6 +74,8 @@ export function TransactionProvider({
         contracts,
       });
 
+      // EOA accounts always fail on writeContracts, returning undefined.
+      // Fallback to writeContract, which works for EOAs.
       if (result === undefined) {
         await fallbackToWriteContract();
       }
--- a/src/transaction/components/TransactionProvider.tsx
+++ b/src/transaction/components/TransactionProvider.tsx
@@ -1,7 +1,11 @@
 import { createContext, useCallback, useContext, useState } from 'react';
 import { useValue } from '../../internal/hooks/useValue';
 import { useCallsStatus } from '../hooks/useCallsStatus';
-import { useWriteContracts } from '../hooks/useWriteContracts';
+import { useWriteContract } from '../hooks/useWriteContract';
+import {
+  genericErrorMessage,
+  useWriteContracts,
+} from '../hooks/useWriteContracts';
 import type {
   TransactionContextType,
   TransactionProviderReact,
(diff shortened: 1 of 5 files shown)


src/transaction/components/TransactionProvider.tsx
```diff
@@ -39,21 +40,49 @@ export function TransactionProvider({
...
 
   const { transactionHash } = useCallsStatus({ onError, transactionId });
 
-  const handleSubmit = useCallback(() => {
+  const fallbackToWriteContract = useCallback(async () => {
+    for (const contract of contracts) {
+      try {
+        await writeContract(contract);
+      } catch (_err) {
+        setErrorMessage('genericErrorMessage');
```
cpcramer, line 66: Following the pattern in the `Transaction Component`. I believe that error handling will be addresses in a fast follow

src/transaction/components/TransactionProvider.tsx
```diff
@@ -39,21 +40,49 @@ export function TransactionProvider({
...
-    });
-  }, [contracts, writeContracts]);
+    try {
+      const result = await writeContracts({
+        contracts,
+      });
+
+      if (result === undefined) {
+        await fallbackToWriteContract();
+      }
```
cpcramer, line 81: This might seem strange. Note that when an EOA calls `writeContracts` the function does not fail. 

`result === undefined` is actually what is calling `fallbackToWriteContract`.

cpcramer: I'm going to update that catch statement to error if it gets hit.

cpcramer:
I'm pushing this PR since we have a time crunch. I will be adding tests either tonight or tomorrow. CI is failing due to low test coverage, I will address this soon.

src/transaction/hooks/useWriteContracts.ts
```diff
@@ -8,9 +8,10 @@ type UseWriteContractsParams = {
   setTransactionId: (id: string) => void;
 };
 
-const genericErrorMessage = 'Something went wrong. Please try again.';
+export const genericErrorMessage = 'Something went wrong. Please try again.';
 const uncaughtErrorCode = 'UNCAUGHT_WRITE_TRANSACTIONS_ERROR';
 const errorCode = 'WRITE_TRANSACTIONS_ERROR';
+const eoaErrorMesssage = 'this request method is not supported';
```
0xAlec, line 14: thoughts on combining all errors from all components into a single file or module called `errors`

cpcramer: + I think this would be great, especially as we align on a standardized error approach - that refactor would be better suited in a future PR

src/transaction/hooks/useWriteContracts.ts
```diff
@@ -8,10 +8,17 @@ type UseWriteContractsParams = {
...
 
+/**
+ * useWriteContracts: Experimental Wagmi hook for batching transactions.
+ * Supports Smart Wallets.
+ * Does not support batch operations and capabilities such as paymasters.
```
Zizzamia, line 19: Is this comment correct?

cpcramer: Fixed! Good call on adding comments while the info is "fresh" already getting confused haha

Zizzamia (approved):

PR #869, opened 2024-07-23
Fix: Fix the usename hook to use chain.id  in the query key to avoid overlap between chains

kirkas:
**What changed? Why?**

**fix**: fix `useName` hook to use `chain.id` in the query key to avoid overlap between chains

### DIFF
--- a/src/identity/hooks/useName.ts
+++ b/src/identity/hooks/useName.ts
@@ -17,7 +17,8 @@ export const useName = (
   queryOptions?: UseNameQueryOptions,
 ) => {
   const { enabled = true, cacheTime } = queryOptions ?? {};
-  const ensActionKey = `ens-name-${address}`;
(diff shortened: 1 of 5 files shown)


src/identity/hooks/useName.ts
```diff
@@ -17,7 +17,8 @@ export const useName = (
   queryOptions?: UseNameQueryOptions,
 ) => {
   const { enabled = true, cacheTime } = queryOptions ?? {};
-  const ensActionKey = `ens-name-${address}`;
+  const chainActionKey = chain ? chain.id : 'default-chain';
```
Zizzamia, line 20: I will say let's just have the default value the ETH chain ID. Which I believe is 0, no?

kirkas: it's 1, I'll set the default chain to mainnet like we do for the getName function!

Zizzamia (approved):

PR #878, opened 2024-07-24
feat: `v2` Swap API

0xAlec:
**What changed? Why?**
- add `aggregators` prop to `<Swap/>` to control `V1` or `V2` API
- this will add the `v2Enabled` parameter in the request body, our API will route these to a beta implementation for serving quotes/trades directly from Uniswap V3

`V1` API does not index pools with under <= $10,000 of liquidity, `V2` API will make a direct call to the `UniswapRouter` contract

**Notes to reviewers**
`slippage` for the `V2` API is hardcoded to 1000 bps (10%) - Tina is aligned for the initial launch and we'll fast-follow to add configurability to the UI

usage
```javascript
<Swap address={address} experimental={{ useAggregator: false }}>
          ...
</Swap>
```

**How has it been tested?**
See Slack for demo

### DIFF
--- a/src/swap/components/SwapProvider.tsx
+++ b/src/swap/components/SwapProvider.tsx
@@ -35,10 +35,12 @@ export function useSwapContext() {
 
 export function SwapProvider({
   address,
+  aggregator,
   children,
 }: {
-  children: React.ReactNode;
   address: Address;
+  aggregator: boolean;
+  children: React.ReactNode;
 }) {
   const [loading, setLoading] = useState(false);
   const [isTransactionPending, setPendingTransaction] = useState(false);
@@ -99,6 +101,7 @@ export function SwapProvider({
 
       try {
         const response = await getSwapQuote({
+          aggregator,
           from: source.token,
           to: destination.token,
           amount,
@@ -141,6 +144,7 @@ export function SwapProvider({
 
       try {
         const response = await buildSwapTransaction({
+          aggregator,
           amount: from.amount,
           fromAddress: address,
           from: from.token,
--- a/src/swap/components/SwapProvider.tsx
+++ b/src/swap/components/SwapProvider.tsx
@@ -126,7 +126,7 @@ export function SwapProvider({
         destination.setLoading(false);
       }
     },
-    [from, to, handleError],
+    [aggregator, from, to, handleError],
   );
 
   const handleSubmit = useCallback(
@@ -188,6 +188,7 @@ export function SwapProvider({
     },
     [
       address,
+      aggregator,
       config,
       handleError,
       from.amount,
--- a/src/swap/components/SwapProvider.tsx
+++ b/src/swap/components/SwapProvider.tsx
@@ -36,12 +36,17 @@ export function useSwapContext() {
 export function SwapProvider({
   address,
   children,
-  useAggregator,
+  experimental,
 }: {
   address: Address;
   children: React.ReactNode;
-  useAggregator: boolean;
+  experimental: {
+    useAggregator: boolean; // Whether to use a DEX aggregator. (default: true)
+  };
 }) {
+  // Feature flags
+  const { useAggregator } = experimental;
+
   const [loading, setLoading] = useState(false);
   const [isTransactionPending, setPendingTransaction] = useState(false);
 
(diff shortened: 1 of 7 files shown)


src/swap/types.ts
```diff
@@ -57,6 +58,7 @@ export type GetSwapAPIParams = GetQuoteAPIParams & {
 export type GetSwapQuoteParams = {
   from: Token; // The source token for the swap
   to: Token; // The destination token for the swap
+  aggregator: boolean; // Whether to use a DEX aggregator
```
Zizzamia, line 61: alphabetical

src/swap/types.ts
```diff
@@ -45,6 +45,7 @@ export type GetQuoteAPIParams = {
   to: AddressOrETH | ''; // The destination address or 'ETH' for Ethereum
   amount: string; // The amount to be swapped
   amountReference?: string; // The reference amount for the swap
+  v2Enabled?: boolean; // Whether to use V2 of the API (default: false)
```
Zizzamia, line 48: why not just enable it as default? After all it should not have a real change for the experience, right?

src/swap/utils/buildSwapTransaction.ts
```diff
@@ -23,7 +23,14 @@ export async function buildSwapTransaction(
     isAmountInDecimals: false,
   };
 
-  const apiParams = getAPIParamsForToken({ ...defaultParams, ...params });
+  let apiParams = getAPIParamsForToken({ ...defaultParams, ...params });
+
+  if (!params.aggregator) {
+    apiParams = {
+      v2Enabled: true,
```
Zizzamia, line 30: oh I see you want to test this on his own for a bit?

src/swap/types.ts
0xAlec: fixed

src/swap/types.ts
0xAlec: see Slack

src/swap/utils/buildSwapTransaction.ts
0xAlec: yeah, we'll do a switchover once the V2 service is more production ready

src/swap/components/SwapProvider.tsx
```diff
@@ -36,9 +36,11 @@ export function useSwapContext() {
 export function SwapProvider({
   address,
   children,
+  useAggregator,
```
Zizzamia, line 39: the provider too should have `experimental`

Zizzamia: we don't want to leak bad experience across too many components, and we want to keep `experimental` nice and neat.

src/swap/types.ts
Zizzamia: use `experimental`

src/swap/types.ts
```diff
@@ -55,11 +56,12 @@ export type GetSwapAPIParams = GetQuoteAPIParams & {
...
  */
 export type GetSwapQuoteParams = {
-  from: Token; // The source token for the swap
-  to: Token; // The destination token for the swap
   amount: string; // The amount to be swapped
   amountReference?: string; // The reference amount for the swap
+  from: Token; // The source token for the swap
   isAmountInDecimals?: boolean; // Whether the amount is in decimals
+  to: Token; // The destination token for the swap
+  useAggregator: boolean; // Whether to use a DEX aggregator
```
Zizzamia, line 64: use experimental

src/swap/components/SwapProvider.tsx
```diff
@@ -36,10 +36,17 @@ export function useSwapContext() {
...
+  experimental,
 }: {
-  children: React.ReactNode;
   address: Address;
+  children: React.ReactNode;
+  experimental: {
+    useAggregator: boolean; // Whether to use a DEX aggregator. (default: true)
+  };
 }) {
+  // Feature flags
```
Zizzamia, line 47: Love the comment

src/swap/types.ts
Zizzamia: `v2Enabled` this is such a bad way to name a property for an API.

Zizzamia: do we really need this year?

Zizzamia (approved):

PR #887, opened 2024-07-24
feat: BaseSepolia Network Switcher 

cpcramer:
**What changed? Why?**
Add `"Switch to Base Sepolia"` button above the Transaction Button in `/transaction/transaction` 
- The button will only display if the users EOA wallet is connect and if they are not on Base Sepolia. 

This is needed because as of today the Transaction Button will error if an EOA account user is on any network other than `Base Sepolia`.

We will eventually use this component throughout OnchainKit.xyz - potentially as a nav bar drop down or prompted in the wallet popup. 

Screenshot 2024-07-24 at 5 15 09 PM

[BOE-85](https://linear.app/coinbase/issue/BOE-85/add-network-switching-functionality-to-onchainkitxyz-that-default-to)
**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/site/docs/components/switchNetworkBaseSepolia.tsx
+++ b/site/docs/components/switchNetworkBaseSepolia.tsx
@@ -2,38 +2,38 @@ import { useSwitchChain, useAccount } from "wagmi";
 import { baseSepolia } from "wagmi/chains";
 import { useState, ReactNode } from "react";
 
-type SwitchTargetBaseSepoliaProps = {
+type SwitchNetworkBaseSepoliaProps = {
   children: (props: {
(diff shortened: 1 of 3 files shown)


Zizzamia:
I like the direction, but I am not sure it's the right UX. Have you chat with Mind on this?

cpcramer:
> I like the direction, but I am not sure it's the right UX. Have you chat with Mind on this?

Agreed! And probably going to table this for a few days. But we likely can use this type of functionality to force/guide the user to the Base network. I'll need to work with Mind about the type of UX we want to have

PR #893, opened 2024-07-25
chore: Transaction Component Add Paymaster Support + QA cleanup

ilikesymmetry:
**What changed? Why?**

- Add `SmartWalletCapabilities` paymaster sponsorship support in the `Transaction Component`. 
- Add new `click` contract for testing.
- Remove `AppBaseSepolia`. 
- Add `chainId` parameter to the `Transaction Component`.
- Fix the bug where we would see `User rejected the request` error despite the transaction succeeding. 
- Auto switch the users wallet to Base Sepolia when interacting with the transaction component.
- Fix bug where successful testnet transaction's "View Transaction" button links to Base Mainnet explorer.
- General clean up.

**Notes to reviewers**

**How has it been tested?**


src/transaction/components/Transaction.tsx
```diff
@@ -7,11 +7,15 @@ export function Transaction({
   className,
   children,
   contracts,
+  chainId,
+  capabilities,
```
Zizzamia, line 11: Keep alphabetic order. 

Btw in other places we have `chain` and not `chainId`, also why this should override the chain set at OnchainProvider level?

src/transaction/hooks/useWriteContracts.ts
```diff
@@ -1,5 +1,6 @@
 import type { TransactionExecutionError } from 'viem';
 import { useWriteContracts as useWriteContractsWagmi } from 'wagmi/experimental';
+import { METHOD_NOT_SUPPORTED_ERROR_SUBSTRING } from '../constants';
```
Zizzamia, line 3: Thank you for starting a `constants` file.

src/transaction/components/Transaction.tsx
cpcramer: Updated to alphabetic

src/transaction/components/Transaction.tsx
```diff
@@ -12,6 +14,8 @@ export function Transaction({
   return (
     <TransactionProvider
       address={address}
+      capabilities={capabilities}
+      chainId={chainId}
```
Zizzamia, line 18: To stay in line with the other components API, this should be
```ts
chain?: Chain;
```

src/transaction/components/Transaction.tsx
```diff
@@ -4,6 +4,8 @@ import { TransactionProvider } from './TransactionProvider';
 
 export function Transaction({
   address,
+  capabilities,
+  chainId,
```
Zizzamia, line 8: To stay in line with the other components API, this should be
```ts
chain?: Chain;
```

src/transaction/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const SEND_CALLS_NOT_SUPPORTED_ERROR = 'SEND_CALLS_NOT_SUPPORTED_ERROR';
+export const METHOD_NOT_SUPPORTED_ERROR_SUBSTRING =
+  'this request method is not supported'; // most likely EOA
```
Zizzamia, line 3: If this message is displayed, we probably want the `t` upper case. as `This request`

src/transaction/components/TransactionProvider.tsx
```diff
@@ -69,26 +69,54 @@ export function TransactionProvider({
     }
   }, [contracts, writeContract]);
 
+  const switchChain = useCallback(
```
Zizzamia, line 72: To keep this code easier to mantain and test, we should probably split those into indipendent custom Hooks.

Let me know how much this is douable or not.

site/docs/pages/transaction/transaction.mdx
```diff
@@ -77,22 +78,31 @@ export default function TransactionComponents() {
...
+            address={address}
+            capabilities={{
+              paymasterService: {
+                url: 'https://api.developer.coinbase.com/rpc/v1/base-sepolia/KkgONfLFgUZKI8nnrttLbI78U2vyhtXg',
+              },
+            }}
+            chainId={84532}
+            contracts={contracts}
+            onError={onError}
+          >
```
ilikesymmetry, line 104: @fakepixels sample use of new `Transaction` component

Zizzamia (approved):
Approving this as it is, and than do a bunch of fast-follow to correct what's breaking.

PR #895, opened 2024-07-25
[fix]: Correctly disconnect the wallet(s)

kirkas:
**What changed? Why?**
**fix**: Fix `WalletDropdownDisconnect` to correclty disconnect all connectors

### DIFF
--- a/src/wallet/components/WalletDropdownDisconnect.test.tsx
+++ b/src/wallet/components/WalletDropdownDisconnect.test.tsx
@@ -1,23 +1,23 @@
 import { fireEvent, render, screen } from '@testing-library/react';
 import { describe, expect, it, vi } from 'vitest';
-import { useConnect, useDisconnect } from 'wagmi';
+import { type Connector, useDisconnect } from 'wagmi';
 import { WalletDropdownDisconnect } from './WalletDropdownDisconnect';
 
 vi.mock('wagmi', () => ({
-  useConnect: vi.fn(),
   useDisconnect: vi.fn(),
 }));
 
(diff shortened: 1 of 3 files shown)


src/wallet/components/WalletDropdownDisconnect.test.tsx
```diff
@@ -1,23 +1,23 @@
 import { fireEvent, render, screen } from '@testing-library/react';
 import { describe, expect, it, vi } from 'vitest';
-import { useConnect, useDisconnect } from 'wagmi';
+import { type Connector, useDisconnect } from 'wagmi';
```
Zizzamia, line 3: nit, let's have `types` always on a different line of import so it will be
```
import { useDisconnect } from 'wagmi';
import type { Connector } from 'wagmi';
```

src/wallet/components/WalletDropdownDisconnect.tsx
```diff
@@ -8,13 +8,10 @@ export function WalletDropdownDisconnect({
...
 }: WalletDropdownDisconnectReact) {
-  const { connectors } = useConnect();
-  const { disconnect } = useDisconnect();
-  const connector = connectors[0];
-
+  const { disconnect, connectors } = useDisconnect();
   const handleDisconnect = useCallback(() => {
-    disconnect({ connector });
-  }, [connector, disconnect]);
+    connectors.map((connector) => disconnect({ connector }));
```
Zizzamia, line 13: Oh nice, can we add a line a comment that explain the why we do this.
It will help future developers in mantaining this.

Zizzamia (approved):

PR #910, opened 2024-07-26
fix: address reload correctly when user switch wallet account

kirkas:
**What changed? Why?**
**fix**: Fix `IdentityProvider` to use address directly when switching address.

**Notes to reviewers**

**How has it been tested?**


Zizzamia:
@kirkas could you rebase from main when you have a minute? With all our tests in order, it'll help our CI pass more easily.

Zizzamia (approved):

PR #911, opened 2024-07-26
docs: Add transaction anatomy

cpcramer:
**What changed? Why?**
Screenshot 2024-07-26 at 11 16 55 AM

**Notes to reviewers**

**How has it been tested?**


site/docs/pages/transaction/transaction.mdx
```diff
@@ -169,6 +169,11 @@ export default function TransactionComponents() {
 </App>
 
 ## Components
+
+<div className="flex flex-col max-w-[648px] gap-6">
+  <img src="/assets/onchainkit-transaction-anatomy.png" />
```
Zizzamia, line 174: Love that you took the time to have a good filename. One more way you can push the limits for SEO, is by asking yourself what would be the words the developer we search. And therefor it will end up this image to appear.

One more way this can be polish, is by adding the word `components` or `component`, so that you really hit the jackpot.

so it would be `onchainkit-components-transaction-anatomy`.

Zizzamia (approved):
Merging as it is, feel free to change the file name as follow up, and also add SEO tag to the image.

site/docs/pages/transaction/transaction.mdx
Zizzamia: Oh also, it's missing the SEO tags for the image.

Zizzamia: Rember to always check for mobile and desktop for production pages how they play out in here https://pagespeed.web.dev/, that site will teach you a lot on great SEO.

cpcramer: Interesting, I never think about SEO! Will fast follow now

cpcramer: [Improve Transaction Component Anatomy SEO](https://github.com/coinbase/onchainkit/pull/927#top)

PR #912, opened 2024-07-26
fix: `formatDecimals` precision

0xAlec:
**What changed? Why?**
- previously we cast `amount` from a string to a javascript number (max 64 bits), which would lose precision for some ERC-20 denominations
- this would result in a trade being requested for more token units than the user had in their account, which resulted in unexecutable transactions
- change to using `BigInt` implementation

**Notes to reviewers**

**How has it been tested?**
image
improved test coverage compared to `main`

unit tested and tested locally

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "clsx": "^2.1.1",
     "graphql": "^14 || ^15 || ^16",
     "graphql-request": "^6.1.0",
-    "jsbi": "^4.3.0",
     "permissionless": "^0.1.29",
     "tailwind-merge": "^2.3.0",
     "viem": "^2.17.4",
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "clsx": "^2.1.1",
     "graphql": "^14 || ^15 || ^16",
     "graphql-request": "^6.1.0",
+    "jsbi": "^4.3.0",
     "permissionless": "^0.1.29",
     "tailwind-merge": "^2.3.0",
     "viem": "^2.17.4",
--- a/src/swap/utils/formatDecimals.ts
+++ b/src/swap/utils/formatDecimals.ts
@@ -15,8 +15,11 @@ export function formatDecimals(
   if (!Number.isInteger(decimals) || decimals < 0) {
     throw new Error('Invalid input: decimals must be a non-negative integer');
   }
-  if (!/^-?\d*\.?\d+$/.test(amount)) {
+  if (!/^-?(?:0|[1-9]\d*)(?:\.\d+)?$/.test(amount)) {
     throw new Error('Invalid input: amount must be a valid number string');
+  }
+  if (amount.startsWith('-')) {
+    throw new Error('Invalid input: negative numbers are not allowed');
   }
 
   let result: string;
--- a/src/swap/utils/formatDecimals.ts
+++ b/src/swap/utils/formatDecimals.ts
@@ -1,4 +1,3 @@
-import type { SwapError } from '../types';
 import { fromReadableAmount } from './fromReadableAmount';
 import { toReadableAmount } from './toReadableAmount';
 
@@ -9,27 +8,7 @@ export function formatDecimals(
   amount: string,
   inputInDecimals = true,
   decimals = 18,
-): string | SwapError {
-  // Input validation
-  if (typeof amount !== 'string' || amount.trim() === '') {
-    return {
-      code: 'INVALID_INPUT',
-      error: 'Invalid input: amount must be a non-empty string',
-    };
-  }
-  if (!Number.isInteger(decimals) || decimals < 0) {
-    return {
-      code: 'INVALID_INPUT',
-      error: 'Invalid input: decimals must be a non-negative integer',
-    };
-  }
-  if (!/^(?:0|[1-9]\d*)(?:\.\d+)?$/.test(amount)) {
-    return {
-      code: 'INVALID_INPUT',
-      error: 'Invalid input: amount must be a number',
-    };
-  }
-
+): string {
   let result: string;
 
   if (inputInDecimals) {
--- a/src/swap/utils/formatDecimals.ts
+++ b/src/swap/utils/formatDecimals.ts
@@ -1,4 +1,5 @@
-import JSBI from 'jsbi';
+import { fromReadableAmount } from './fromReadableAmount';
+import { toReadableAmount } from './toReadableAmount';
 
 /**
  * Formats an amount according to the decimals. Defaults to 18 decimals for ERC-20s.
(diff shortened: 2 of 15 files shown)


src/swap/utils/formatDecimals.ts
```diff
@@ -6,8 +8,55 @@ export function formatDecimals(
...
+  }
+  if (!Number.isInteger(decimals) || decimals < 0) {
+    throw new Error('Invalid input: decimals must be a non-negative integer');
+  }
+  if (!/^-?(?:0|[1-9]\d*)(?:\.\d+)?$/.test(amount)) {
+    throw new Error('Invalid input: amount must be a valid number string');
+  }
+  if (amount.startsWith('-')) {
+    throw new Error('Invalid input: negative numbers are not allowed');
+  }
```
cpcramer, line 23: Do we need 2 checks for negative integers?

src/swap/utils/formatDecimals.ts
```diff
@@ -6,8 +8,55 @@ export function formatDecimals(
...
+
+  let result: string;
+
   if (inputInDecimals) {
-    return (Number(amount) / 10 ** decimals).toString();
+    // If input is already in decimals, convert to readable amount
```
cpcramer, line 28: What do we mean by "readable" amount?

cpcramer (approved):

src/swap/utils/formatDecimals.ts
0xAlec: updated the regex to check for negatives

Zizzamia:

package.json
```diff
@@ -38,6 +38,7 @@
     "clsx": "^2.1.1",
     "graphql": "^14 || ^15 || ^16",
     "graphql-request": "^6.1.0",
+    "jsbi": "^4.3.0",
```
Zizzamia, line 41: Very intersting, i believe as well a library of this kind can help us. But before we merge it, @0xAlec can you take me into the trade-off in adding this libraries. Usual things like:
- how many Kb the library is
- what's the state of being manteined
- what risk are we accepting to take by adding this library
- and anything else?

src/swap/utils/formatDecimals.ts
```diff
@@ -1,3 +1,5 @@
+import JSBI from 'jsbi';
```
Zizzamia, line 1: Direct imports are strictly not advised as are impossible to tree-shake. This is something the Javascript community  moved away from in 2016, after the big drama libraries like `lodash` created.

I did not checked the library directly, but is it possible to just import what we need, or we always have to get everything?

Importing always everything is quite a risk for us to take, as open source library.

src/swap/utils/formatDecimals.ts
```diff
@@ -6,8 +8,54 @@ export function formatDecimals(
...
+  // Input validation
+  if (typeof amount !== 'string' || amount.trim() === '') {
+    throw new Error('Invalid input: amount must be a non-empty string');
+  }
+  if (!Number.isInteger(decimals) || decimals < 0) {
+    throw new Error('Invalid input: decimals must be a non-negative integer');
+  }
+  if (!/^(?:0|[1-9]\d*)(?:\.\d+)?$/.test(amount)) {
+    throw new Error(
+      'Invalid input: amount must be a non-negative number string',
```
Zizzamia, line 20: Love the care for the messages.

src/swap/utils/formatDecimals.ts
```diff
@@ -6,8 +8,54 @@ export function formatDecimals(
   inputInDecimals = true,
   decimals = 18,
 ): string {
```
Zizzamia, line 10: Is `throw` the correct way to approach this? And the type `string` can be missleading when returing an error.

I am leaning towards stick with our Error standard, we use in other utils and componets. So we stay predictable.

I believe we decided to `throw` only when there are must cases at config the developer has to adjust before using the library.

src/swap/utils/formatDecimals.ts
```diff
@@ -6,8 +8,54 @@ export function formatDecimals(
...
+  } else {
+    // If input is not in decimals, convert from readable amount
+    result = fromReadableAmount(amount, decimals);
   }
-  return (Number(amount) * 10 ** decimals).toString();
+
+  return result;
+}
+
+export function fromReadableAmount(amount: string, decimals: number): string {
```
Zizzamia, line 37: We want every function to be his own file. This is something we mentioned in the past a few times, as it is a more verbose way to create a library, but in the long run makes it easier to mantain and test.

src/swap/utils/formatDecimals.ts
0xAlec: this should be abstracted from the developer. `formatDecimals` is called by `getAPIParamsForToken` which is called by `buildSwapTransaction`. I wrapped the part in `buildSwapTransaction` in a try/catch and throw a `SwapError` if `formatDecimals` fails.

we will need to do a follow-up PR to handle invalid inputs in the FE component.

src/swap/utils/formatDecimals.ts
0xAlec: not the expert here - but seems like this is a known (unresolved) issue https://github.com/GoogleChromeLabs/jsbi/issues/22

however, performance wise it should be on par with native BigInt implementations from newer versions of browsers

```
On average, JSBI is performance-competitive with the native implementation that Google Chrome is currently shipping.
```

package.json
0xAlec: - 568 kB unpacked size
- maintained by Google Chrome team
- not sure, we can use alternatively use native BigInts at expense of compatibility for older browsers

Zizzamia: `568 kB` is quite a lot, I wonder how much will be minified.

src/swap/utils/formatDecimals.ts
Zizzamia: Should we just use the native when detected? And default to do nothing when is not available?

It seems the library is pretty heavy

src/swap/utils/formatDecimals.ts
Zizzamia: An utility like `A` should always be the most pure as possible, with characteristics that makes it useful on his own. Expecting `A` doing the right thing when `B` calls `A` it creates mental complexity when in the long-run mantaining any complex library.

In a few words, we want to avoid what commonly is called "spaghetti code" which the definision says
```
 Spaghetti code is the general term used for any source code that's hard to understand because it has no defined structure. While an end user might not see anything wrong with a program, a programmer might find it virtually illegible if the code base's flow is too convoluted—like a bowl of twisted, tangled spaghetti.
```

That's why it's important to avoid at the origin this kind of slippery slope in our code.

src/swap/utils/formatDecimals.ts
0xAlec: migrated to native `BigInt`

src/swap/utils/formatDecimals.ts
0xAlec: returned `SwapError` and moved the input validation a level up

Zizzamia (approved):
So niiiice

PR #914, opened 2024-07-27
feat: add `onStart` hook to `SwapButton`

0xAlec:
**What changed? Why?**
add `onStart` hook to `SwapButton`, which will be called with the `txHash` (tx receipt has not been awaited yet). 

for swaps from ERC-20s, the `onStart` hook will be ran twice (one for the approval transaction, one for the actual transaction)

this is to expose more transaction state in order to enable use cases like custom toast messages

**Notes to reviewers**

**How has it been tested?**
locally and in unit tests

```
            <SwapButton
              onStatus={{
                onStart: async (txHash) => {
                  await new Promise((resolve) => setTimeout(resolve, 1000));
                  console.log('async onStart:', txHash);
                },
              }}
            />
```

### DIFF
--- a/src/swap/components/SwapButton.tsx
+++ b/src/swap/components/SwapButton.tsx
@@ -7,6 +7,7 @@ export function SwapButton({
   className,
   disabled = false,
   onError,
+  onStart,
   onSuccess,
 }: SwapButtonReact) {
   const { to, from, loading, isTransactionPending, handleSubmit } =
@@ -34,7 +35,7 @@ export function SwapButton({
         text.headline,
         className,
       )}
-      onClick={() => handleSubmit(onError, onSuccess)}
+      onClick={() => handleSubmit(onError, onStart, onSuccess)}
       disabled={isDisabled}
       data-testid="ockSwapButton_Button"
     >
--- a/src/swap/components/SwapButton.tsx
+++ b/src/swap/components/SwapButton.tsx
@@ -7,8 +7,8 @@ export function SwapButton({
   className,
   disabled = false,
   onError,
+  onStart,
   onSuccess,
-  onStatus,
 }: SwapButtonReact) {
   const { to, from, loading, isTransactionPending, handleSubmit } =
     useSwapContext();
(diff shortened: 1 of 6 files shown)


Zizzamia:

src/swap/components/SwapButton.tsx
```diff
@@ -7,6 +7,7 @@ export function SwapButton({
   className,
   disabled = false,
   onError,
+  onStart,
```
Zizzamia, line 10: We are moving more and more into the land of creating a component life-cycle.

Those are big decision to take, so let's take a minute to reflect on every step we are taking.

Do we see `onStart` being used in other components as well? And overall what kind of expecation we want to train the developer when they see `onStart` into a component?

0xAlec: this is a specific customer request from PartyDAO to manage the component life-cycle, so they can implement their own toasts

regardless, even if we ship TransactionKit, we should expose hooks that allow the developer to customize the transaction states (in this instance, removing our `SwapMessage` and using their own)

0xAlec: since we already have `onSuccess` (which imo is pretty intuitive since the developer may want to make a callback with the resulting swap transaction hash), I think the customer request for an `onStart` is reasonable as well

this specific use case they want to add their own toast, but I can imagine it being used for other reasons as well

Zizzamia: > I'm of the opinion we should expose hooks that allow the developer to customize the transaction states

I fully agree on this, in fact if you remember I mentioned a few times we should have a `onStatus` hook, to give that level of customization.

Do we think `onStatus` can satisfay `onStart` and future lifecycle states? What I want to avoid is creating 3 more custom listeners, but instead have one that listens all states and get give great inisights on what's going on.

Ideally `onStatus` could deprecate `onSuccess` at some point.

0xAlec: added `onStatus` prop

Zizzamia (approved):

PR #915, opened 2024-07-27
feat: refetch balances and clear input amounts post-swap

0xAlec:
**What changed? Why?**
- clear the input amounts
- refetch balances post swap

for a more polished user experience

**Notes to reviewers**
N/A

**How has it been tested?**
locally


Zizzamia:

src/swap/components/SwapProvider.tsx
```diff
@@ -189,15 +187,21 @@ export function SwapProvider({
         }
       } finally {
         setLoading(false);
+        from.setAmount('');
```
Zizzamia, line 190: Do we have tests to cover this case scenario?

src/swap/hooks/useSwapBalances.tsx
```diff
@@ -48,5 +57,13 @@ export function useSwapBalances({
 
     toBalanceString,
     toTokenBalanceError,
+
+    refetchBalances: async () => {
```
Zizzamia, line 61: I believe we miss tests for this.

Zizzamia: Btw, I don't think this is a clean way to write this. We are creating a callback inside an object, and than we are sending this callback everywhere.

This approach makes debugging this very complex. We need to rewrite this.

We need a way to start the `refetch` when at Provider level by listening to a particular case.

PR #944, opened 2024-07-30
continued Transaction component QA

abcrane123:
**What changed? Why?**
- add `delayMs` prop to transaction toast
- update button text on error
- remove sponsored by name
- update to use `writeContractAsync` instead of `writeContract`

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/transaction/components/TransactionSponsor.tsx
+++ b/src/transaction/components/TransactionSponsor.tsx
@@ -2,10 +2,7 @@ import { cn, color, text } from '../../styles/theme';
 import type { TransactionSponsorReact } from '../types';
 import { useTransactionContext } from './TransactionProvider';
 
-export function TransactionSponsor({
-  className,
-  text: sponsorName,
-}: TransactionSponsorReact) {
+export function TransactionSponsor({ className }: TransactionSponsorReact) {
   const {
     errorMessage,
     hasPaymaster,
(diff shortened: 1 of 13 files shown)


src/transaction/components/TransactionSponsor.tsx
```diff
@@ -30,18 +27,7 @@ export function TransactionSponsor({
...
-          <p className={cn(color.foregroundMuted, 'pl-2')}>•</p>
-          <p className={cn(color.foregroundMuted, 'pl-2')}>
-            Sponsored by{' '}
-            <span className={cn(text.label1, color.primary)}>
-              {sponsorName}
-            </span>
-          </p>
-        </div>
-      )}
+      <p className={color.foregroundMuted}>Zero transaction fee</p>
```
Zizzamia, line 30: We can still have the `Text` handle this content. In case someone wants to customize it.

src/transaction/types.ts
```diff
@@ -123,6 +122,7 @@ export type TransactionStatusLabelReact = {
 export type TransactionToastReact = {
   children: ReactNode; // The child components to be rendered within the toast component.
   className?: string; // An optional CSS class name for styling the toast component.
+  delayMs?: number; // An optional value to customize time until toast disappears
```
Zizzamia, line 125: I believe we said to do `durationMs` for this one.

Zizzamia (approved):
Making an expection here, and approving + merging without test coverage increased and asking for the notes to be done as follow up.

PR #945, opened 2024-07-30
Update txn success handler to pass multiple receipts for multiple contracts (EOA)

abcrane123:
**What changed? Why?**
- implement `waitForTransactionReceipt` to handle case where user passes multiple contracts and is using EOA
- this function will be called for each transaction hash so that we can pass multiple receipts in onsuccess handler

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/swap/components/Swap.test.tsx
+++ b/src/swap/components/Swap.test.tsx
@@ -1,0 +1,26 @@
+import { render, screen } from '@testing-library/react';
+import { describe, it, vi } from 'vitest';
+import { Swap } from './Swap';
+
+vi.mock('./SwapProvider', () => ({
+  SwapProvider: ({ children }) => (
+    <div data-testid="mock-SwapProvider">{children}</div>
+  ),
+  useSwapContext: vi.fn(),
+}));
+
+describe('Swap Component', () => {
+  it('should render the title correctly', () => {
+    render(<Swap title="Test Swap" />);
+
+    const title = screen.getByTestId('ockSwap_Title');
+    expect(title).toHaveTextContent('Test Swap');
+  });
+
+  it('should pass className to container div', () => {
+    render(<Swap className="custom-class" />);
+
+    const container = screen.getByTestId('ockSwap_Container');
+    expect(container).toHaveClass('custom-class');
+  });
+});
--- a/src/transaction/components/TransactionProvider.tsx
+++ b/src/transaction/components/TransactionProvider.tsx
@@ -5,12 +5,13 @@ import {
   useEffect,
   useState,
 } from 'react';
-import type { TransactionExecutionError } from 'viem';
-import {
-  useAccount,
-  useSwitchChain,
-  useWaitForTransactionReceipt,
-} from 'wagmi';
+import type {
+  Address,
+  TransactionExecutionError,
+  TransactionReceipt,
+} from 'viem';
+import { useAccount, useConfig, useSwitchChain, useWaitForTransactionReceipt } from 'wagmi';
+import { waitForTransactionReceipt } from 'wagmi/actions';
 import { useValue } from '../../internal/hooks/useValue';
 import {
   GENERIC_ERROR_MESSAGE,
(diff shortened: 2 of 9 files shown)


src/transaction/components/TransactionProvider.tsx
```diff
@@ -73,17 +84,48 @@ export function TransactionProvider({
     hash: writeContractTransactionHash || transactionHash,
   });
 
+  const getTransactionReceipts = useCallback(async () => {
+    const receipts = [];
+    for (const hash of transactionHashArray) {
+      try {
+        const txnReceipt = await waitForTransactionReceipt(config, {
```
wilsoncusack, line 91: this generalizes for both wallet_getCallsStatus and eth_getTransactionReceipt right?

src/transaction/components/TransactionProvider.tsx
```diff
@@ -73,17 +84,48 @@ export function TransactionProvider({
...
+      } catch (err) {
+        console.error('getTransactionReceiptsError', err);
+        setErrorMessage(genericErrorMessage);
+      }
+    }
+    setReceiptArray(receipts);
+  }, [transactionHashArray]);
+
+  useEffect(() => {
+    if (transactionHashArray.length === contracts.length && contracts?.length > 1) {
```
wilsoncusack, line 105: why contracts.length > 1?

src/transaction/components/TransactionProvider.tsx
```diff
@@ -73,17 +84,48 @@ export function TransactionProvider({
...
+          'UserRejectedRequestError'
+        ) {
+          setErrorMessage('Request denied.');
+        } else {
+          setErrorMessage(genericErrorMessage);
+        }
       }
     }
-  }, [contracts, writeContract]);
+  }, [contracts, writeContractAsync]);
```
wilsoncusack, line 128: please make sure to test this on mobile safari. We've had some issues with them blocking our popup with this async call

src/transaction/components/TransactionProvider.tsx
abcrane123: if a user only passes one contract and is using an EOA i felt like it was better to return just the one receipt in the onSuccess instead of an array with that one receipt inside of it.

src/transaction/components/TransactionProvider.tsx
abcrane123: @Zizzamia how would you recommend we test this on mobile safari ?

src/transaction/components/TransactionProvider.tsx
abcrane123: this i'm not sure. i'm looking through the wagmi docs and haven't found any info yet.

src/swap/components/Swap.test.tsx
```diff
@@ -0,0 +1,26 @@
+import { render, screen } from '@testing-library/react';
```
abcrane123, line 1: i can pull these out into a separate PR. just needed to increase test coverage.

src/transaction/types.ts
```diff
@@ -62,7 +62,7 @@ export type TransactionProviderReact = {
   children: ReactNode; // The child components to be rendered within the provider component.
   contracts: ContractFunctionParameters[]; // An array of contract function parameters provided to the child components.
   onError?: (e: TransactionError) => void; // An optional callback function that handles errors within the provider.
-  onSuccess?: (response: TransactionResponse) => void; // An optional callback function that exposes transaction hash
+  onSuccess?: (response: TransactionReceipt | TransactionReceipt[]) => void; // An optional callback function that exposes transaction hash
```
Zizzamia, line 65: A few questions:
1. why two types? I think we should lean towards `TransactionReceipt[]`.
2. the return value from `onSuccess` should probably be an object. And have `transactionReceipts` as part of it. Just because it will keep the door open for potential future things.

abcrane123: i made it so that if there is only one `transactionReceipt` (i.e for smart wallet) we would return just the receipt instead of an array with one receipt inside but agree that your approach would simplify things. updating now!

Zizzamia: thank you

Zizzamia (approved):

PR #946, opened 2024-07-30
feat: add configurable `slippage` for V2 Swap API

0xAlec:
**What changed? Why?**
- add feature flag in `experimental` to use customizable `slippage` parameter for swaps

**Notes to reviewers**

**How has it been tested?**
locally

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -86,7 +86,8 @@
       "esm",
       "framegear",
       "playground",
-      "site"
+      "site",
+      "src/**/*.test.tsx"
     ]
   }
 }
--- a/biome.json
+++ b/biome.json
@@ -86,8 +86,7 @@
       "esm",
       "framegear",
       "playground",
-      "site",
-      "src/**/*.test.tsx"
+      "site"
     ]
   }
 }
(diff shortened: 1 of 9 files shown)


src/swap/components/SwapProvider.tsx
```diff
@@ -42,6 +42,7 @@ export function SwapProvider({
   children: React.ReactNode;
   experimental: {
     useAggregator: boolean; // Whether to use a DEX aggregator. (default: true)
+    slippage?: string; // Maximum acceptable slippage for a swap. (default: 10) This is as a percent, not basis points
```
Zizzamia, line 45: Based on your comment, should this be called `maxSlippage` and being a `number` instead of a string?

biome.json
```diff
@@ -86,7 +86,8 @@
       "esm",
       "framegear",
       "playground",
-      "site"
+      "site",
+      "src/**/*.test.tsx"
```
Zizzamia, line 90: We want our test files be nice an clean as well. Please remove this line.
And if there are exception you want to make in some test file, do it at line level.

Or if there is a particular file that is super complex, we can talk about it case by case.

src/swap/types.ts
```diff
@@ -201,6 +203,7 @@ export type SwapReact = {
   className?: string; // Optional className override for top div element.
   experimental?: {
     useAggregator: boolean; // Whether to use a DEX aggregator. (default: true)
+    slippage?: string; // Maximum acceptable slippage for a swap. (default: 10) This is as a percent, not basis points
```
Zizzamia, line 206: Should this be `maxSlippage` as well?

Zizzamia (approved):

PR #966, opened 2024-07-31
chore: Base Names testnet compatibility 

cpcramer:
**What changed? Why?**
Base Names was only working on Mainnet. This updates the component to work with EOAs and Smart Wallets on Mainnet and Testnet. 

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/wallet/types.ts
+++ b/src/wallet/types.ts
@@ -1,6 +1,6 @@
 import type { UserOperation } from 'permissionless';
 import type { Dispatch, ReactNode, SetStateAction } from 'react';
-import type { Address, PublicClient } from 'viem';
+import type { Address, Chain, PublicClient } from 'viem';
 import type { UseBalanceReturnType, UseReadContractReturnType } from 'wagmi';
 import type { SwapError } from '../swap';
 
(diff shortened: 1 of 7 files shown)


src/wallet/types.ts
```diff
@@ -61,6 +61,7 @@ export type UseGetTokenBalanceResponse = {
  */
 export type WalletContextType = {
   address?: Address | null; // The Ethereum address to fetch the avatar and name for.
+  chain?: Chain; // Optional chain for domain resolution
```
Zizzamia, line 64: So, I think we probably want to start move away from using `chain` as type. Mostly because we are seeing occasional missmatch with Viem types.

Let's do instead `chainId` for those things going forward.

cpcramer: thoughts on merging this and then updating to chainId in the next PR? Because the chainId change will require `getName`, `Name`, and `useName` updates.

cpcramer: I'll push up the chainId changes and we can split it into two PRs if it's too large

Zizzamia (approved):
Thank you for fixing this.
Make sure to follow up tomorrow with a updated version of `WalletContextType` type.

We should never ship that changes a type without the docs updated as well.

PR #1021, opened 2024-08-09
feat: `WalletDropdownFundLink`

0xAlec:
**What changed? Why?**
- add `WalletDropdown` component that links to https://keys.coinbase.com/funding flow for a Coinbase Smart Wallet.
- has override able text, icon, classname, and different methods (popup vs. link)
- add docs

**Notes to reviewers**
- add component docs in next PR

**How has it been tested?**
locally

## Size `md`
image
image

## Size `sm`
image
image


.changeset/witty-crabs-change.md
```diff
@@ -0,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
+
+**feat**: WalletDropdownFundLink by @0xAlec #1021
```
Zizzamia, line 5: One more thing we need to learn to be good with our Changelog, is to add more readable informations.

Put yourself in the shoes of someone reading this. There is not much meaning in this sentence.

site/docs/pages/wallet/types.mdx
```diff
@@ -86,3 +86,18 @@ export type WalletDropdownLinkReact = {
   target?: string;
 };
 ```
+
+## `WalletDropdownFundLinkReact`
+
+```ts
+export type WalletDropdownFundLinkReact = {
```
Zizzamia, line 93: Keep the alphabetic order.

site/docs/pages/wallet/types.mdx
```diff
@@ -86,3 +86,18 @@ export type WalletDropdownLinkReact = {
...
+  className?: string; // Optional className override for the element
+  href: string;
+  icon?: ReactNode;
+  rel?: string;
+  target?: string;
+  text?: string;
+  type?: 'popup' | 'link';
```
Zizzamia, line 100: Let's add comments to each of one of them. As I don't quite follow what's going on with some of those properties.

site/docs/pages/wallet/wallet-dropdown-fund-link.mdx
```diff
@@ -0,0 +1,191 @@
...
+
+export function WalletComponents() {
+  return (
+    <div className="flex justify-end">
+      <Wallet>
+        <ConnectWallet>
+          <Avatar className="h-6 w-6" />
+          <Name />
+        </ConnectWallet>
+        // [!code focus]
```
Zizzamia, line 58: The location of `code focus` seems broken.

site/docs/pages/wallet/wallet-dropdown-fund-link.mdx
```diff
@@ -0,0 +1,191 @@
...
+        <WalletDropdown>
+          // [!code focus]
+          <Identity className="px-4 pt-3 pb-2" hasCopyAddressOnClick>
+            <Avatar />
+            <Name />
+            <Address className={color.foregroundMuted} />
```
Zizzamia, line 64: Do we need `color` for this example?

site/docs/pages/wallet/wallet-dropdown-fund-link.mdx
```diff
@@ -0,0 +1,191 @@
...
+```tsx
+// omitted for brevity
+
+<Wallet>
+  <ConnectWallet>
+    ...
+  </ConnectWallet>
+  <WalletDropdown>
+    ...
+    <WalletDropdownFundLink type="window" /> // [!code focus]
```
Zizzamia, line 140: `type` is confusing as prop for what you are tyring to do. What's a better name?

site/docs/pages/wallet/wallet-dropdown-fund-link.mdx
```diff
@@ -0,0 +1,191 @@
...
+```tsx
+// omitted for brevity
+
+<Wallet>
+  <ConnectWallet>
+    ...
+  </ConnectWallet>
+  <WalletDropdown>
+    ...
+    <WalletDropdownFundLink type="window" size="s" /> // [!code focus]
```
Zizzamia, line 171: We should not use `size` as prop, because this is a special word we want to use across all OnchainKit.

When you build a Design System you want to think 7 steps ahead. This implementation things only 3 steps ahead, and there is the risk of losing the game and lose our queen.

What's a better API for this?

src/internal/svg/fundWallet.tsx
```diff
@@ -0,0 +1,19 @@
...
+    aria-label="fund-wallet-svg"
+    width="16"
+    height="16"
+    viewBox="0 0 16 16"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
+      d="M8 16C12.4183 16 16 12.4183 16 8C16 3.58172 12.4183 0 8 0C3.58172 0 0 3.58172 0 8C0 12.4183 3.58172 16 8 16ZM1.63962 4.33685C5.1477 6.36224 9.63276 5.16047 11.6582 1.65239C9.63276 5.16047 10.8345 9.64553 14.3413 11.6702C10.8345 9.64553 6.35021 10.846 4.32482 14.3541C6.35021 10.846 5.1477 6.36224 1.63962 4.33685Z"
+      fill="#0A0B0D"
```
Zizzamia, line 15: Does this support the Dark Theme style? cc @kyhyco

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -0,0 +1,74 @@
...
+      width: number;
+      height: number;
+    }
+  > = {
+    s: { width: 400, height: 500 },
+    m: { width: 600, height: 700 },
+    l: { width: 800, height: 900 },
+  };
+
+  const handleClick = (e: React.MouseEvent) => {
```
Zizzamia, line 38: All this logic should be his own utility.

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -0,0 +1,74 @@
...
+      ? {
+          ...commonProps,
+          href: `http://keys.coinbase.com/funding?dappName=${tabName}&dappUrl=${currentURL}`,
+          target,
+          rel,
+        }
+      : { ...commonProps, onClick: handleClick };
+
+  return (
+    <a {...linkProps}>
```
Zizzamia, line 67: Please never do this!!!!!!

Use `...` is a terrible React coding pattern, as you put yourself in the risk to not fully understand and control what are the props used.

site/docs/pages/wallet/wallet-dropdown-fund-link.mdx
0xAlec: `windowSize`?

0xAlec: I'm not sure what the plans are for `size` across OnchainKit but I imagine it'd be similar to this use case

0xAlec: Maybe this can be the first implementation!

site/docs/pages/wallet/wallet-dropdown-fund-link.mdx
0xAlec: `linkType`? `openIn`?

0xAlec: will go with `openIn` for now

site/docs/pages/wallet/wallet-dropdown-fund-link.mdx
0xAlec: changed to `windowSize` for now

src/wallet/components/WalletDropdownFundLink.tsx
0xAlec: most [handleClick](https://github.com/search?q=repo%3Acoinbase%2Fonchainkit%20handleclick&type=code) in our repo are in the same component, do we really need a separate `util` file for just formatting the link? 

it's 6 lines of code - happy to do so if this is a new standard moving forward

site/docs/pages/wallet/types.mdx
```diff
@@ -74,6 +74,21 @@ export type WalletDropdownDisconnectReact = {
...
 
+## `WalletDropdownFundLinkReact`
+
+```ts
+export type WalletDropdownFundLinkReact = {
+  className?: string; // Optional className override for the element
+  href: string;
+  icon?: ReactNode;
+  rel?: string;
+  openIn?: 'window' | 'tab'; // Whether to open the funding flow in a tab or a window
```
Zizzamia, line 85: is this is a `popup` should we probably use that word instead?

site/docs/pages/wallet/types.mdx
```diff
@@ -74,6 +74,21 @@ export type WalletDropdownDisconnectReact = {
...
+  target?: string; // Where to open the target if `openIn` is set to tab
+  text?: string; // Optional text override
+  windowSize?: 's' | 'm' | 'l'; // Size of the popup window if `openIn` is set to window
```
Zizzamia, line 88: same here, is this a `popupSize` ?

src/wallet/components/WalletDropdownFundLink.test.tsx
```diff
@@ -0,0 +1,97 @@
...
+  'http://keys.coinbase.com/funding?dappName=&dappUrl=http://localhost:3000/';
+
+describe('WalletDropdownFundLink', () => {
+  it('renders correctly with default props', () => {
+    render(<WalletDropdownFundLink />);
+
+    const linkElement = screen.getByRole('link');
+    expect(linkElement).toBeInTheDocument();
+    expect(linkElement).toHaveAttribute('href', FUNDING_URL);
+    expect(screen.getByText('Fund wallet')).toBeInTheDocument();
```
Zizzamia, line 15: Github is showing some failure here. Are you seeing these?

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -0,0 +1,81 @@
...
+    }
+    if (isValidElement(icon)) {
+      return icon;
+    }
+  }, [icon]);
+
+  const currentURL = window.location.href;
+  const tabName = document.title;
+  const fundingUrl = `http://keys.coinbase.com/funding?dappName=${tabName}&dappUrl=${currentURL}&onchainkit=${version}`;
+  const windowSizes: Record<
```
Zizzamia, line 28: Let's have his own type for this, so it's easier to read.

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -0,0 +1,81 @@
...
+    's' | 'm' | 'l',
+    {
+      width: number;
+      height: number;
+    }
+  > = {
+    s: { width: 400, height: 500 },
```
Zizzamia, line 35: Who is deciding those sizes? are those standardize anywhere?

src/wallet/components/WalletDropdownFundLink.tsx
Zizzamia: What needs to be in his own util are isolate `input` / `output` operations.

In this case the `getWindowDetails` is his own utility.

Plus will make the test and code, easier to write and mantain.

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const handleClick = (e: React.MouseEvent) => {
+    e.preventDefault();
+    const { width, height } = windowSizes[windowSize];
+    const left = (window.screen.width - width) / 2;
+    const top = (window.screen.height - height) / 2;
+    const windowFeatures = `width=${width},height=${height},resizable,scrollbars=yes,status=1,left=${left},top=${top}`;
+    window.open(fundingUrl, 'Coinbase Fund Wallet', windowFeatures);
+  };
+
+  const commonClassName = cn(
```
Zizzamia, line 49: `common`? what's a better wording here?

src/wallet/components/WalletDropdownFundLink.test.tsx
0xAlec: oops lol, forget i need to update the test every time i make a copy change here

site/docs/pages/wallet/types.mdx
0xAlec: it is a **popup window** but I like `window` terminology better since `popup` has a negative connotation and makes it seems like it's spam (popup spam)

site/docs/pages/wallet/types.mdx
0xAlec: see above comment on usage of `window` instead of `popup`

src/wallet/components/WalletDropdownFundLink.tsx
0xAlec: changed to `overrideClassName`

src/wallet/components/WalletDropdownFundLink.tsx
0xAlec: nope - just values I think look good from testing on a variety of screen sizes. changed to viewport units instead

site/docs/pages/wallet/types.mdx
Zizzamia: `window` is a very special word in web. As it means both the global object, and the window browser size.

I am kind of hesitant to use that word.

site/docs/pages/wallet/types.mdx
Zizzamia: yeah `window` is a very special word in Javascript, not sure how I feel in using that.

src/wallet/components/WalletDropdownFundLink.tsx
Zizzamia: ok if that is the case, would love to use different size wording, like `small`, `medium`, `large`.

Just dropping s, m, l, I am worry to start some bad practice. As this is the first time we do this kind of thing.

src/wallet/components/WalletDropdownFundLink.test.tsx
Zizzamia: So that's the thing, if you write code that is more atomic, you can also write tests are less incline to change all the time.

src/internal/svg/fundWallet.tsx
0xAlec: image

image

src/wallet/utils/getWindowDimensions.ts
```diff
@@ -0,0 +1,35 @@
+import type { WindowSizes } from '../types';
+
+const popupSizes: WindowSizes = {
+  sm: { width: '24.67vw', height: '30.83vw' },
+  md: { width: '29vw', height: '36.25vw' },
+  lg: { width: '35vw', height: '43.75vw' },
+};
```
cpcramer, line 7: Nice

site/docs/pages/wallet/types.mdx
```diff
@@ -82,6 +82,21 @@ export type WalletDropdownDisconnectReact = {
...
+## `WalletDropdownFundLinkReact`
+
+```ts
+export type WalletDropdownFundLinkReact = {
+  className?: string; // Optional className override for the element
+  href: string;
+  icon?: ReactNode;
+  rel?: string;
+  openIn?: 'popup' | 'tab'; // Whether to open the funding flow in a tab or a popup window
+  target?: string; // Where to open the target if `openIn` is set to tab
```
cpcramer, line 94: I'm a little confused about the `target`. Isn't the only target `https://keys.coinbase.com/funding`?

If they want to have a customizable link then they would use `WalletDropdownLink`

cpcramer: Perhaps i'm confusing the target with the src - but still confused about what the target is.

0xAlec: https://www.w3schools.com/tags/att_a_target.asp

`target` is a different attribute than `href` - `target` will control where `href` opens (i.e. new tab or replace current tab)

src/internal/svg/fundWallet.tsx
```diff
@@ -0,0 +1,19 @@
+import { fill } from '../../styles/theme';
+
+export const FundWalletSvg = (
+  <svg
+    role="img"
+    aria-label="fund-wallet-svg"
+    width="16"
+    height="16"
+    viewBox="0 0 16 16"
```
cpcramer, line 9:
```suggestion
    width="100%"
    height="100%"
    viewBox="0 0 20 20"
```

site/docs/pages/wallet/types.mdx
0xAlec: `href` is hardcoded here to be keys.coinbase.com/fund

src/internal/svg/fundWallet.tsx
cpcramer: This should be more flexible as it allows the parent div to control the sizing

src/wallet/components/WalletDropdownFundLink.test.tsx
```diff
@@ -0,0 +1,118 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { expect, it, vi } from 'vitest';
+import { version } from '../../version';
+import type { WindowSizes } from '../types';
+import { WalletDropdownFundLink } from './WalletDropdownFundLink';
+
+const FUNDING_URL = `http://keys.coinbase.com/fund?dappName=&dappUrl=http://localhost:3000/&version=${version}&source=onchainkit`;
```
cpcramer, line 7: Lets put this const in /src/wallet/constants.ts

https://github.com/coinbase/onchainkit/blob/main/src/wallet/constants.ts

site/docs/pages/wallet/types.mdx
```diff
@@ -82,6 +82,21 @@ export type WalletDropdownDisconnectReact = {
...
+  text?: string; // Optional text override
+  popupSize?: 'sm' | 'md' | 'lg'; // Size of the popup window if `openIn` is set to `popup`
```
cpcramer, line 96: I'm wondering if we should even allow the popupSize to be adjustable. 

It seems to add a bit of complexity. 

What if we only supported the "recommended" size - what we think looks best?

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -0,0 +1,76 @@
...
+    e.preventDefault();
+    const { width, height } = getWindowDimensions(popupSize);
+
+    const left = Math.round((window.screen.width - width) / 2);
+    const top = Math.round((window.screen.height - height) / 2);
+
+    const windowFeatures =
+      popupFeatures ||
+      `width=${width},height=${height},resizable,scrollbars=yes,status=1,left=${left},top=${top}`;
+    window.open(fundingUrl, 'Coinbase Fund Wallet', windowFeatures);
```
cpcramer, line 41: We can assign 'Coinbase Fund Wallet' to a const

site/docs/pages/wallet/types.mdx
cpcramer: Thanks I was confusing the two

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { getWindowDimensions } from '../utils/getWindowDimensions';
+
+export function WalletDropdownFundLink({
+  className,
+  icon,
+  rel,
+  popupFeatures,
+  openIn = 'tab',
+  target,
+  text = 'Fund wallet',
```
cpcramer, line 15: Do we want to offer text override? 

In the BaseNames component we hardcode the text, I feel like we could do the same with the fund link.

src/internal/svg/fundWallet.tsx
0xAlec: done, thanks!

src/wallet/components/WalletDropdownFundLink.tsx
0xAlec: this should actually just be the `target` passed in by the user - fixed in new commit

src/wallet/components/WalletDropdownFundLink.tsx
0xAlec: design wanted a text override

site/docs/pages/wallet/types.mdx
0xAlec: not opposed to just hardcoding since we already allow the user to override the `windowFeatures` parameter on `window.open` using the `popupFeatures` prop

0xAlec: but IMO having to perform this calculation as a developer is pretty annoying, we should have some sort of pre-built responsive design BUT still allow the user full customizability

cpcramer: + to having a pre-build responsive design

What type of size customizability would the user want? 

Personally, as long as the popup window looks "normal" and everything is in view then I would be happy with it.

site/docs/pages/wallet/types.mdx
Zizzamia: I think overall there is an interesting question to ask. What are properties we want to be able to customize and what we don't want to have customized.

Otherwise it feels a weird overalap with the Link component.

src/wallet/types.ts
```diff
@@ -96,6 +96,21 @@ export type WalletDropdownDisconnectReact = {
...
 };
 
+/**
+ * Note: exported as public Type
+ */
+export type WalletDropdownFundLinkReact = {
+  className?: string; // Optional className override for the element
+  href: string;
+  icon?: ReactNode;
+  rel?: string;
```
abcrane123, line 106: can you add descriptions for these three as well

site/docs/pages/wallet/types.mdx
Zizzamia: How the popup looks like? I didn't see a screenshot in the recap. Can we see how each form looks like.

site/docs/pages/wallet/types.mdx
```diff
@@ -82,6 +82,21 @@ export type WalletDropdownDisconnectReact = {
...
 ```
 
+## `WalletDropdownFundLinkReact`
+
+```ts
+export type WalletDropdownFundLinkReact = {
+  className?: string; // Optional className override for the element
+  href: string;
+  icon?: ReactNode;
+  rel?: string;
```
abcrane123, line 92: can you add descriptions for these three as well

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -0,0 +1,76 @@
...
+  popupSize = 'md',
+}: WalletDropdownFundLinkReact) {
+  const iconSvg = useMemo(() => {
```
Zizzamia, line 18: This logic is identical to the one we have in WalletDropdownLink, we should probably have a shared Hook or Utils for both of them.

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -0,0 +1,76 @@
...
+    window.open(fundingUrl, target, windowFeatures);
+  };
+
+  const overrideClassName = cn(
+    pressable.default,
+    'relative flex items-center px-4 py-3',
+    className,
+  );
+
+  const linkContent = (
```
abcrane123, line 50: you could wrap this in useMemo. benefit would be small but i think it is best practice.

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -0,0 +1,76 @@
...
+    if (isValidElement(icon)) {
+      return icon;
+    }
+  }, [icon]);
+
+  const currentURL = window.location.href;
+  const tabName = document.title;
+  const fundingUrl = `http://keys.coinbase.com/fund?dappName=${tabName}&dappUrl=${currentURL}&version=${version}&source=onchainkit`;
+
+  const handleClick = (e: React.MouseEvent) => {
```
abcrane123, line 31: i would wrap in useCallback

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const iconSvg = useMemo(() => {
+    if (icon === undefined) {
+      return FundWalletSvg;
+    }
+    if (isValidElement(icon)) {
+      return icon;
+    }
+  }, [icon]);
+
+  const currentURL = window.location.href;
```
Zizzamia, line 27: Something to reflect here, is the cost of creating this variable every time the React component render. Which is usually a lot.

So this is possibly something that should be wrapped by a useEffect.

Also in general for @0xAlec @cpcramer @abcrane123 please watch a few times this video https://www.youtube.com/watch?v=NGxVLnJKhP8 it really helps understand React internals.

src/wallet/utils/getWindowDimensions.ts
```diff
@@ -0,0 +1,35 @@
+import type { WindowSizes } from '../types';
+
+const popupSizes: WindowSizes = {
+  sm: { width: '24.67vw', height: '30.83vw' },
```
Zizzamia, line 4: `vw` are we sure we want this? I feel a bit hesitant of using `vw`. Like how this behavies on mobile?

src/wallet/components/WalletDropdownFundLink.tsx
0xAlec: added

src/wallet/utils/getWindowDimensions.ts
0xAlec: added some screenshots on window sizing on desktop and mobile

0xAlec: I think on mobile it will always open in a new tab? I don't think there's popup windows on most mobile browsers

site/docs/pages/wallet/types.mdx
0xAlec: added screenshots on desktop

src/wallet/utils/getWindowDimensions.ts
0xAlec: I'm hoping what will happen on mobile is it will ignore the `windowFeatures` parameter and just open in a new tab, but no way to emulate this mobile behavior on Chrome devtools I don't think

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -0,0 +1,85 @@
+import { useCallback, useMemo } from 'react';
+import { useEffect, useState } from 'react';
+import { cn, pressable, text as themeText } from '../../styles/theme';
+import { version } from '../../version';
+import { useIcon } from '../hooks/useIcon';
+import type { WalletDropdownFundLinkReact } from '../types';
+import { getWindowDimensions } from '../utils/getWindowDimensions';
+
+export function WalletDropdownFundLink({
```
Zizzamia, line 9: nit. alphabetical order

Zizzamia (approved):
Amazing, great job!!!!

src/wallet/components/WalletDropdownFundLink.tsx
0xAlec: fixed in new diff, updated in `types.mdx` as well

PR #1053, opened 2024-08-14
Feat: Better Handling of FrameButton post actions

brendan-defi:
**What changed? Why?**
* updated `FrameButtonMetadata` to allow custom `postUrl` property
* updated `getFrameMetadata` to handle custom `postUrl`s on Frame Buttons

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/frame/types.ts
+++ b/src/frame/types.ts
@@ -95,6 +95,7 @@ export type FrameButtonMetadata =
   | {
       action?: 'post' | 'post_redirect';
       label: string;
+      postUrl?: string;
       target?: string;
     }
   | {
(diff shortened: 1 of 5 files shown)


src/frame/types.ts
```diff
@@ -95,6 +95,7 @@ export type FrameButtonMetadata =
   | {
       action?: 'post' | 'post_redirect';
       label: string;
+      postUrl?: string;
```
Zizzamia, line 98: Is this part of the Frame Spec? And where is this writtend?

brendan-defi: image

Per [Button Actions](https://docs.farcaster.xyz/developers/frames/spec#button-actions), we are told that a `post` should send an HTTP request to the button's `post_url`.

But as currently configured, `getFrameMetadata` does not set a `fc:frame:button:[id]:post_url` property for post actions, only tx actions.

Zizzamia (approved):

PR #1054, opened 2024-08-15
fix: playground build

0xAlec:
**What changed? Why?**
- fix the `next build` step in playground so we can successfully create dev deployments off branches

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/playground/nextjs-app-router/next.config.mjs
+++ b/playground/nextjs-app-router/next.config.mjs
@@ -1,4 +1,16 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {};
-
-export default nextConfig;
+const nextConfig = {
+    experimental: {
+      swcPlugins: [
+        ['next-superjson-plugin', {}]
+      ]
+    },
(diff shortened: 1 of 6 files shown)


playground/nextjs-app-router/next.config.mjs
```diff
@@ -1,4 +1,8 @@
 /** @type {import('next').NextConfig} */
-const nextConfig = {};
-
-export default nextConfig;
+const nextConfig = {
+  typescript: {
+    ignoreBuildErrors: true,
```
0xAlec, line 4: @Zizzamia this is what we landed on to get this working with our current build/link steps, but Bunday pointed me at how they handle local imports in `base-org/web` as a potential alternative:

- https://github.com/base-org/web/blob/master/apps/web/next.config.js#L167
- https://github.com/base-org/web/blob/master/apps/web/tsconfig.json#L34
- https://github.com/base-org/web/blob/master/apps/web/tsconfig.json#L11

Zizzamia (approved):

PR #1055, opened 2024-08-15
feat: keep polishing onStatus lifecycle

Zizzamia:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/transaction/components/TransactionProvider.test.tsx
```diff
@@ -252,30 +259,6 @@ describe('TransactionProvider', () => {
     });
   });
 
-  it('should call onSuccess when receipts are available', async () => {
```
Zizzamia, line 255: Apperently this test was not testing what was suppose to test 🤦🏻‍♂️

src/transaction/components/TransactionProvider.test.tsx
```diff
@@ -382,34 +365,6 @@ describe('TransactionProvider', () => {
       );
     });
   });
-
-  it('should call onSuccess when receiptArray has receipts', async () => {
```
Zizzamia, line 386: Same for this one. I am rewriting the entire logic in a more simple way so we can actually test this.

PR #1071, opened 2024-08-15
fix: default accountChain for Identity component

Zizzamia:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/identity/components/Identity.tsx
```diff
@@ -6,16 +6,14 @@ import { IdentityProvider } from './IdentityProvider';
...
-  schemaId,
   hasCopyAddressOnClick = false,
-  chain,
+  schemaId,
 }: IdentityReact) {
-  // istanbul ignore next
   const { chain: contextChain } = useOnchainKit();
-
-  const accountChain = contextChain || chain;
+  const accountChain = chain ?? contextChain;
```
Zizzamia, line 16: @abcrane123 that's why we were getting confused. We had the wrong order.

playground/nextjs-app-router/next.config.mjs
```diff
@@ -1,5 +1,6 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  basePath: '/playground',
```
Zizzamia, line 3: Adding some extra stuff to fix the playground story

PR #1089, opened 2024-08-16
feat: Swap success state - refetch balances and clear inputs

0xAlec:
**What changed? Why?**
- add `FromTo` type
- add `balanceResponse` in `SwapUnit` type
- return `balanceResponse` from Wagmi balance queries in `useFromTo` and `useSwapBalances`
- add `useResetInputs` hook

clears the inputs and refetches balances on a successful swap

**Notes to reviewers**

- no success `SwapMessage` rendered yet
- this does not count for transaction reverts, we will need to add this handling separately

**How has it been tested?**
playground - try making a swap


playground/nextjs-app-router/components/demo/Swap.tsx
```diff
@@ -94,7 +94,7 @@ function SwapComponent() {
             token={usdcToken}
             type="to"
           />
-          <SwapButton disabled={true} />
+          <SwapButton />
```
0xAlec, line 97: TODO: Redisable this before merging PR

src/swap/hooks/useFromTo.ts
```diff
@@ -17,6 +23,9 @@ export const useFromTo = (address: Address) => {
     fromTokenBalanceError,
     toBalanceString,
     toTokenBalanceError,
+
```
abcrane123, line 26: super nit but you can remove this

abcrane123 (dismissed):
this looks good to me !

abcrane123 (commented):
actually make sure to add back disabled flag

src/swap/hooks/useFromTo.test.ts
```diff
@@ -57,6 +61,77 @@ describe('useFromTo', () => {
...
+
+  it('should call fromTokenResponse.refetch when from.refetch is called', async () => {
+    const mockFromRefetch = vi.fn().mockResolvedValue(undefined);
+    const mockToRefetch = vi.fn().mockResolvedValue(undefined);
+
+    (useSwapBalances as vi.Mock).mockReturnValue({
+      fromTokenResponse: { refetch: mockFromRefetch },
+      toTokenResponse: { refetch: mockToRefetch },
+    });
+
```
Zizzamia, line 76: Let's get into the habit to remove spaces from tests, so they are much easier to read, review, mantain.

src/swap/hooks/useFromTo.ts
```diff
@@ -2,9 +2,15 @@ import { useState } from 'react';
...
+import type { SwapInput } from '../types';
 import { useSwapBalances } from './useSwapBalances';
 
-export const useFromTo = (address?: Address) => {
+export const useFromTo = (
+  address?: Address,
+): {
+  from: SwapInput;
+  to: SwapInput;
+} => {
```
Zizzamia, line 13: Let's create a new type for this

src/swap/hooks/useFromTo.ts
```diff
@@ -28,6 +36,9 @@ export const useFromTo = (address?: Address) => {
     loading: fromLoading,
     setLoading: setFromLoading,
     error: fromTokenBalanceError,
+    refetch: async () => {
```
Zizzamia, line 39: Not a fan of this, as it creates weird bad practices.

I want us to think more into the Life Cycle Status mindset.

Like what status we are changing, and that will trigger an action.

0xAlec: PR was incomplete when you reviewed - please check the updated diffs

to clarify, do you mean returning the entire `tokenResponse` instead of just the `refetch` attribute of the `tokenResponse` object here? I doubt we'll need the other fields

0xAlec: either `refetch` or the entire object has to be returned here for us to be able to refetch the token balances from the `SwapProvider` lifecycle status hook

src/swap/components/SwapProvider.test.tsx
```diff
@@ -245,6 +251,22 @@ describe('SwapProvider', () => {
...
 
+  it('should reset inputs when setLifeCycleStatus is called with success', async () => {
+    const { result } = renderHook(() => useSwapContext(), { wrapper });
+    await act(async () => {
+      result.current.setLifeCycleStatus({
+        statusName: 'success',
+        statusData: { transactionReceipt: '0x123' },
+      });
+    });
+
```
Zizzamia, line 262: In the future let's avoid spaces in tests, I noticed that when reading them in a very long file, those spaces really make hard to parse many tests all at once.

Let's keep a space only between `it` but not inside an `it`.

src/swap/components/SwapProvider.tsx
```diff
@@ -58,6 +59,9 @@ export function SwapProvider({
   const { from, to } = useFromTo(address);
   const { sendTransactionAsync } = useSendTransaction(); // Sending the transaction (and approval, if applicable)
 
+  // Refreshes balances and inputs post-swap
```
Zizzamia, line 62: Love the comment

src/swap/types.ts
```diff
@@ -219,6 +232,7 @@ export type SwapUnit = {
   balance?: string;
   error?: SwapError;
   loading: boolean;
+  refetch: () => Promise<void>;
```
Zizzamia, line 235: Let's add a line of comment to explain what this does.

The word `refetch` feels very vague and hard to understand in a few weeks from now.

src/swap/types.ts
```diff
@@ -147,6 +155,11 @@ export type SwapError = {
   message: string; // The error message providing details about the swap error.
 };
 
+export type SwapErrorState = {
```
Zizzamia, line 158: Why this is here? I don't see it used in the file.

src/swap/index.ts
```diff
@@ -8,6 +8,7 @@ export type {
   BuildSwapTransaction,
   BuildSwapTransactionResponse,
   Fee,
+  FromTo,
```
Zizzamia, line 11: Why are exporting this to all developers, it feels very vague and not sure I see the usage for them.

src/swap/hooks/useFromTo.ts
```diff
@@ -39,6 +45,9 @@ export const useFromTo = (address?: Address) => {
     loading: toLoading,
     setLoading: setToLoading,
     error: toTokenBalanceError,
+    refetch: async () => {
+      await toTokenResponse?.refetch();
```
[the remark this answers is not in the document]
Zizzamia: So should `to` and `from` just have a section named `tokenResponse`?

src/swap/index.ts
0xAlec: removed

src/swap/types.ts
0xAlec: using `tokenResponse` instead

src/swap/components/SwapProvider.test.tsx
0xAlec: removed - thanks for catching this

src/swap/hooks/useFromTo.ts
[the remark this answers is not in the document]
0xAlec: changed to `tokenResponse` instead - https://github.com/coinbase/onchainkit/pull/1089/commits/33644ecd97f0e098bbc091dc90c034796534a2de#diff-8b9703e75aad4aea2ad9c4027b4746333b9d5471c15f41a4505293cd78d50be9

src/swap/types.ts
```diff
@@ -219,6 +228,7 @@ export type SwapUnit = {
   balance?: string;
   error?: SwapError;
   loading: boolean;
+  response: UseBalanceReturnType | UseReadContractReturnType | undefined;
```
Zizzamia, line 231: nit. we can just add `response?` to the type, and that will be as `undefined`. So it will be
```
response?: UseBalanceReturnType | UseReadContractReturnType;
```

src/swap/types.ts
```diff
@@ -1,6 +1,10 @@
 import type { Dispatch, ReactNode, SetStateAction } from 'react';
 import type { Address, Hex, TransactionReceipt } from 'viem';
-import type { Config } from 'wagmi';
+import type {
+  Config,
+  UseBalanceReturnType,
```
Zizzamia, line 5: Niiice, this will make much easier understand what's going on.

src/swap/hooks/useSwapBalances.tsx
```diff
@@ -37,12 +46,21 @@ export function useSwapBalances({
...
+    ? ethBalanceResponse
+    : _toTokenResponse;
 
   return useValue({
     fromBalanceString,
     fromTokenBalanceError,
 
     toBalanceString,
     toTokenBalanceError,
+
```
Zizzamia, line 62: Why are we adding those places, if you want to add spaces to split the meaning, we probably just want to add a line of comments between them, and explain more clearly the three groups.

src/swap/hooks/useResetInputs.ts
```diff
@@ -0,0 +1,14 @@
+import { useCallback } from 'react';
+import type { FromTo } from '../types';
+
+// Refreshes balances and inputs post-swap
+export const useResetInputs = ({ from, to }: FromTo) => {
+  return useCallback(async () => {
+    await Promise.all([
+      from.response?.refetch(),
+      to.response?.refetch(),
```
Zizzamia, line 9: Not my favorite, but at least feels a bit less crazy.

src/swap/hooks/useFromTo.ts
```diff
@@ -28,6 +31,7 @@ export const useFromTo = (address?: Address) => {
     loading: fromLoading,
     setLoading: setFromLoading,
     error: fromTokenBalanceError,
+    response: fromTokenResponse,
```
Zizzamia, line 34: Ok dope, can you give it an alphabetical order clean up, for the `from` and `to`. So it's easier to follow.

src/swap/hooks/useSwapBalances.tsx
0xAlec: added to the two previous groups

src/swap/types.ts
0xAlec: done

src/swap/hooks/useFromTo.ts
0xAlec: cleaned up/ordered alphabetically

src/swap/hooks/useFromTo.ts
```diff
@@ -39,6 +43,7 @@ export const useFromTo = (address?: Address) => {
     loading: toLoading,
     setLoading: setToLoading,
     error: toTokenBalanceError,
+    response: toTokenResponse,
```
Zizzamia, line 46: One more open question.

We say this is a `response`. But what actually we are saying this is a `response` to an Hook related to `UseBalanceReturnType`.

So in reality this is a balance object that helps us `fetching native currency or token balance`.

Mmmmmmm so `response` as name is not actually correct.

It feels more a `balanceSomething`. mmmmm I mean this is the return type https://wagmi.sh/react/api/hooks/useBalance#return-type .

0xAlec: it can also be [useReadContractReturnType](https://wagmi.sh/react/api/hooks/useReadContract#return-type) which isn't necessary a balance. I think the `response` convention is from [Tanstack mutations](https://tanstack.com/query/v4/docs/framework/react/guides/updates-from-mutation-responses)

0xAlec: but happy to call this something like `balanceResponse` since we only use it for balances (ETH or tokens)

0xAlec: regardless I think since this is only used internally for now we can always rename it later on without introducing a breaking change

0xAlec: IMO refetching is the only useful field exposed from this object but not that familiar with Tanstack, maybe @alissacrane-cb can provide more input

Zizzamia: > regardless since this is only used internally
if somethings it's hard to understand, than for us it's hard to refactor. Especially as we have so many folks working on the same files.

Let's go ahead with `balanceResponse`.

abcrane123 (approved):
nice this lgtm!

Zizzamia (approved):

cpcramer (approved):

PR #1092, opened 2024-08-17
chore: Fix hover state for all components

cpcramer:
**What changed? Why?**
Fix hover state styling for all components and update specific buttons

Issue:
- Our themed CSS class variables were not correctly applying the custom hover styles.
- As a result, none of the components were displaying our intended default hover styling.

Fix:
1. In theme.ts: Corrected how we access the hover state variable.
2. For TransactionButton and WalletDropdown:
   - Implemented the 'pressable' class.
   - This maintains their current styling while also enabling the use of our defined state classes (including hover).

Outcome:
- All components now correctly display the default hover styling.
- TransactionButton and WalletDropdown have consistent styling with added state-based effects.

[Watch the video walkthrough to see the tested behavior](https://www.loom.com/share/82aeac2b82744924b40eb8262f9edb01?sid=86caa2d3-998b-449c-8279-fb9ec0261f7b)

**Notes to reviewers**

**How has it been tested?**
Confirmed for the Connect Wallet button and Transaction button in light and dark modes.

### DIFF
--- a/src/transaction/components/TransactionButton.tsx
+++ b/src/transaction/components/TransactionButton.tsx
@@ -86,7 +86,7 @@ export function TransactionButton({
   return (
     <button
       className={cn(
-        pressable.primary,
+        background.primary,
         'w-full rounded-xl',
         'mt-4 px-4 py-3 font-medium text-base text-white leading-6',
         isDisabled && pressable.disabled,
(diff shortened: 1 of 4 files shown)


src/transaction/components/TransactionButton.tsx
```diff
@@ -86,7 +86,7 @@ export function TransactionButton({
   return (
     <button
       className={cn(
-        background.primary,
+        pressable.primary,
```
Zizzamia, line 89: oh, very interesting

Zizzamia (approved):

PR #1113, opened 2024-08-20
chore: Move all scattered SVGs to shared location

cpcramer:
**What changed? Why?**
More SVGs that are scattered around our code base into the common `/src/internal/svg` folder
**Notes to reviewers**

**How has it been tested?**


src/internal/svg/badgeSvg.tsx
```diff
@@ -0,0 +1,20 @@
+import { fill } from '../../styles/theme';
+
+export const badgeSvg = (
+  <svg
+    role="img"
+    aria-label="ock-attestation-icon"
```
Zizzamia, line 6: ock-badgeSvg instead

src/internal/svg/caretDownSvg.tsx
```diff
@@ -0,0 +1,19 @@
+import { fill } from '../../styles/theme';
+
+export const caretDownSvg = (
+  <svg
+    data-testid="ockTokenSelectButton_CaretDown"
```
Zizzamia, line 5: same here and in other places.

cpcramer: okay on it

cpcramer: Updating the existing data-testid fields and the aria-label fields.

Zizzamia (approved):

Issue #1077, opened 2024-08-15
Bug: Hydration issue with wallet component


mykcryptodev:
### Describe the bug and the steps to reproduce it

1. Create a standard Next.js project with the pages router using client side rendering
2. Follow [steps to put the wallet component](https://onchainkit.xyz/wallet/wallet#wallet-) anywhere on the page
3. Connect your wallet
4. Refresh the page

### What's the expected behavior?

Expecting no hydration issues when reloading

https://www.loom.com/share/e6a77177fc974dccb44438f0e1ae25d7

Zizzamia:
A few quick question:
- what next version?
- is app or pages routing?
- are you using `use client`?

mykcryptodev:
I am using
- "next": "^14.2.4"
- pages routing
- not using use client because I am using pages routing

Zizzamia:
@abcrane123 fix this with `v0.30`

PR #1117, opened 2024-08-20
chore: Fix wallet hydration issue on next.js

abcrane123:
**What changed? Why?**
Following up on https://github.com/coinbase/onchainkit/issues/1077

User reported receiving a hydration warning from their next.js app when integrating the `Wallet` component which is likely because the `Wallet` component depends on client-specific data. The user is using pages routing and components in the pages/ directory are considered to be client-side by default, which, in this case, results in a hydration error. 

I am resolving this issue by returning `null` in `Wallet` component until the component mounts. 

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/swap/components/Swap.tsx
+++ b/src/swap/components/Swap.tsx
@@ -7,6 +7,7 @@ import { SwapButton } from './SwapButton';
 import { SwapMessage } from './SwapMessage';
 import { SwapProvider } from './SwapProvider';
 import { SwapToggleButton } from './SwapToggleButton';
+import { useIsMounted } from '../../useIsMounted';
 
 export function Swap({
   address,
@@ -25,6 +26,13 @@ export function Swap({
       swapMessage: childrenArray.find(findComponent(SwapMessage)),
     };
   }, [children]);
+
+  const isMounted = useIsMounted();
+
+  // prevents SSR hydration issue
+  if (!isMounted) {
+    return null;
+  }
 
   return (
     <SwapProvider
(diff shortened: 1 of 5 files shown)


src/wallet/components/Wallet.tsx
```diff
@@ -43,6 +43,17 @@ const WalletContent = ({ children }: WalletReact) => {
 };
 
 export const Wallet = ({ children }: WalletReact) => {
+  const [isClient, setIsClient] = useState(false);
+
+  useEffect(() => {
+    setIsClient(true);
```
Zizzamia, line 49: Let's try to solve this for 2~3 components at the same time.

As I see this being an issue for Identity components as well.

src/wallet/components/Wallet.tsx
```diff
@@ -43,6 +43,17 @@ const WalletContent = ({ children }: WalletReact) => {
 };
```
Zizzamia, line 44: Let's make sure there is a changeset for this PR as well.

src/swap/components/Swap.tsx
```diff
@@ -26,6 +27,13 @@ export function Swap({
     };
   }, [children]);
 
+  const isMounted = useIsMounted();
```
Zizzamia, line 30: This is really nice

src/wallet/components/Wallet.tsx
abcrane123: i tested `Identity` in my app and it didn't seem to be an issue

Zizzamia (approved):

PR #1119, opened 2024-08-20
chore: Swap lifeCycleStatus cleanup

Zizzamia:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/swap/utils/getErrorMessage.ts
```diff
@@ -0,0 +1,22 @@
+import {
```
cpcramer, line 1: Nice! I like these getErrorMessage files

src/swap/components/SwapProvider.tsx
```diff
@@ -139,13 +133,20 @@ export function SwapProvider({
 
         destination.setAmount(formattedAmount);
       } catch (err) {
-        handleError({ quoteError: err as SwapError });
+        setLifeCycleStatus({
+          statusName: 'error',
+          statusData: {
+            code: 'TmSPc01', // Transaction module SwapProvider component 01 error
```
cpcramer, line 139: Hmm I don't love this error code

Zizzamia: This is going to be the new starnard for all error code, where the goal is to create location from where the error is happening.

See error code as longitude and latitude.

So in this case TmSPc01 means
Tm = Transaction module
SPc = Swap Provider component
01 = First error of the file

Zizzamia: I am going to rewrite all the error code to be more like this, and instead have the Message and Error value be more about description on what's up.

Zizzamia: Have an error code as location, help debugging, and support developers when something is broken, as the first thing we will check, it's where the Error Code is originated.

cpcramer: Interesting, ok got it! Once all error codes match this format i'm sure I'll get used to it quickly

Zizzamia: cc @abcrane123 @cpcramer @0xAlec

cpcramer (approved):

src/swap/components/SwapProvider.test.tsx
```diff
@@ -144,41 +172,6 @@ describe('SwapProvider', () => {
...
   });
 
-  it('should handle submit correctly', async () => {
-    const mockOnError = vi.fn();
-    const mockOnSuccess = vi.fn();
-    let _submitFunction: (
-      onError?: (error: SwapError) => void,
-      onSuccess?: (txReceipt: TransactionReceipt) => void | Promise<void>,
-    ) => void;
-    const TestComponent = () => {
```
Zizzamia, line 154: Should be only one TestComponent we use for various tests.

src/swap/components/SwapProvider.test.tsx
```diff
@@ -326,29 +310,87 @@ describe('SwapProvider', () => {
...
+          .textContent,
+      ).toBe('error');
+      expect(
+        screen.getByTestId('context-value-lifeCycleStatus-statusData-code')
+          .textContent,
+      ).toBe('TmSPc02');
+    });
+  });
+
+  it('should setLifeCycleStatus to error when buildSwapTransaction returns an error', async () => {
```
Zizzamia, line 377: Adding more tests and re-organizing the tests as the order on where the code is located.

src/swap/components/SwapProvider.tsx
```diff
@@ -46,28 +46,15 @@ export function SwapProvider({
...
+  const config = useConfig();
   const [loading, setLoading] = useState(false);
   const [isTransactionPending, setPendingTransaction] = useState(false);
   const [lifeCycleStatus, setLifeCycleStatus] = useState<LifeCycleStatus>({
     statusName: 'init',
     statusData: null,
   }); // Component lifecycle
-  const [error, setError] = useState<SwapErrorState>();
-
-  const handleError = useCallback(
```
Zizzamia, line 57: handleError is replaced with the setLifeCycleStatus

src/swap/types.ts
```diff
@@ -72,7 +72,7 @@ export type GetSwapQuoteParams = {
 export type GetSwapQuoteResponse = SwapQuote | SwapError;
 
 export type GetSwapMessageParams = {
-  error?: SwapErrorState;
```
Zizzamia, line 75: Goodbye many type of errors, and we should have only one type.

src/swap/utils/getSwapMessage.test.ts
```diff
@@ -93,7 +93,7 @@ describe('getSwapMessage', () => {
     expect(getSwapMessage(params2)).toBe(SwapMessage.FETCHING_QUOTE);
   });
 
-  test('returns INCOMPLETE_FIELD when required fields are missing', () => {
```
Zizzamia, line 96: it should always be `it`, those `test` is probably some bad CoPilot advice.

vitest.config.ts
```diff
@@ -23,10 +23,10 @@ export default defineConfig({
       ],
       reportOnFailure: true,
       thresholds: {
-        statements: 99.58,
-        branches: 99.56,
+        statements: 99.65,
```
Zizzamia, line 26: Getting closer to 100%

PR #1121, opened 2024-08-21
feat: add commerce charge hydration utilities and supporting types

avidreder:
**What changed? Why?**
- Adds network util to hydrate a Commerce change via commerce api, with error parsing utility
- Adds types for response and errors

**Notes to reviewers**
- The Charge is a large type, I added only the fields that I deemed necessary for this milestone of the project; we can extend it later if we need to

**How has it been tested?**
- Have tested it locally with a larger component, which is not included in this PR

### DIFF
--- a/src/pay/types/charge.ts
+++ b/src/pay/types/charge.ts
@@ -1,121 +1,120 @@
 export type CommerceCharge = {
-    code: string;
-    description?: string;
-    expires_at: string;
-    hosted_url: string;
-    id: string;
-    name?: string;
-    organization_name?: string;
-    pricing: Pricing;
-    pricing_type: PricingType;
-    redirects?: Redirects;
-    support_email: string;
-    timeline: TimelineItem[];
-    web3_data: Web3Data;
-  };
-  
-  export type SettlementCurrencyAddresses = {
-    [chainId: string]: `0x${string}`;
-  };
-  
-  export type ContractAddresses = {
-    [chainId: string]: `0x${string}`;
-  };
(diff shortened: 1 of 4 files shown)


src/pay/types/charge.ts
```diff
@@ -0,0 +1,120 @@
+export type CommerceCharge = {
```
Zizzamia, line 1: Dope!

For all types, just add them in `src/pay/types.ts`.

src/pay/types/charge.ts
```diff
@@ -0,0 +1,120 @@
...
+  signature: `0x${string}`;
+};
+
+export type TransferIntent = {
+  call_data: TransferIntentCallData;
+  metadata: TransferIntentMetadata;
+};
+
+export type SuccessEventsItem = {
+  chain_id: number;
```
Zizzamia, line 57: No type should be snake_case, all types should be camelCase.

src/pay/types/charge.ts
Zizzamia: Also do not add types that are not used yet.

It will help us have better code review of adding types only of things used.

src/pay/types/errorResponse.ts
```diff
@@ -0,0 +1,4 @@
+export type CommerceApiError = {
```
Zizzamia, line 1: All types have to be in the same file.

avidreder:
Closing in favor of an RPC implementation

PR #1123, opened 2024-08-21
feat: moved onSuccess and onError for Swap at top level

Zizzamia:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/swap/components/Swap.tsx
+++ b/src/swap/components/Swap.tsx
@@ -13,7 +13,9 @@ export function Swap({
   children,
   className,
   experimental = { useAggregator: true },
+  onError,
   onStatus,
+  onSuccess,
   title = 'Swap',
 }: SwapReact) {
   const { inputs, toggleButton, swapButton, swapMessage } = useMemo(() => {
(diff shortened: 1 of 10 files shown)


src/swap/components/Swap.tsx
```diff
@@ -13,7 +13,9 @@ export function Swap({
   children,
   className,
   experimental = { useAggregator: true },
+  onError,
```
Zizzamia, line 16: This will make the experience between Swap and Transaction similar.

In the next couple PRs, than I need to make the breaking change of removing those from the Button component.

src/swap/index.ts
```diff
@@ -13,6 +13,7 @@ export type {
   Fee,
   GetSwapQuoteParams,
   GetSwapQuoteResponse,
+  LifeCycleStatus,
```
Zizzamia, line 16: Exporting this, so we can start making the changes in the docs and Playground.

vitest.config.ts
```diff
@@ -23,10 +23,10 @@ export default defineConfig({
       ],
       reportOnFailure: true,
       thresholds: {
-        statements: 99.65,
-        branches: 99.65,
+        statements: 99.67,
+        branches: 99.7,
```
Zizzamia, line 27: So close to 100!!!!!!!

PR #1124, opened 2024-08-21
chore: test

Zizzamia:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


vitest.config.ts
```diff
@@ -25,7 +25,7 @@ export default defineConfig({
       thresholds: {
         statements: 99.67,
         branches: 99.7,
-        functions: 99.57,
+        functions: 100,
```
Zizzamia, line 28: And on this one, we are back to 100!!!

PR #1126, opened 2024-08-21
docs: Remove landing page figma image

cpcramer:
**What changed? Why?**
We are removing the figma image on the landing page because the example was meant to be interactive, but the current image is static. Perhaps once we have bandwidth we can add back in the interactive figma demo. 

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>

Screenshot 2024-08-21 at 10 15 15 AM

</td>
    <td>
Screenshot 2024-08-21 at 10 14 49 AM

   </td>
  </tr>
</table>

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/site/docs/pages/index.mdx
+++ b/site/docs/pages/index.mdx
@@ -440,7 +440,7 @@ bun add @coinbase/onchainkit
 
 {' '}
 
-<section className="css-alternate-container w-full flex flex-col items-center py-24">
+<section className="css-alternate-container w-full flex flex-col items-center pt-20">
   <div className="w-[1225px] max-w-full flex flex-col justify-between">
     <div className="flex flex-col md:flex-row justify-between items-center pb-[20px]">
       <div>
@@ -461,17 +461,7 @@ bun add @coinbase/onchainkit
   </div>
 </section>
 
-{' '}
-
-<section className="css-alternate-container w-full flex items-center justify-center">
-  <aside className="flex flex-col pb-10 md:flex-row md:px-2 w-[940px]">
-    <img alt="OnchainKit figma" src="/assets/onchain-figma.png" width="auto" />
-  </aside>
-</section>
-
-{' '}
-
-<section className="css-alternate-container w-full flex flex-col items-center py-24 gap-[72px]">
+<section className="css-alternate-container w-full flex flex-col items-center pt-20 pb-24 gap-[72px]">
   <div>
     <h3 className="basis-1/2 py-[15px] text-4xl md:text-4xl max-w-[525px] text-center text-gray-950 dark:text-gray-50">Builders ship faster with OnchainKit</h3>
   </div>


site/docs/pages/index.mdx
```diff
@@ -440,7 +440,7 @@ bun add @coinbase/onchainkit
 
 {' '}
 
-<section className="css-alternate-container w-full flex flex-col items-center py-24">
+<section className="css-alternate-container w-full flex flex-col items-center pt-20">
```
cpcramer, line 443: Updating the top and bottom padding. Otherwise the spacing looks off.

site/docs/pages/index.mdx
```diff
@@ -461,17 +461,7 @@ bun add @coinbase/onchainkit
   </div>
 </section>
 
-{' '}
-
-<section className="css-alternate-container w-full flex items-center justify-center">
-  <aside className="flex flex-col pb-10 md:flex-row md:px-2 w-[940px]">
```
Zizzamia, line 467: Why removing in it?

What's the long term plan for this?

How do we tell the story around Figma

cpcramer: We are removing the image because the current image is meant to be interactive. It's odd to see the zoom in and out buttons for example. 

Screenshot 2024-08-21 at 10 40 43 AM

cpcramer: So I was thinking that we could update to a more appropriate image that is meant to be static. Or perhaps just remove it until we can implement the interactive version.

cpcramer: I agree that this doesn't tell a story. I'd recommend to update to a more flattering image such as:
image

Zizzamia: We should update for something more interesting, something like about the Logic of components. That's the fun part and super useful.

Screenshot 2024-08-21 at 11 13 15 AM

Zizzamia: Maybe for Swap, just focus on the logic part.
Screenshot 2024-08-21 at 11 15 13 AM

cpcramer: Those images are zoomed so far out you can't read anything

cpcramer: Adam was talking about how embedding the interactive Figma files is easy. So maybe we just wait for that update

Zizzamia: Ok, but what's the story telling we want here? I don't think embedding is the answear. We want to tease the fact we have an OSS Figma, we don't want to distract people.

So howe can we tease without distracting?

@cpcramer again, the question always is, "what's the story we are telling, and how we can be intentional about it."

If we don't know the story, than not even the person reading will know that.

PR #1130, opened 2024-08-21
feat: add `useSendCall` and `useSendCalls` hooks to support call-type transactions in `Transaction` component

0xAlec:
**What changed? Why?**
- scaffold `useSendCall` and `useSendCalls` in the same format as `useWriteContract` and `useWriteContracts` to support calldata in the `Transaction` component

**Notes to reviewers**

**How has it been tested?**
N/A

### DIFF
--- a/.changeset/dry-actors-bathe.md
+++ b/.changeset/dry-actors-bathe.md
@@ -2,4 +2,4 @@
 "@coinbase/onchainkit": patch
 ---
 
-**feat**: `useSendCall` and `useSendCalls` by @0xAlec #1130
+**feat**: add `useSendCall` and `useSendCalls` hooks to support call-type transactions in `Transaction` component by @0xAlec #1130
--- a/.changeset/dry-actors-bathe.md
+++ b/.changeset/dry-actors-bathe.md
@@ -1,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
+
+**feat**: `useSendCall` and `useSendCalls` by @0xAlec #1130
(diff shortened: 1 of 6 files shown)


cpcramer (approved):

abcrane123 (approved):

.changeset/dry-actors-bathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
+
+**feat**: `useSendCall` and `useSendCalls` by @0xAlec #1130
```
Zizzamia, line 5: Let's make sure we are a bit more verbose in what those function do.

The goal of a CHANGELOG is to inform and help the developer understand what's going on.

src/transaction/hooks/useSendCall.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Wagmi hook for single transactions with calldata.
+ * Supports both EOAs and Smart Wallets.
+ * Does not support transaction batching or paymasters.
+ */
+export function useSendCall({
+  setLifeCycleStatus,
+  transactionHashList,
+}: UseSendCallParams) {
+  const { status, sendTransactionAsync, data } = useSendCallWagmi({
```
abcrane123, line 16: what are your thoughts on renaming this to be `sendCallAsync` instead of `sendTransactionAsync` for readability sake

Zizzamia: oh yeah, great point

0xAlec: good call! added the alias

.changeset/dry-actors-bathe.md
0xAlec: done in new diff!

Zizzamia (approved):

PR #1132, opened 2024-08-21
chore: move away from onStart and using only onStatus for advance lifeCycleStatus

Zizzamia:
…eCycleStatus

**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/swap/types.ts
+++ b/src/swap/types.ts
@@ -110,21 +110,15 @@ export type LifeCycleStatus =
       statusData: null;
     }
   | {
-      statusName: 'transactionApproved';
+      statusName: 'erc20Approved';
       statusData: {
-        transactionHash: Address;
+        transactionHash: Hex;
       };
     }
   | {
-      statusName: 'transactionPermit'; // Need to triple check this
+      statusName: 'permit2Approved';
       statusData: {
-        transactionPermit: Address;
-      };
-    }
-  | {
-      statusName: 'transactionPermitApproved'; // Need to triple check this
-      statusData: {
-        transactionHash: Address;
+        transactionHash: Hex;
       };
     }
   | {
--- a/src/swap/types.ts
+++ b/src/swap/types.ts
@@ -110,16 +110,19 @@ export type LifeCycleStatus =
       statusData: null;
     }
   | {
-      statusName: 'erc20Approved';
+      statusName: 'transactionPending';
+      statusData: null;
+    }
+  | {
+      statusName: 'transactionApproved';
       statusData: {
         transactionHash: Hex;
+        transactionType: 'ERC20' | 'Permit2';
       };
     }
   | {
-      statusName: 'permit2Approved';
-      statusData: {
-        transactionHash: Hex;
-      };
+      statusName: 'swapPending';
+      statusData: null;
     }
   | {
       statusName: 'success';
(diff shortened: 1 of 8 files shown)


src/swap/types.ts
```diff
@@ -107,13 +109,40 @@ export type LifeCycleStatus =
...
+      statusData: {
+        transactionPermit: Address;
+      };
+    }
+  | {
+      statusName: 'transactionPermitApproved'; // Need to triple check this
+      statusData: {
+        transactionHash: Address;
+      };
+    }
```
0xAlec, line 129: 3 states for swap - `erc20Approved`, `permit2Approved`, and `success`

for V2 API, it goes `permit2Approved` -> `erc20Approved` -> `success`
for V1 API, it goes `erc20Approved` -> `success`

```suggestion
  | {
      statusName: 'erc20Approved';
      statusData: {
        transactionHash: Hex;
      };
    }
  | {
      statusName: 'permit2Approved';
      statusData: {
        transactionHash: Hex;
      };
    }
```

src/swap/utils/processSwapTransaction.ts
```diff
@@ -44,7 +29,12 @@ export async function processSwapTransaction({
       value: approveTransaction.value,
       data: approveTransaction.data,
     });
-    await Promise.resolve(onStart?.(approveTxHash));
+    setLifecycleStatus({
+      statusName: 'transactionApproved',
```
0xAlec, line 33:
```suggestion
      statusName: useAggregator ? 'erc20Approved' : 'permit2Approved',
```

src/swap/utils/processSwapTransaction.ts
```diff
@@ -76,7 +66,12 @@ export async function processSwapTransaction({
         data: data,
         value: 0n,
       });
-      await Promise.resolve(onStart?.(permitTxnHash));
+      setLifecycleStatus({
+        statusName: 'transactionPermit',
```
0xAlec, line 70:
```suggestion
        statusName: 'erc20Approved',
```

src/swap/utils/processSwapTransaction.ts
```diff
@@ -92,17 +87,24 @@ export async function processSwapTransaction({
...
     data: transaction.data,
   });
-  await Promise.resolve(onStart?.(txHash));
-
+  setLifecycleStatus({
+    statusName: 'transactionPermitApproved',
+    statusData: {
+      transactionHash: txHash, // Check with Alec what this should be
+    },
+  });
```
0xAlec, line 95: we can probably omit this - the only difference between this state and `statusName: 'success'` is the former is when the swap transaction is submitted and the latter is when the swap transaction is confirmed onchain, it should be less than 1 second between both. if developers ask for this we can add a `statusName: 'swapSubmitted'` or similar state

```suggestion

```

src/swap/utils/processSwapTransaction.ts
```diff
@@ -76,7 +66,12 @@ export async function processSwapTransaction({
         data: data,
         value: 0n,
       });
-      await Promise.resolve(onStart?.(permitTxnHash));
+      setLifecycleStatus({
+        statusName: 'erc20Approved',
+        statusData: {
+          transactionPermit: permitTxnHash, // Check with Alec what this should be
```
0xAlec, line 72:
```suggestion
          transactionHash: permitTxnHash,
```

PR #1153, opened 2024-08-23
fix: re-add `error` so `SwapMessage` renders

0xAlec:
**What changed? Why?**
- readd the `error` field on context, and set by the component lifecycle emitters so `SwapMessage` renders

**Notes to reviewers**

**How has it been tested?**
playground
image

### DIFF
--- a/src/swap/components/SwapProvider.tsx
+++ b/src/swap/components/SwapProvider.tsx
@@ -17,6 +17,7 @@ import { useFromTo } from '../hooks/useFromTo';
 import type {
   LifeCycleStatus,
   SwapContextType,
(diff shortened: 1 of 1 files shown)


src/swap/components/SwapProvider.tsx
```diff
@@ -233,6 +236,7 @@ export function SwapProvider({
   ]);
 
   const value = useValue({
+    error,
```
Zizzamia, line 239: This makes me think so much we need to refactor `getErrorMessage`. And this `error` story.

Zizzamia (approved):

PR #1177, opened 2024-08-28
feat: buildPayTransaction utilities in preparation for Pay button

avidreder:
**What changed? Why?**
- Adds utility for making RPC call to hydrate a charge and build a Pay Transaction
- Adds supporting types
- Adds utility and constants for handling errors

**Notes to reviewers**

**How has it been tested?**


src/pay/types.ts
```diff
@@ -0,0 +1,21 @@
+export type HydrateChargeAPIResponse = {
+  id: string;
+  callData: {
+    deadline: string;
+    feeAmount: string;
+    id: string;
+    operator: `0x${string}`;
```
Zizzamia, line 7: When we can, let's use the type from viem.

In this case
```
import type { Address } from 'viem';
````

**Comment by Zizzamia** [comment]
Going forward for all APIs, let's have them live in the `api` module :)

File: src/pay/types.ts (hunk 1 above)
Comment (avidreder) [reply src/pay/types.ts]: Updated

File: src/api/hydrateCharge.ts [hunk 2]
```diff
@@ -0,0 +1,41 @@
...
+import { sendRequest } from '../network/request';
+import { UNCAUGHT_HYDRATE_CHARGE_ERROR_CODE } from '../pay/constants';
+import { getPayErrorCode } from '../pay/utils/getPayErrorCode';
+import type {
+  HydrateChargeAPIParams,
+  HydrateChargeParams,
+  HydrateChargeResponse,
+} from './types';
+
+export async function hydrateCharge({
```
Zizzamia, line 11: @avidreder help me understand, what `hydrateCharge` is?

I read it, and it means nothing to me.

Like how a developer has suppose to use this?

src/api/hydrateCharge.ts
```diff
@@ -0,0 +1,41 @@
...
+  chargeId,
+}: HydrateChargeParams): Promise<HydrateChargeResponse> {
+  try {
+    const res = await sendRequest<
+      HydrateChargeAPIParams,
+      HydrateChargeResponse
+    >(PAY_HYDRATE_CHARGE, [
+      {
+        sender: sender,
+        chainId: 8453,
```
Zizzamia, line 22: for Chain ID, let's use viem.

so it is more explicity like `base.id`

Also, we probably want to get the chain from the config.

So even better this API should have a `chainId` as argument.

src/api/hydrateCharge.ts
```diff
@@ -0,0 +1,41 @@
...
+        chargeId,
+      },
+    ]);
+    if (res.error) {
+      return {
+        code: getPayErrorCode(res.error?.code),
```
Zizzamia, line 28: So code for OnchainKit, it's about the location of the error.
In this case it will be
`AmHCa01` which stands for Api module Hydrate Charge api O1 error

src/api/hydrateCharge.test.ts
```diff
@@ -0,0 +1,153 @@
...
+    const mockAPIParams: HydrateChargeAPIParams = {
+      ...mockParams,
+      chainId: 8453,
+    };
+    const mockResponse = {
+      id: 1,
+      jsonrpc: '2.0',
+      result: {
+        id: testCharge,
+        callData: {
```
Zizzamia, line 84: Should the mock data be stored in a different place?

How can we make those tests a bit easier to read and re-use the mock data where we can.

src/pay/utils/getPayErrorCode.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  GENERAL_PAY_ERROR_CODE,
+  PAY_INVALID_CHARGE_ERROR_CODE,
+  PAY_INVALID_PARAMETER_ERROR_CODE,
+  PAY_TOO_MANY_REQUESTS_ERROR_CODE,
+} from '../constants';
+
+export function getPayErrorCode(errorCode?: number) {
```
Zizzamia, line 8: So it seems we are generating short Error Messages, so let's rename this as `getPayErrorMessage`.

src/api/types.ts
```diff
@@ -79,3 +80,22 @@ export type RawTransactionData = {
 };
 
 export type SwapAPIParams = GetQuoteAPIParams | GetSwapAPIParams;
+
+/**
+ * Note: exported as public Type
+ */
+export type HydrateChargeParams = {
+  sender: Address; // The address of the wallet paying
```
Zizzamia, line 88: in OnchainKit when we refer to the wallet in charge, we call it `address: Address`.

As it simplify the conversation as this is the account address in charge of this operation.

Using the word `sender` put the conversation as the API is the boss, and it says "who is the sender?" but the boss is the developer that using their own wallet, or customer wallet.

So if you say to the developer give me the `sender`, I honestly think "sender who? who is sending what? do I have an amazon package to deliver?" it just doesn't make sense.

So when we tell the story, the developer is at the center of our wording, not the backend.

src/api/hydrateCharge.test.ts
avidreder: Sounds good. I've moved API responses to a fixtures file, along with named constants for inputs for clearer intent.

src/api/hydrateCharge.ts
avidreder: Ok, I've updated error sites to have codes in this format.

src/pay/utils/getPayErrorCode.ts
avidreder: Makes sense to me, renamed this function and constants.

src/api/types.ts
avidreder: I'm fine with having the developer API refer to it as address, and keep sender as an API/Contract concept. Updated.

src/pay/types.ts
```diff
@@ -0,0 +1,23 @@
+import type { Address } from 'viem';
+
+export type HydratedCharge = {
```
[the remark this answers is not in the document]
avidreder: I took a stab at adding comments that will be useful, take a look.

src/api/hydrateCharge.ts
avidreder: Totally agree charge hydration is more of an internal detail than a developer concern. I've renamed this function and supporting code to `buildPayTransaction`, let me know what you think.

src/api/buildPayTransaction.ts
```diff
@@ -0,0 +1,50 @@
...
+} from './types';
+
+export async function buildPayTransaction({
+  address,
+  chainId,
+  chargeId,
+}: BuildPayTransactionParams): Promise<BuildPayTransactionResponse> {
+  if (chainId !== base.id) {
+    return {
+      code: 'AmBPT01', // Api Module Build Pay Transaction Error 01
```
Zizzamia, line 19: nit. Change this to `AmBPTa01`

src/pay/utils/getPayErrorMessage.ts
```diff
@@ -0,0 +1,27 @@
+import {
+  GENERAL_PAY_ERROR_MESSAGE,
+  PAY_INVALID_CHARGE_ERROR_MESSAGE,
+  PAY_INVALID_PARAMETER_ERROR_MESSAGE,
+  PAY_TOO_MANY_REQUESTS_ERROR_MESSAGE,
+  UNCAUGHT_PAY_ERROR_MESSAGE,
+} from '../constants';
+
+export function getPayErrorMessage(errorCode?: number) {
```
Zizzamia, line 9: On nice

src/pay/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const GENERAL_PAY_ERROR_MESSAGE = 'PAY_ERROR';
```
Zizzamia, line 1: I am leaning towards consolidate this into `src/api/constants`.

src/pay/utils/getPayErrorMessage.ts
Zizzamia: Also, let's probably move this into `src/api/utils`

src/api/types.ts
```diff
@@ -79,3 +79,47 @@ export type RawTransactionData = {
...
 
 export type SwapAPIParams = GetQuoteAPIParams | GetSwapAPIParams;
+
+/**
+ * Note: exported as public Type
+ */
+export type BuildPayTransactionParams = {
+  address: Address; // The address of the wallet paying
+  chainId: number; // The Chain ID of the payment Network (only Base is supported)
+  chargeId: string; // The ID of the Commerce Charge to be paid
```
Zizzamia, line 89: Where can I read more about what's a Commerce Charge? Do we have live docs anywhere?

src/pay/constants.ts
avidreder: I think that's fine, though I do worry about that file growing large over time.

Were you picturing component error messages being in a single constants file as well per domain?

Zizzamia: I think many error messages are going to be the exact same between APIs, so having them in one place it will make for us easier to oversee the entire dev experience, and help those API converge into similar language.

src/api/buildPayTransaction.ts
```diff
@@ -0,0 +1,50 @@
...
+    >(PAY_HYDRATE_CHARGE, [
+      {
+        sender: address,
+        chainId: chainId,
+        chargeId,
+      },
+    ]);
+    if (res.error) {
+      return {
+        code: 'AmBPT02', // Api Module Build Pay Transaction Error 02
```
Zizzamia, line 37: `AmBPT02` to `AmBPTa02`

src/api/buildPayTransaction.ts
```diff
@@ -0,0 +1,50 @@
...
+        error: res.error.message,
+        message: getPayErrorMessage(res.error?.code),
+      };
+    }
+    return res.result;
+  } catch (_error) {
+    return {
+      code: 'AmBPT03', // Api Module Build Pay Transaction Error 03
```
Zizzamia, line 45: `AmBPT03` to `AmBPTa03`

src/api/bulidPayTransaction.test.ts
```diff
@@ -0,0 +1,134 @@
...
+      address: MOCK_VALID_PAYER_ADDRESS,
+      chainId: base.id,
+      chargeId: MOCK_VALID_CHARGE_ID,
+    };
+    const mockAPIParams: HydrateChargeAPIParams = {
+      sender: MOCK_VALID_PAYER_ADDRESS,
+      chargeId: MOCK_VALID_CHARGE_ID,
+      chainId: base.id,
+    };
+
```
Zizzamia, line 45: nit. let's avoid spaces inside an `it()` so it's easier read many of them at the same time.

But still have one space between `it()`

src/api/fixtures/buildPayTransactionFixtures.ts
```diff
@@ -0,0 +1,38 @@
+export const MOCK_VALID_CHARGE_ID = '1b03e80d-4e87-46fd-9772-422a1b693fb7';
```
Zizzamia, line 1: Oh yeah, let's have the mocks in the `src/api/mocks.ts` file.

Like we have in https://github.com/coinbase/onchainkit/blob/main/src/swap/mocks.ts

src/api/types.ts
avidreder: This is probably the page I would point to: https://docs.cdp.coinbase.com/commerce-onchain/docs/crypto-payments for an overview of what a charge is (though not great info for this context, we'll have to update it).

Where do you think we should add this link in the code?

Zizzamia: Ok let's come back on this one, mostly i am questioning the word `charge`.

Zizzamia (approved):

Zizzamia:
Great start!!!

PR #1181, opened 2024-08-29
feat: EIP-5792 in `OnchainKitProvider`

0xAlec:
**What changed? Why?**
- add `useCapabilitiesSafe` hook
- add EIP-5792 capabilities to `OnchainKitProvider` for consumption in child components

**Notes to reviewers**

**How has it been tested?**
unit tests and console logging the capabilities on Playground

### DIFF
--- a/src/OnchainKitProvider.tsx
+++ b/src/OnchainKitProvider.tsx
@@ -1,5 +1,4 @@
 import { createContext, useMemo } from 'react';
-import { useAccount } from 'wagmi';
 import { ONCHAIN_KIT_CONFIG, setOnchainKitConfig } from './OnchainKitConfig';
 import { checkHashLength } from './internal/utils/checkHashLength';
 import type { OnchainKitContextType, OnchainKitProviderReact } from './types';
@@ -22,7 +21,6 @@ export function OnchainKitProvider({
   if (schemaId && !checkHashLength(schemaId, 64)) {
     throw Error('EAS schemaId must be 64 characters prefixed with "0x"');
   }
-  const account = useAccount();
   const walletCapabilities = useCapabilitiesSafe({ chain });
 
   const value = useMemo(() => {
@@ -36,7 +34,7 @@ export function OnchainKitProvider({
     };
     setOnchainKitConfig(onchainKitConfig);
     return onchainKitConfig;
-  }, [account, address, chain, schemaId, apiKey, rpcUrl]);
+  }, [address, chain, schemaId, apiKey, rpcUrl]);
 
   return (
     <OnchainKitContext.Provider value={value}>
--- a/src/OnchainKitProvider.tsx
+++ b/src/OnchainKitProvider.tsx
@@ -21,7 +21,7 @@ export function OnchainKitProvider({
   if (schemaId && !checkHashLength(schemaId, 64)) {
     throw Error('EAS schemaId must be 64 characters prefixed with "0x"');
   }
-  const walletCapabilities = useCapabilitiesSafe({ chain });
+  const walletCapabilities = useCapabilitiesSafe({ chainId: chain.id });
 
   const value = useMemo(() => {
     const onchainKitConfig = {
--- a/src/OnchainKitProvider.tsx
+++ b/src/OnchainKitProvider.tsx
@@ -27,10 +27,14 @@ export function OnchainKitProvider({
     const onchainKitConfig = {
       address: address ?? null, // this can maybe be updated to account.address
       apiKey: apiKey ?? null,
-      capabilities: walletCapabilities ?? null,
       chain: chain,
       rpcUrl: rpcUrl ?? null,
       schemaId: schemaId ?? null,
+      walletCapabilities: walletCapabilities ?? {
+        paymasterServiceEnabled: false,
+        atomicBatchEnabled: false,
+        auxiliaryFundsEnabled: false,
+      },
     };
     setOnchainKitConfig(onchainKitConfig);
     return onchainKitConfig;
--- a/src/OnchainKitProvider.tsx
+++ b/src/OnchainKitProvider.tsx
@@ -25,7 +25,7 @@ export function OnchainKitProvider({
 
   const value = useMemo(() => {
     const onchainKitConfig = {
-      address: address ?? null, // this can maybe be updated to account.address
+      address: address ?? null,
       apiKey: apiKey ?? null,
       chain: chain,
       rpcUrl: rpcUrl ?? null,
--- a/src/OnchainKitProvider.tsx
+++ b/src/OnchainKitProvider.tsx
@@ -34,7 +34,7 @@ export function OnchainKitProvider({
     };
     setOnchainKitConfig(onchainKitConfig);
     return onchainKitConfig;
-  }, [address, chain, schemaId, apiKey, rpcUrl]);
+  }, [address, chain, schemaId, apiKey, rpcUrl, walletCapabilities]);
   return (
     <OnchainKitContext.Provider value={value}>
       {children}
--- a/src/OnchainKitProvider.tsx
+++ b/src/OnchainKitProvider.tsx
@@ -2,6 +2,8 @@ import { createContext, useMemo } from 'react';
 import { ONCHAIN_KIT_CONFIG, setOnchainKitConfig } from './OnchainKitConfig';
 import { checkHashLength } from './internal/utils/checkHashLength';
 import type { OnchainKitContextType, OnchainKitProviderReact } from './types';
+import { useAccount } from 'wagmi';
+import { useCapabilitiesSafe } from './useCapabilitiesSafe';
 
 export const OnchainKitContext =
   createContext<OnchainKitContextType>(ONCHAIN_KIT_CONFIG);
@@ -20,17 +22,22 @@ export function OnchainKitProvider({
   if (schemaId && !checkHashLength(schemaId, 64)) {
     throw Error('EAS schemaId must be 64 characters prefixed with "0x"');
   }
+  const account = useAccount();
+  const walletCapabilities = useCapabilitiesSafe({ chain });
+
   const value = useMemo(() => {
     const onchainKitConfig = {
-      address: address ?? null,
+      address: address ?? null, // this can maybe be updated to account.address
       apiKey: apiKey ?? null,
+      capabilities: walletCapabilities ?? null,
       chain: chain,
       rpcUrl: rpcUrl ?? null,
       schemaId: schemaId ?? null,
     };
     setOnchainKitConfig(onchainKitConfig);
     return onchainKitConfig;
-  }, [address, chain, schemaId, apiKey, rpcUrl]);
+  }, [account, address, chain, schemaId, apiKey, rpcUrl]);
+
   return (
     <OnchainKitContext.Provider value={value}>
       {children}
--- a/src/OnchainKitProvider.tsx
+++ b/src/OnchainKitProvider.tsx
@@ -35,7 +35,6 @@ export function OnchainKitProvider({
     setOnchainKitConfig(onchainKitConfig);
     return onchainKitConfig;
   }, [address, chain, schemaId, apiKey, rpcUrl]);
-
   return (
     <OnchainKitContext.Provider value={value}>
       {children}
--- a/src/OnchainKitProvider.tsx
+++ b/src/OnchainKitProvider.tsx
@@ -31,9 +31,9 @@ export function OnchainKitProvider({
       rpcUrl: rpcUrl ?? null,
       schemaId: schemaId ?? null,
       walletCapabilities: walletCapabilities ?? {
-        atomicBatchEnabled: false,
-        auxiliaryFundsEnabled: false,
-        paymasterServiceEnabled: false,
+        hasAtomicBatchEnabled: false,
+        hasAuxiliaryFundsEnabled: false,
+        hasPaymasterServiceEnabled: false,
       },
     };
     setOnchainKitConfig(onchainKitConfig);
--- a/src/OnchainKitProvider.tsx
+++ b/src/OnchainKitProvider.tsx
@@ -31,9 +31,9 @@ export function OnchainKitProvider({
       rpcUrl: rpcUrl ?? null,
       schemaId: schemaId ?? null,
       walletCapabilities: walletCapabilities ?? {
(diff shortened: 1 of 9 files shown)


src/useCapabilitiesSafe.ts
```diff
@@ -0,0 +1,36 @@
...
+export function useCapabilitiesSafe({
+  chain,
+}: {
+  chain: Chain;
+}): WalletCapabilities {
+  const { connector, isConnected } = useAccount();
+
+  // Metamask doesn't support wallet_getCapabilities
+  const isMetamaskWallet = connector?.id === 'io.metamask';
+  const enabled = isConnected && !isMetamaskWallet;
```
0xAlec, line 15: optionally we can also just check for `error` returned from `useCapabilities` but there will still be an error displayed in the console for unsupported RPCs - IMO this is a more scalable solution than manually excluding individual connector IDs

0xAlec: cc @alissacrane-cb on Tanstack query best practices

src/OnchainKitProvider.tsx
```diff
@@ -20,17 +21,20 @@ export function OnchainKitProvider({
...
     throw Error('EAS schemaId must be 64 characters prefixed with "0x"');
   }
+  const walletCapabilities = useCapabilitiesSafe({ chain });
+
   const value = useMemo(() => {
     const onchainKitConfig = {
-      address: address ?? null,
+      address: address ?? null, // this can maybe be updated to account.address
       apiKey: apiKey ?? null,
+      capabilities: walletCapabilities ?? null,
```
0xAlec, line 30: should this default to every value set as `false`?

the type is `WalletCapabilities` (non-optional parameter)

0xAlec: image

0xAlec: this way we can just use the below code in a component without worrying about null `capabilities`

```
const { capabilities } = useOnchainKit()

if (capabilities.paymaster) {
...
}
```

src/useCapabilitiesSafe.ts
```diff
@@ -0,0 +1,36 @@
+import type { Chain } from 'viem';
+import { useAccount } from 'wagmi';
+import { useCapabilities } from 'wagmi/experimental';
+import type { WalletCapabilities } from './types';
+
+export function useCapabilitiesSafe({
+  chain,
```
Zizzamia, line 7: Going forward we are moving away from using `chain`, instead we will have directly `chainId`.

src/useCapabilitiesSafe.ts
```diff
@@ -0,0 +1,36 @@
+import type { Chain } from 'viem';
+import { useAccount } from 'wagmi';
+import { useCapabilities } from 'wagmi/experimental';
+import type { WalletCapabilities } from './types';
+
+export function useCapabilitiesSafe({
```
Zizzamia, line 6: why not just `useCapabilities`? like we did for other similar cases.

src/types.ts
```diff
@@ -45,3 +46,12 @@ export type OnchainKitProviderReact = {
...
 };
+
+/**
+ * Note: exported as public Type
+ */
+export type WalletCapabilities = {
+  paymaster: boolean; // If the wallet supports ERC-4337 Paymasters for gas sponsorship
+  batching: boolean; // If the wallet supports atomic batching of transactions
+  funding: boolean; // If the wallet supports auxiliary funding of accounts (e.g. Magic Spend)
+};
```
Zizzamia, line 57: mmmm if those are boolean, probably we should use the `has` prefix

```
export type WalletCapabilities = {
  hasPaymasterSupport: boolean;
  hasBatchingSupport: boolean;
  hasFundingSupport: boolean;
};
```

src/OnchainKitProvider.tsx
```diff
@@ -20,17 +21,20 @@ export function OnchainKitProvider({
...
       chain: chain,
       rpcUrl: rpcUrl ?? null,
       schemaId: schemaId ?? null,
     };
     setOnchainKitConfig(onchainKitConfig);
     return onchainKitConfig;
-  }, [address, chain, schemaId, apiKey, rpcUrl]);
+  }, [address, chain, schemaId, apiKey, rpcUrl, walletCapabilities]);
```
Zizzamia, line 37: alphabetical

src/OnchainKitProvider.tsx
Zizzamia: > should this default to every value set as false?

yes

src/OnchainKitProvider.tsx
```diff
@@ -20,17 +21,20 @@ export function OnchainKitProvider({
   if (schemaId && !checkHashLength(schemaId, 64)) {
     throw Error('EAS schemaId must be 64 characters prefixed with "0x"');
   }
+  const walletCapabilities = useCapabilitiesSafe({ chain });
+
   const value = useMemo(() => {
     const onchainKitConfig = {
-      address: address ?? null,
+      address: address ?? null, // this can maybe be updated to account.address
```
Zizzamia, line 28: yeah we need to deprecate the `address` story here, but that will require a much bigger PR.

src/useCapabilitiesSafe.ts
```diff
@@ -0,0 +1,36 @@
...
+}): WalletCapabilities {
+  const { connector, isConnected } = useAccount();
+
+  // Metamask doesn't support wallet_getCapabilities
+  const isMetamaskWallet = connector?.id === 'io.metamask';
+  const enabled = isConnected && !isMetamaskWallet;
+
+  const { data: capabilities } = useCapabilities({ query: { enabled } });
+
+  const atomicBatchEnabled =
```
Zizzamia, line 19: I think we can simplify the readability of this code as

```ts
let hasAtomicBatch = false;
let hasPaymasterService = false;
let hasAuxiliaryFunds= false;

if (capabilities &&  capabilities[chain.id]) {
  const capabilitiesByChain = capabilities[chain.id];
  hasAtomicBatch = capabilitiesByChain.atomicBatch?.supported  === true;
  hasPaymasterService = capabilitiesByChain.paymasterService?.supported  === true;
  hasAuxiliaryFunds = capabilitiesByChain.auxiliaryFunds?.supported === true;
}
```

or something like that. Just avoid all the crazy tuples, that are always hard to mantain.

src/useCapabilitiesSafe.test.ts
```diff
@@ -0,0 +1,108 @@
...
+
+describe('useCapabilitiesSafe', () => {
+  const mockChain = {
+    id: 1,
+  };
+
+  beforeEach(() => {
+    vi.resetAllMocks();
+  });
+
```
Zizzamia, line 23: Open question, what are we missing to test?

Mostly what could go wrong? and we should add some negative tests to check crazy scenarios.

src/useCapabilitiesSafe.ts
```diff
@@ -0,0 +1,36 @@
...
+    (capabilities &&
+      capabilities[chain.id]?.paymasterService?.supported === true) ??
+    false;
+  const auxiliaryFundsEnabled =
+    (capabilities &&
+      capabilities[chain.id]?.auxiliaryFunds?.supported === true) ??
+    false;
+
+  return {
+    paymaster: paymasterServiceEnabled,
```
wilsoncusack, line 32: why not just use the names from the ERCs and bubbling these up directly?

src/useCapabilitiesSafe.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export function useCapabilitiesSafe({
+  chain,
+}: {
+  chain: Chain;
+}): WalletCapabilities {
+  const { connector, isConnected } = useAccount();
+
+  // Metamask doesn't support wallet_getCapabilities
+  const isMetamaskWallet = connector?.id === 'io.metamask';
```
wilsoncusack, line 14: This feels like whack-a-mole. Should we just try catch and have an empty object?

src/useCapabilitiesSafe.ts
0xAlec: changed in latest diff

src/useCapabilitiesSafe.ts
0xAlec: need to add handling for wallets that do not support `useCapabilities`

0xAlec: also used as a helper to automatically parse out the capabilities `chainId` we are currently on

src/useCapabilitiesSafe.ts
0xAlec: i just copy pasted the Basenames code for `useCapabilitiesSafe`, just edited to make it more succinct

src/useCapabilitiesSafe.ts
0xAlec: edited it to just bubble up the capabilities names from the ERCs, 

cc @Zizzamia if we still want the `has` or `is` prefix pattern

I've also renamed the context key `capabilities` -> `walletCapabilities` more increased clarity

src/useCapabilitiesSafe.ts
0xAlec: agreed - edited to return `false` for every field depending on the TanStack `error` field from `useCapabilities`

a console error is now emitted from the wallet extension which was previously suppressed by this code

0xAlec: agreed - edited to return `false` for every field depending on the TanStack `error` field from `useCapabilities`

a console error is now emitted from the wallet extension which was previously suppressed by this code

0xAlec: this code was from Leo on the Basenames team:

```
It would just throw an error
Not breaking the app, but potentially breaking code execution order
```

I don't think this would be the case since we're handling the error here unless there's application specific logic within the extension that breaks on failed `wallet_getCapabilities` requests but I think that's unlikely

src/useCapabilitiesSafe.ts
Zizzamia: Year using the the ERCs name was a good call.

I think to keep the same standard we have around OchainKit for boolean using the prefix `is` or `has`, I think it's important to stick with that.

Based on the context, seems like `hasSomething` seems the right prefix.

src/OnchainKitProvider.tsx
Zizzamia: Are we keeping this comment?
```
this can maybe be updated to account.address
```

src/useCapabilitiesSafe.ts
```diff
@@ -0,0 +1,29 @@
+import { useAccount } from 'wagmi';
+import { useCapabilities } from 'wagmi/experimental';
+import type { WalletCapabilities } from './types';
+
+export function useCapabilitiesSafe({
+  chainId,
+}: {
```
Zizzamia, line 7: Let's create his own type here.

src/types.ts
```diff
@@ -45,3 +46,12 @@ export type OnchainKitProviderReact = {
   rpcUrl?: string;
   schemaId?: EASSchemaUid;
 };
+
+/**
+ * Note: exported as public Type
```
Zizzamia, line 51: I don't see this exported in the `index.ts`

src/useCapabilitiesSafe.ts
```diff
@@ -0,0 +1,29 @@
...
+  chainId: number;
+}): WalletCapabilities {
+  const { isConnected } = useAccount();
+
+  const { data: capabilities, error } = useCapabilities({
+    query: { enabled: isConnected },
+  });
```
Zizzamia, line 14: I believe you mentioned this if fails logs an error.

I think the all point of this wrapper, is to improve the developer experience, and have those error being gone. Otherwise folks will ping us over Twitter and say, "I see an error, did I do something wrong?", when the reality is an exptected error.

So ideally we want to get into the habit that errors that are predicted, and the user can't do nothing about it, we hide them with a try/catch.

@0xAlec what do you think?

src/useCapabilitiesSafe.ts
0xAlec: fixed in latest diff

src/types.ts
0xAlec: fixed in latest diff

src/useCapabilitiesSafe.ts
```diff
@@ -0,0 +1,27 @@
+import { useAccount } from 'wagmi';
+import { useCapabilities } from 'wagmi/experimental';
+import type { UseCapabilitiesSafeParams, WalletCapabilities } from './types';
+
+export function useCapabilitiesSafe({
+  chainId,
+}: UseCapabilitiesSafeParams): WalletCapabilities {
+  const { isConnected } = useAccount();
```
Zizzamia, line 8: Btw as follow up PR, check with @abcrane123 if this will cause a SSR issue.

Zizzamia (approved):

src/useCapabilitiesSafe.ts
```diff
@@ -0,0 +1,27 @@
...
+      hasPaymasterService: false,
+      hasAtomicBatch: false,
+      hasAuxiliaryFunds: false,
+    };
+  }
+
+  return {
+    hasPaymasterService: capabilities[chainId]?.paymasterService?.supported,
+    hasAtomicBatch: capabilities[chainId]?.atomicBatch?.supported,
+    hasAuxiliaryFunds: capabilities[chainId]?.auxiliaryFunds?.supported,
```
alessey, line 25: if these can be undefined, should they be cast to a boolean?

0xAlec: should be fine per [this test](https://github.com/coinbase/onchainkit/pull/1181/files#diff-c09e9ab67f227b7b90d1763987dfcaa7479e344e463d7030c4c4d178ee4703f8R73-R83)?

alessey: that just tests if !capabilities though right?  is it possible for useCapabilities to return some of the capabilities, but not all?

0xAlec: good catch :) fixed in the latest diff!

alessey (approved):

Zizzamia (approved):

src/useCapabilitiesSafe.ts
abcrane123: i don't think this will be an issue

abcrane123 (approved):

PR #1183, opened 2024-08-29
fix: reset lifecycle status after success

0xAlec:
**What changed? Why?**
- reset lifeCycleStatus to `init` after a swap `success`, otherwise the inputs will clear in an infinite loop

**Notes to reviewers**

**How has it been tested?**
playground

### DIFF
--- a/.changeset/quick-scissors-hug.md
+++ b/.changeset/quick-scissors-hug.md
@@ -1,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
+
+-**patch**: reset lifecycle status after swap success. by @0xAlec #1183
--- a/src/swap/components/SwapProvider.tsx
+++ b/src/swap/components/SwapProvider.tsx
@@ -93,6 +93,19 @@ export function SwapProvider({
     }
     // Emit Status
     onStatus?.(lifeCycleStatus);
+
+    // Reset lifecycle status after a success
+    if (
+      lifeCycleStatus.statusName !== 'init' &&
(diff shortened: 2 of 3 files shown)


cpcramer (approved):

src/swap/components/SwapProvider.tsx
```diff
@@ -93,6 +93,18 @@ export function SwapProvider({
     }
     // Emit Status
     onStatus?.(lifeCycleStatus);
+    // Reset lifecycle status after a success
+    if (
+      lifeCycleStatus.statusName !== 'init' &&
```
abcrane123, line 98: we actually don't need that first one because it cannot be `init` if it is `success`

.changeset/quick-scissors-hug.md
```diff
@@ -0,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
+
+-**patch**: reset lifecycle status after swap success. by @0xAlec #1183
```
Zizzamia, line 5: `patch` is the wrong word. Something to fix in the next PR.

src/swap/components/SwapProvider.tsx
```diff
@@ -93,6 +93,18 @@ export function SwapProvider({
...
+      lifeCycleStatus.statusName === 'success'
+    ) {
+      setLifeCycleStatus({
```
Zizzamia, line 101: This is BAD, do not run `setLifeCycleStatus` inside this hook, it will potentially trigger an infit-loop.

PR #1189, opened 2024-08-29
chore: Add Swap Settings v1 api support

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/api/buildSwapTransaction.ts
```diff
@@ -45,6 +45,17 @@ export async function buildSwapTransaction(
     };
   }
```
Zizzamia, line 47: Let's remove those extra spaces.
Let's add tests.

src/api/buildSwapTransaction.ts
```diff
@@ -45,6 +45,17 @@ export async function buildSwapTransaction(
...
 
+  // Adjust slippage for V1 API (aggregator)
+  // V1 expects slippage in tenths of a percent (e.g., 30 = 3%)
+  if (params.useAggregator && params.maxSlippage) {
+    apiParams = {
+      slippagePercentage: (
+        Number.parseFloat(params.maxSlippage) * 10
+      ).toString(),
+      ...apiParams,
```
Zizzamia, line 55: Why are we doing this? this seems odd.

You just can do
`apiParams.slippagePercentage = potato`
no need all the extra code

cpcramer: Updated to:
```

    let slippage = Number(params.maxSlippage);
  
    slippage = slippage * 10;
    
    apiParams.slippagePercentage = slippage.toString();
```

src/api/buildSwapTransaction.test.ts
```diff
@@ -283,7 +283,7 @@ describe('buildSwapTransaction', () => {
     expect(sendRequest).toHaveBeenCalledTimes(1);
     expect(sendRequest).toHaveBeenCalledWith(CDP_GET_SWAP_TRADE, [
       {
-        slippagePercentage: '3',
```
cpcramer, line 286: @0xAlec Has this test been broken? It seems like we've been calling the v1 api with the wrong slippagePercentage (not multiplying the value by 10?)

src/api/buildSwapTransaction.test.ts
```diff
@@ -201,7 +201,7 @@ describe('buildSwapTransaction', () => {
 
   it('should return a swap with an approve transaction', async () => {
     const mockParams = {
-      useAggregator: true,
```
cpcramer, line 204: This is a broken test, useAggregator should be set to false.

src/api/buildSwapTransaction.test.ts
cpcramer: confirmed that this test has been broken

src/api/buildSwapTransaction.ts
```diff
@@ -30,19 +30,19 @@ export async function buildSwapTransaction(
...
   }
-  let apiParams = apiParamsOrError as SwapAPIParams;
+  const apiParams = apiParamsOrError as SwapAPIParams;
 
   if (!params.useAggregator) {
-    apiParams = {
-      v2Enabled: true,
-      ...apiParams,
-    };
+    apiParams.v2Enabled = true;
```
Zizzamia, line 36: yeah sorry, my bad, let's go back to the 
```
{
      v2Enabled: true,
      ...apiParams,
    };
```
approach

cpcramer: done

cpcramer:
```
  let apiParams = apiParamsOrError as SwapAPIParams;
  if (!params.useAggregator) {
    apiParams = {
      v2Enabled: true,
      ...apiParams,
    };
  }
  if (params.maxSlippage) {
    let slippagePercentage = params.maxSlippage;
    // Adjust slippage for V1 API (aggregator)
    // V1 expects slippage in tenths of a percent (e.g., 30 = 3%)
    if (params.useAggregator) {
      slippagePercentage = (Number(params.maxSlippage) * 10).toString();
    }
    apiParams = {
      slippagePercentage,
      ...apiParams,
    };
  }
```

src/api/buildSwapTransaction.ts
```diff
@@ -39,8 +39,14 @@ export async function buildSwapTransaction(
     };
   }
   if (params.maxSlippage) {
+    let slippagePercentage = params.maxSlippage;
+    // Adjust slippage for V1 API (aggregator)
+    // V1 expects slippage in tenths of a percent (e.g., 30 = 3%)
+    if (params.useAggregator) {
```
Zizzamia, line 45: I still don't follow, is `useAggregator: true` what everyone uses at the moment?

If that so, we should not change this code.

cpcramer: There are two different APIs, V1 [wallet/swap-service](https://github.cbhq.net/cloud/swap-router) and V2 [cloud/swap-router](https://github.cbhq.net/cloud/swap-router). 

V1:
- `useAggregator: true`
- Expects slippage in tenths of a percent (e.g., 30 = 3%) which is why we must multiply maxSlippage by 10 before sending the request. 

V2:
- `useAggregator: false`
- Expects slippage in a percent, which is why we do not have to adjust the maxSlippage value.

PartyDOA is the only known customer who is using the maxSlippage parameter.

Zizzamia (approved):

PR #1191, opened 2024-08-30
chore: Add Swap Settings Slippage Title

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/swap/components/SwapSettingsSlippageTitle.tsx
```diff
@@ -0,0 +1,18 @@
...
+import type { SwapSettingsSlippageTitleReact } from '../types';
+
+export function SwapSettingsSlippageTitle({
+  children,
+  className,
+}: SwapSettingsSlippageTitleReact) {
+  return (
+    <h3
+      className={cn(
+        'mb-2 font-semibold text-base text-gray-950 leading-normal dark:text-gray-50',
```
Zizzamia, line 11: `dark` color should be automatic, why is here custom?

Create this fork in the theme, make harder for developer to do custom themes at global level.

cpcramer: Ah great point, I've removed `text-gray-950` and `dark:text-gray-50` and i've added `--text-ock-foreground`

Zizzamia (approved):

PR #1193, opened 2024-08-30
docs: consistent titles

rin-st:
**What changed? Why?**
Added backticks to titles where it should be

Example:
Before:
image

After:
image

**Notes to reviewers**

**How has it been tested?**


Zizzamia (approved):
Brilliant!!!

Love those little polish

PR #1196, opened 2024-08-30
feat: Swap Settings Slippage Layout 

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/swap/components/SwapSettingsSlippageInput.tsx
```diff
@@ -3,15 +3,20 @@ import { background, border, cn, color, pressable } from '../../styles/theme';
 import type { SwapSettingsSlippageInputReact } from '../types';
 import { useSwapContext } from './SwapProvider';
 
-const slippageSettings = ['Auto', 'Custom'];
+const SLIPPAGE_SETTINGS = {
```
cpcramer, line 6: Input changes from this PR https://github.com/coinbase/onchainkit/pull/1206

Zizzamia (approved):
Looks good, general point for later is around better HTML semantic.

cpcramer:
> Looks good, general point for later is around better HTML semantic.

I always default to using `Divs`, we be more deliberate around selecting the appropriate element!

PR #1200, opened 2024-08-31
Swap doc improvements

mykcryptodev:
**What changed? Why?**
- The goal of this PR is to make the Swap component instantly usable after copy/pasting from docs with no errors
- Added missing imports for Avatar and Swap Message
- Added images for tokens to fix type errors in IDE
- Placed the tokens into the swappableTokens const so that the component is usable after copy/pasting

**Notes to reviewers**
- Open to suggestions about which img urls to use for tokens

**How has it been tested?**
- Visually on localhost
- Copy pasted the code and ran it without errors


Zizzamia (approved):
Very smooth, very mindful

Love those, keep them coming @mykcryptodev

PR #1209, opened 2024-09-04
Commerce pay button

0xAlec:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


Zizzamia:
This PR is disconnected from the work @avidreder already did, and how we aligned.
A few top of mind points:
1. the components will live inside `pay`
2. the api will live inside `api`, and the work for that already started.

PR #1210, opened 2024-09-04
feat: Add Swap Settings

cpcramer:
**What changed? Why?**
Add `SwapSettings` sub-component - we are not displaying it yet.

Update `SwapSettingsSlippageInput`
- Fix a bug to allow the input field to clear.
- Fix a bug where closing and opening the Slippage dropdown would reset state. 

Add `maxSlippage` field to `statusData` in every `LifeCycleStatus` except `Error`.
**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/swap/components/Swap.tsx
+++ b/src/swap/components/Swap.tsx
@@ -7,7 +7,6 @@ import { SwapAmountInput } from './SwapAmountInput';
 import { SwapButton } from './SwapButton';
 import { SwapMessage } from './SwapMessage';
 import { SwapProvider } from './SwapProvider';
-import { SwapSettings } from './SwapSettings';
 import { SwapToggleButton } from './SwapToggleButton';
 
 export function Swap({
@@ -19,17 +18,15 @@ export function Swap({
   onSuccess,
   title = 'Swap',
 }: SwapReact) {
-  const { inputs, toggleButton, swapButton, swapMessage, swapSettings } =
-    useMemo(() => {
-      const childrenArray = Children.toArray(children);
-      return {
-        inputs: childrenArray.filter(findComponent(SwapAmountInput)),
-        toggleButton: childrenArray.find(findComponent(SwapToggleButton)),
-        swapButton: childrenArray.find(findComponent(SwapButton)),
-        swapMessage: childrenArray.find(findComponent(SwapMessage)),
-        swapSettings: childrenArray.find(findComponent(SwapSettings)),
-      };
-    }, [children]);
+  const { inputs, toggleButton, swapButton, swapMessage } = useMemo(() => {
+    const childrenArray = Children.toArray(children);
+    return {
+      inputs: childrenArray.filter(findComponent(SwapAmountInput)),
+      toggleButton: childrenArray.find(findComponent(SwapToggleButton)),
+      swapButton: childrenArray.find(findComponent(SwapButton)),
+      swapMessage: childrenArray.find(findComponent(SwapMessage)),
+    };
+  }, [children]);
 
   const isMounted = useIsMounted();
 
@@ -37,6 +34,7 @@ export function Swap({
   if (!isMounted) {
     return null;
   }
+
   return (
     <SwapProvider
       experimental={experimental}
@@ -52,11 +50,8 @@ export function Swap({
         )}
         data-testid="ockSwap_Container"
       >
-        <div className="mb-4 flex items-center justify-between">
-          <h3
-            className={cn(text.title3, 'text-inherit')}
-            data-testid="ockSwap_Title"
-          >
+        <div className="mb-4">
+          <h3 className={text.title3} data-testid="ockSwap_Title">
             {title}
           </h3>
         </div>
--- a/src/swap/components/Swap.tsx
+++ b/src/swap/components/Swap.tsx
@@ -37,7 +37,6 @@ export function Swap({
   if (!isMounted) {
     return null;
   }
-
   return (
     <SwapProvider
       experimental={experimental}
@@ -60,7 +59,7 @@ export function Swap({
           >
             {title}
           </h3>
-          <div>{swapSettings}</div>
+          {swapSettings}
         </div>
         {inputs[0]}
         <div className="relative h-1">{toggleButton}</div>
--- a/src/swap/components/Swap.tsx
+++ b/src/swap/components/Swap.tsx
@@ -59,7 +59,6 @@ export function Swap({
           >
             {title}
           </h3>
-          {swapSettings}
         </div>
         {inputs[0]}
         <div className="relative h-1">{toggleButton}</div>
--- a/src/swap/components/Swap.tsx
+++ b/src/swap/components/Swap.tsx
@@ -7,6 +7,7 @@ import { SwapAmountInput } from './SwapAmountInput';
 import { SwapButton } from './SwapButton';
 import { SwapMessage } from './SwapMessage';
 import { SwapProvider } from './SwapProvider';
+import { SwapSettings } from './SwapSettings';
 import { SwapToggleButton } from './SwapToggleButton';
 
 export function Swap({
@@ -18,15 +19,17 @@ export function Swap({
   onSuccess,
   title = 'Swap',
 }: SwapReact) {
-  const { inputs, toggleButton, swapButton, swapMessage } = useMemo(() => {
-    const childrenArray = Children.toArray(children);
-    return {
-      inputs: childrenArray.filter(findComponent(SwapAmountInput)),
-      toggleButton: childrenArray.find(findComponent(SwapToggleButton)),
-      swapButton: childrenArray.find(findComponent(SwapButton)),
-      swapMessage: childrenArray.find(findComponent(SwapMessage)),
-    };
-  }, [children]);
+  const { inputs, toggleButton, swapButton, swapMessage, swapSettings } =
+    useMemo(() => {
+      const childrenArray = Children.toArray(children);
+      return {
+        inputs: childrenArray.filter(findComponent(SwapAmountInput)),
+        toggleButton: childrenArray.find(findComponent(SwapToggleButton)),
+        swapButton: childrenArray.find(findComponent(SwapButton)),
+        swapMessage: childrenArray.find(findComponent(SwapMessage)),
+        swapSettings: childrenArray.find(findComponent(SwapSettings)),
+      };
+    }, [children]);
 
   const isMounted = useIsMounted();
 
@@ -50,10 +53,14 @@ export function Swap({
         )}
         data-testid="ockSwap_Container"
       >
-        <div className="mb-4">
-          <h3 className={text.title3} data-testid="ockSwap_Title">
+        <div className="mb-4 flex items-center justify-between">
+          <h3
+            className={cn(text.title3, 'text-inherit')}
+            data-testid="ockSwap_Title"
+          >
             {title}
           </h3>
+          <div>{swapSettings}</div>
         </div>
         {inputs[0]}
         <div className="relative h-1">{toggleButton}</div>
--- a/src/swap/components/SwapProvider.tsx
+++ b/src/swap/components/SwapProvider.tsx
@@ -46,7 +46,7 @@ export function SwapProvider({
   const { address } = useAccount();
   // Feature flags
   const { useAggregator } = experimental;
-
+  const maxSlippage = experimental.maxSlippage || 3;
   // Core Hooks
   const config = useConfig();
   const [loading, setLoading] = useState(false);
@@ -56,6 +56,7 @@ export function SwapProvider({
     statusName: 'init',
     statusData: {
       isMissingRequiredField: true,
+      maxSlippage,
     },
   }); // Component lifecycle
   const [hasHandledSuccess, setHasHandledSuccess] = useState(false);
@@ -120,10 +121,11 @@ export function SwapProvider({
         statusName: 'init',
         statusData: {
           isMissingRequiredField: false,
+          maxSlippage,
         },
       });
     }
-  }, [hasHandledSuccess, lifeCycleStatus.statusName]);
+  }, [hasHandledSuccess, lifeCycleStatus.statusName, maxSlippage]);
 
   const handleToggle = useCallback(() => {
     from.setAmount(to.amount);
@@ -153,6 +155,7 @@ export function SwapProvider({
           statusData: {
             amountFrom: from.amount,
             amountTo: to.amount,
+            maxSlippage,
             tokenFrom: from.token,
             tokenTo: to.token,
             // token is missing
@@ -175,6 +178,7 @@ export function SwapProvider({
           // amount is irrelevant
           amountFrom: type === 'from' ? amount : '',
           amountTo: type === 'to' ? amount : '',
+          maxSlippage,
           tokenFrom: from.token,
           tokenTo: to.token,
           // when fetching quote, the destination
@@ -215,6 +219,7 @@ export function SwapProvider({
           statusData: {
             amountFrom: type === 'from' ? amount : formattedAmount,
             amountTo: type === 'to' ? amount : formattedAmount,
+            maxSlippage,
             tokenFrom: from.token,
             tokenTo: to.token,
             // if quote was fetched successfully, we
@@ -236,7 +241,7 @@ export function SwapProvider({
         destination.setLoading(false);
       }
     },
-    [from, experimental.maxSlippage, to, useAggregator],
+    [from, maxSlippage, to, useAggregator],
   );
 
   const handleSubmit = useCallback(async () => {
@@ -247,6 +252,7 @@ export function SwapProvider({
       statusName: 'init',
       statusData: {
         isMissingRequiredField: false,
+        maxSlippage,
       },
     });
 
@@ -276,6 +282,7 @@ export function SwapProvider({
         setLifeCycleStatus,
         swapTransaction: response,
         useAggregator,
+        maxSlippage,
       });
 
       // TODO: refresh balances
@@ -297,10 +304,10 @@ export function SwapProvider({
     config,
     from.amount,
     from.token,
+    maxSlippage,
     sendTransactionAsync,
     to.token,
     useAggregator,
-    experimental.maxSlippage,
   ]);
 
   const value = useValue({
--- a/src/swap/components/SwapProvider.tsx
+++ b/src/swap/components/SwapProvider.tsx
@@ -193,7 +193,7 @@ export function SwapProvider({
           amountReference: 'from',
           from: source.token,
           to: destination.token,
-          maxSlippage: experimental.maxSlippage?.toString(),
+          maxSlippage: maxSlippage.toString(),
           useAggregator,
         });
         // If request resolves to error response set the quoteError
@@ -263,7 +263,7 @@ export function SwapProvider({
         from: from.token,
         to: to.token,
         useAggregator,
-        maxSlippage: experimental.maxSlippage?.toString(),
+        maxSlippage: maxSlippage.toString(),
       });
       if (isSwapError(response)) {
         setLifeCycleStatus({
--- a/src/swap/components/SwapProvider.tsx
+++ b/src/swap/components/SwapProvider.tsx
@@ -280,7 +280,7 @@ export function SwapProvider({
       }
       await processSwapTransaction({
         config,
-        maxSlippage,
+        lifeCycleStatus,
         sendTransactionAsync,
         setLifeCycleStatus,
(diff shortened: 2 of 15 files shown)


src/swap/index.ts
```diff
@@ -20,9 +21,9 @@ export type {
   SwapQuote,
   SwapReact,
   SwapSettingsReact,
+  SwapSettingsSlippageDescriptionReact,
```
Zizzamia, line 24: nice

src/swap/components/SwapSettingsSlippageInput.tsx
```diff
@@ -33,9 +46,13 @@ export function SwapSettingsSlippageInput({
   // Parses the input and updates slippage if valid
   const handleSlippageChange = useCallback(
     (newSlippage: string) => {
-      const newSlippageNumber = Number.parseFloat(newSlippage);
-      if (!Number.isNaN(newSlippageNumber)) {
-        updateSlippage(newSlippageNumber);
+      if (newSlippage === '') {
+        setSlippage(0);
```
Zizzamia, line 50: A cleaner way to write code, it's to add `return;` if the logic is over.
So that you don't need the final `else`.

src/swap/components/SwapSettingsSlippageInput.tsx
```diff
@@ -33,9 +46,13 @@ export function SwapSettingsSlippageInput({
   // Parses the input and updates slippage if valid
   const handleSlippageChange = useCallback(
     (newSlippage: string) => {
-      const newSlippageNumber = Number.parseFloat(newSlippage);
-      if (!Number.isNaN(newSlippageNumber)) {
-        updateSlippage(newSlippageNumber);
+      if (newSlippage === '') {
```
Zizzamia, line 49: Make sure to add a line of comment to explain when slippage is empty

src/swap/components/SwapSettingsSlippageInput.tsx
```diff
@@ -12,10 +12,23 @@ export function SwapSettingsSlippageInput({
...
+      return lifeCycleStatus.statusData.maxSlippage;
+    }
+    return defaultSlippage;
+  };
+
+  // TODO: Add comment about how we do these checks so the dropdown opens to the correct slippage value and button that it left off on
+  const [slippage, setSlippage] = useState(getMaxSlippage());
   const [slippageSetting, setSlippageSetting] = useState(
-    SLIPPAGE_SETTINGS.AUTO,
+    getMaxSlippage() === defaultSlippage
```
Zizzamia, line 29: I don't follow when this happens?

src/swap/components/SwapSettingsSlippageInput.tsx
```diff
@@ -12,10 +12,23 @@ export function SwapSettingsSlippageInput({
...
 }: SwapSettingsSlippageInputReact) {
-  const { setLifeCycleStatus } = useSwapContext();
-  const [slippage, setSlippage] = useState(defaultSlippage);
+  const { setLifeCycleStatus, lifeCycleStatus } = useSwapContext();
+  const getMaxSlippage = () => {
+    if (
+      !!lifeCycleStatus.statusData &&
+      'maxSlippage' in lifeCycleStatus.statusData
+    ) {
+      return lifeCycleStatus.statusData.maxSlippage;
```
Zizzamia, line 21: Two points here:
1. why are we checking if `statusData` is not there? should always be there.
2. we are making the logic more complex because `maxSlippage` is not in all state, so let's have it in all state, so the overl logic is simpler.

src/swap/components/Swap.tsx
```diff
@@ -50,10 +52,14 @@ export function Swap({
...
-        <div className="mb-4">
-          <h3 className={text.title3} data-testid="ockSwap_Title">
+        <div className="mb-4 flex items-center justify-between">
+          <h3
+            className={cn(text.title3, 'text-inherit')}
+            data-testid="ockSwap_Title"
+          >
             {title}
           </h3>
+          {swapSettings}
```
Zizzamia, line 62: Let's add this later, in the next PR. Otherwise this will block the merging of this one.

src/swap/components/SwapSettingsSlippageInput.tsx
cpcramer: This is overly complicated because I have to have these checks otherwise I get errors:

This is all to avoid Typescript errors. For example, if I just have:

` const maxSlippage = String(lifeCycleStatus.statusData.maxSlippage)`

Then I'll see the following errors: 

```
-  'lifeCycleStatus.statusData' is possibly 'null'.ts(18047)

-  'lifeCycleStatus.statusData' is possibly 'null'.ts(18047)

-  Property 'maxSlippage' does not exist on type 'SwapError | { isMissingRequiredField: boolean; } | { amountFrom: string; amountTo: string; tokenFrom?: Token | undefined; tokenTo?: Token | undefined; isMissingRequiredField: boolean; } | { ...; } | { ...; } | { ...; }'.Property 'maxSlippage' does not exist on type 'SwapError'.ts(2339)
```

src/swap/components/SwapSettingsSlippageInput.tsx
cpcramer: Nice tip, I like that

src/swap/components/Swap.tsx
```diff
@@ -18,23 +19,24 @@ export function Swap({
...
-  }, [children]);
+  const { inputs, toggleButton, swapButton, swapMessage, swapSettings } =
+    useMemo(() => {
+      const childrenArray = Children.toArray(children);
+      return {
+        inputs: childrenArray.filter(findComponent(SwapAmountInput)),
+        toggleButton: childrenArray.find(findComponent(SwapToggleButton)),
+        swapButton: childrenArray.find(findComponent(SwapButton)),
+        swapMessage: childrenArray.find(findComponent(SwapMessage)),
+        swapSettings: childrenArray.find(findComponent(SwapSettings)),
```
cpcramer, line 30: We are not rendering `SwapSettings` here yet, so I can remove this as well + the import

src/swap/components/SwapSettingsSlippageInput.tsx
cpcramer: This was to fix a bug when that dropdown was opened and closed it would always reopen to the state `Auto` which would then set the slippage to the default value of 3. 

This code checks if we have updated the slippage value already, if yes then we know that we are in the `CUSTOM` slippage_setting and reopen the dropdown to that state.

src/swap/components/SwapSettingsSlippageInput.tsx
cpcramer: Two possibilities for adding maxSlippage in all states:

Add in every lifecycle statusData:
`    const maxSlippage = lifeCycleStatus.statusData.maxSlippage`
```
export type LifeCycleStatus =
  | {
      statusName: 'init';
      statusData: {
        isMissingRequiredField: boolean;
        maxSlippage: number;
      };
    }
  | {
      statusName: 'error';
      statusData: {
        maxSlippage: number;
        SwapError: SwapError,
      } 
    }
  | {
      statusName: 'amountChange';
      statusData: {
        amountFrom: string;
        amountTo: string;
        maxSlippage: number;
        tokenFrom?: Token;
        tokenTo?: Token;
        isMissingRequiredField: boolean;
      };
    }
  | {
      statusName: 'slippageChange';
      statusData: {
        maxSlippage: number;
      };
    }
  | {
      statusName: 'transactionPending';
      statusData: {
        maxSlippage: number;
      };
    }
  | {
      statusName: 'transactionApproved';
      statusData: {
        maxSlippage: number;
        transactionHash: Hex;
        transactionType: 'ERC20' | 'Permit2';
      };
    }
  | {
      statusName: 'success';
      statusData: {
        maxSlippage: number;
        transactionReceipt: TransactionReceipt;
      };
    };
```

cpcramer: option 2:

export type LifeCycleStatus =
  | {
      statusName: 'init';
      statusData: {
        isMissingRequiredField: boolean;
      };
      maxSlippage: number;
    }
  | {
      statusName: 'error';
      statusData: SwapError;
      maxSlippage: number;
    }
  | {
      statusName: 'amountChange';
      statusData: {
        amountFrom: string;
        amountTo: string;
        tokenFrom?: Token;
        tokenTo?: Token;
        isMissingRequiredField: boolean;
      };
      maxSlippage: number;
    }
  | {
      statusName: 'transactionPending';
      statusData: null;
      maxSlippage: number;
    }
  | {
      statusName: 'transactionApproved';
      statusData: {
        transactionHash: Hex;
        transactionType: 'ERC20' | 'Permit2';
      };
      maxSlippage: number;
    }
  | {
      statusName: 'success';
      statusData: {
        transactionReceipt: TransactionReceipt;
      };
      maxSlippage: number;
    };

Zizzamia: Add everywhere, but not in the Error state.

src/swap/components/Swap.tsx
Zizzamia: yup

src/swap/components/SwapSettingsSlippageInput.tsx
cpcramer: Nice, still need to check for the error state but much cleaner:

```
  const getMaxSlippage = () => {
    if (lifeCycleStatus.statusName !== 'error') {
      return lifeCycleStatus.statusData.maxSlippage;
    }
    return defaultSlippage;
  };
```

Zizzamia: dope

src/swap/components/SwapSettingsSlippageInput.tsx
Zizzamia: ok, just know this condition `getMaxSlippage() === defaultSlippage` is literally fired at every re-render, which makes this an expensive computation.

You always want to think how to have `useEffect` to avoid those computation cost.

Zizzamia: Pretty much never run a function `potato()` in the component without making sure is bounded with useCallback or useEffect.

Otherwise the cost computation goes up

cpcramer: Ah makes sense, updated to wrap getMaxSlippage with useCallback

src/swap/utils/processSwapTransaction.ts
```diff
@@ -24,7 +25,9 @@ export async function processSwapTransaction({
   if (approveTransaction?.data) {
     setLifeCycleStatus({
       statusName: 'transactionPending',
-      statusData: null,
+      statusData: {
+        maxSlippage,
+      },
```
cpcramer, line 30: I was getting build errors in this file post adding the maxSlippage field to lifecycle statusData. I had to update the input params and add the maxSlippage field to all setLifeCycleStatus calls

src/swap/components/SwapProvider.tsx
```diff
@@ -46,7 +46,7 @@ export function SwapProvider({
   const { address } = useAccount();
   // Feature flags
   const { useAggregator } = experimental;
-
+  const maxSlippage = experimental.maxSlippage || 3;
```
cpcramer, line 49: I was getting build errors post adding the `maxSlippage` field to lifecycle statusData. I had to update the input params and add the maxSlippage field to all setLifeCycleStatus calls

Zizzamia: This this should be
```
const { maxSlippage, setMaxSlippage } = useState(experimental.maxSlippage || 3);
```

That's because in the next PRs you are going to set the value of `maxSlippage`.

src/swap/components/SwapSettingsSlippageInput.tsx
```diff
@@ -12,10 +12,21 @@ export function SwapSettingsSlippageInput({
   className,
   defaultSlippage = 3,
```
Zizzamia, line 13: Something I am reflecting more and more, is where `defaultSlippage` should be set.

If at top level `Swap` component. Or in this component.

The reason is that we need a default value in the `SwapProvider`, but we actually set it here.

So here an open question, what's the developer experience we want to offer? Counting we need to set the slippage but also need to have that value as default for the input.

src/swap/utils/processSwapTransaction.ts
```diff
@@ -13,6 +13,7 @@ export async function processSwapTransaction({
   setLifeCycleStatus,
   swapTransaction,
   useAggregator,
+  maxSlippage,
```
Zizzamia, line 16: alphabetical order

src/swap/components/SwapProvider.tsx
```diff
@@ -276,6 +282,7 @@ export function SwapProvider({
         setLifeCycleStatus,
         swapTransaction: response,
         useAggregator,
+        maxSlippage,
```
Zizzamia, line 285: alphabetical order

Zizzamia: Also, I don't think is clean to just pass the `maxSlippage`. Like think 7 steps ahead, `statusData` will contain other info that need to persist inside `processSwapTransaction` when `setLifeCycleStatus` is called.

So we should pass `lifeCycleStatus`, inside `processSwapTransaction`.

So that we can persit the info we need from `statusData` when we recall `setLifeCycleStatus`.

src/swap/components/SwapSettingsSlippageInput.tsx
cpcramer: `SwapSettingsSlippageInput` is an optional sub-component, therefore I think we need to set defaultSlippage at the top level of the `Swap` component

src/swap/components/SwapProvider.tsx
```diff
@@ -272,6 +278,7 @@ export function SwapProvider({
       }
       await processSwapTransaction({
         config,
+        maxSlippage,
```
Zizzamia, line 281: In case was miss, let's cover this comment

https://github.com/coinbase/onchainkit/pull/1210/files#r1744336194

src/swap/components/SwapSettingsSlippageInput.tsx
cpcramer: We can move the defaultSlippage parameter to the outer layer and always set it in the top level

Zizzamia: Ok so are we thinking about
```
<Swap maxSlippage={} >
```

and use that for the both the default value and maxSlippage value?

Zizzamia: or as we will have several default options, we should consider to have a more approach on this.

And do
```
<Swap options={} >
```

and is `options` the right word, or should be `settings`?

Zizzamia: @abcrane123 curis to hear your thoughts here as well.

cpcramer: How about: 

`<Swap defaultMaxSlippage={} >`

This would be used as the `defaultMaxSlippage` value. 

If the developer includes the `SwapSettingsSlippageInput` sub-component, then the input `maxSlippage` values will override the `defaultMaxSlippage`.

src/swap/components/SwapProvider.tsx
cpcramer: This link is broken, it is sending me to the top of the updated files page.

Zizzamia: Also, I don't think is clean to just pass the maxSlippage. Like think 7 steps ahead, statusData will contain other info that need to persist inside processSwapTransaction when setLifeCycleStatus is called.

So we should pass lifeCycleStatus, inside processSwapTransaction.

So that we can persit the info we need from statusData when we recall setLifeCycleStatus.

src/swap/components/SwapSettingsSlippageInput.tsx
cpcramer: good point 👍

src/swap/components/SwapProvider.tsx
```diff
@@ -46,7 +46,9 @@ export function SwapProvider({
   const { address } = useAccount();
   // Feature flags
   const { useAggregator } = experimental;
-
+  const [maxSlippage, _setMaxSlippage] = useState(
+    experimental.maxSlippage || 3,
+  );
```
alessey, line 51: don't want this in state do we?  If the experimental prop changes, should that update here?

src/swap/components/SwapProvider.tsx
alessey: should useAggregator be handled the same way?

src/swap/components/SwapProvider.tsx
cpcramer: The experimental prop won't change after the component is rendered. 

It will set the initial `maxSlippage` value. Any updates will be handled by `LifeCycleStatus` slippageChange maxSlippage property

https://github.com/coinbase/onchainkit/pull/1210/files#diff-9bca97e90be8d6cfaa2c6813dbf1d684ed792b80b395f6b033d3d06963f23371R91-R94

cpcramer: And we update this value using `setLifeCycleStatus`'

https://github.com/coinbase/onchainkit/pull/1210/files#diff-704c4aa565af1d2becb4f70ce49ac884069cdb8e451c6e3afe603a6d47de07bcR35-R38

src/swap/components/SwapProvider.tsx
cpcramer: @Zizzamia Updated to pass in `lifeCycleStatus` instead of `maxSlippage`.

And now within `processSwapTransaction` we are grabbing `maxSlippage`:

```
  const maxSlippage =
    lifeCycleStatus.statusName !== 'error'
      ? lifeCycleStatus.statusData.maxSlippage
      : 3;
```

src/swap/components/SwapSettingsSlippageInput.tsx
cpcramer: @Zizzamia I can add this in when I update the `Swap.tsx` file in my following PR

0xAlec: I like `options` prop IMO

src/swap/components/SwapProvider.tsx
cpcramer: Some context on why we are setting it this way 

https://github.com/coinbase/onchainkit/pull/1210#discussion_r1744323560

alessey: If we are only using this const to initialize lifecycle status, can it just be a const?  or alternatively we could move the initiallifecycle status into a const?

0xAlec (approved):

src/swap/components/SwapProvider.tsx
Zizzamia: > The experimental prop won't change after the component is rendered.

@cpcramer that sentence it's incorrect! We should always expect a prop can change, that's the all point of prop. :)

src/swap/components/SwapSettingsSlippageInput.tsx
Zizzamia: Yeah, I am 50/50 between `options` and `settings`. If we go with `options` than we need to make sure it will be use as new best practices in other componetns as well.

So let's go with `options`.

Zizzamia: Btw, in the next PR, let's make sure we remove `defaultSlippage`, and get the info about `maxSlippage` from `lifeCycleStatus.statusData.maxSlippage`.

@cpcramer let me super clear, we don't want this prop, because it will confuse people.

src/swap/utils/processSwapTransaction.ts
```diff
@@ -9,13 +9,19 @@ import type { ProcessSwapTransactionParams } from '../types';
...
   setLifeCycleStatus,
   swapTransaction,
   useAggregator,
 }: ProcessSwapTransactionParams) {
   const { transaction, approveTransaction, quote } = swapTransaction;
 
+  const maxSlippage =
+    lifeCycleStatus.statusName !== 'error'
+      ? lifeCycleStatus.statusData.maxSlippage
+      : 3;
```
Zizzamia, line 23: If 3 is our default for everything, this should be in a share `const`.

Zizzamia: Follow up on this in the next PR.

Zizzamia (approved):
I want to keep the ball moving, so I am going to merge this, and @cpcramer follow up with the notes in your next PR.

PR #1173, opened 2024-08-27
feat: add connect wallet functionality to Swap component

abcrane123:
**What changed? Why?**
- if no wallet connected, prompt user to connect wallet after filling out inputs

https://github.com/user-attachments/assets/fe23c13e-a5c0-4b34-930f-a2ae69b938e1

Screenshot 2024-08-28 at 11 22 45 AM

**Notes to reviewers**

**How has it been tested?**

PR #1216, opened 2024-09-05
prompt for wallet connection in `TransactionButton`

0xAlec:
**What changed? Why?**
- do not disable `Transaction` if missing `address`
- instead, prompt the user to connect their wallet

**Notes to reviewers**

**How has it been tested?**
- playground

### DIFF
--- a/src/transaction/components/TransactionButton.tsx
+++ b/src/transaction/components/TransactionButton.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useMemo } from 'react';
-import { useAccount, useChainId } from 'wagmi';
+import { useAccount, useChainId, useConnect } from 'wagmi';
 import { useShowCallsStatus } from 'wagmi/experimental';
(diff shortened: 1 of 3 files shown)


src/transaction/components/TransactionButton.tsx
```diff
@@ -59,7 +60,10 @@ export function TransactionButton({
     return buttonText;
   }, [buttonText, errorMessage, receipt]);
 
-  const handleSubmit = useCallback(() => {
+  const handleSubmit = useCallback(async () => {
+    if (!address) {
+      await connectAsync({ connector: connectors[0] });
```
Zizzamia, line 65: Let's avoid create a branch of developer experience by adding a new way to solve the same problem we did with https://github.com/coinbase/onchainkit/pull/1173 .

We want to make sure both experience re-use Wallet logic, and do not create internal branch on how the same problem can be solved.

PR #1218, opened 2024-09-05
Fix: Framegear with Multiple Text Inputs

brendan-defi:
**What changed? Why?**
* Currently, when using Framegear, if there are multiple frames with text input fields, the `inputText` state will persist to the latter frames (ie. if a user inputs "blah" in the first input field, and the next frame has an `input` field, that field will be pre-filled with "blah"

* This PR fixes the issue by enabling Framegear to better handle multiple text inputs
  * achieved by creating a `TextInputs` context
  * leveraged `useTextInputs` in `FrameInput` and `Frame` to appropriately reset inputText state for various frame changes
  * the following frame interactions reset the `inputText` state:
    * `FrameInput`'s `fetch` button
    * Any `post` or `post_redirect` button clicks in the `Frame` component

**Notes to reviewers**

**How has it been tested?**
locally

### DIFF
--- a/framegear/contexts/TextInputs.tsx
+++ b/framegear/contexts/TextInputs.tsx
@@ -1,0 +1,34 @@
+import React, { createContext, useState, useContext, ReactNode } from 'react';
+
+type TextInputsContextType = {
+  inputText: string;
+  setInputText: React.Dispatch<React.SetStateAction<string>>;
+};
+
+const TextInputsContext = createContext<TextInputsContextType | undefined>(
+  undefined
+);
+
+type TextInputsProviderProps = {
+  children?: ReactNode;
+};
+
+export default function TextInputsProvider({
+  children,
+}: TextInputsProviderProps) {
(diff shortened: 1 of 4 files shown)


framegear/contexts/TextInputs.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { createContext, useState, useContext, ReactNode } from 'react';
+
+type TextInputsContextType = {
```
Zizzamia, line 3: At some point we want to start organize the types in a `types.ts` file.

Zizzamia (approved):

framegear/contexts/TextInputs.tsx
brendan-defi: Yeah I was wondering about that. And also moving the `RedirectModalContext` into contexts. 

I'll do a little clean-up PR in the next couple days with some of those chores.

PR #1224, opened 2024-09-09
feat: add scaled text input

alessey:
**What changed? Why?**
Screen Recording 2024-09-09 at 10 21 17 AM

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/internal/components/ScaledTextInput.tsx
+++ b/src/internal/components/ScaledTextInput.tsx
@@ -1,0 +1,53 @@
+import { useEffect, useRef, useState } from 'react';
+import { cn } from '../../styles/theme';
+import type { TextInputReact } from './TextInput';
+import { TextInput } from './TextInput';
+
+export function ScaledTextInput(props: TextInputReact) {
+  const inputContainer = useRef<HTMLDivElement | null>(null);
+  const offscreenRef = useRef<HTMLSpanElement | null>(null);
+  const [containerScale, setContainerScale] = useState(1);
+
+  useEffect(() => {
+    if (props.value && offscreenRef.current && inputContainer.current) {
+      const containerWidth = inputContainer.current.offsetWidth;
+      const textWidth = offscreenRef.current.offsetWidth;
+      const scaleFactor = containerWidth / textWidth;
+
+      if (scaleFactor < 1) {
+        setContainerScale(scaleFactor);
+      } else {
+        setContainerScale(1);
+      }
+    } else {
+      setContainerScale(1);
+    }
+  }, [props.value]);
+
+  return (
+    <>
+      <div ref={inputContainer}>
+        <div
+          className="origin-[0_50%] overflow-hidden"
+          style={{
+            transform: `scale(${containerScale})`,
+            width: `${100 / containerScale}%`,
+          }}
+          data-testid="scaled-text-input-container"
+        >
+          <TextInput {...props} />
+        </div>
+      </div>
+      <span
+        ref={offscreenRef}
+        className={cn(
+          props.className,
+          'invisible absolute right-0 bottom-0 w-auto translate-x-[-100%] opacity-0',
+        )}
+        aria-hidden="true"
+      >
+        {props.value}
+      </span>
+    </>
+  );
+}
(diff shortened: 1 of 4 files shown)


src/internal/components/ScaledTextInput.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  return (
+    <>
+      <div ref={inputContainer}>
+        <div
+          className="origin-[0_50%] overflow-hidden"
+          style={{
+            transform: `scale(${containerScale})`,
+            width: `${100 / containerScale}%`,
+          }}
```
cpcramer, line 35: We should refrain from using inline-styling

src/internal/components/ScaledTextInput.tsx
```diff
@@ -0,0 +1,53 @@
+import { useEffect, useRef, useState } from 'react';
+import { cn } from '../../styles/theme';
+import type { TextInputReact } from './TextInput';
+import { TextInput } from './TextInput';
+
+export function ScaledTextInput(props: TextInputReact) {
```
Zizzamia, line 6: Can you add some comments that explain when to use `ScaledTextInput` vs `TextInput`?

Zizzamia, line 6: Can you add some comments that explain when to use `ScaledTextInput` vs `TextInput`?

src/internal/components/ScaledTextInput.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+      if (scaleFactor < 1) {
+        setContainerScale(scaleFactor);
+      } else {
+        setContainerScale(1);
+      }
+    } else {
+      setContainerScale(1);
+    }
+  }, [props.value]);
```
Zizzamia, line 25: Open question, could this re-fire too many times if we keep `props.value` too wide?

src/internal/components/ScaledTextInput.tsx
```diff
@@ -0,0 +1,53 @@
+import { useEffect, useRef, useState } from 'react';
+import { cn } from '../../styles/theme';
+import type { TextInputReact } from './TextInput';
+import { TextInput } from './TextInput';
+
+export function ScaledTextInput(props: TextInputReact) {
+  const inputContainer = useRef<HTMLDivElement | null>(null);
+  const offscreenRef = useRef<HTMLSpanElement | null>(null);
+  const [containerScale, setContainerScale] = useState(1);
```
Zizzamia, line 9: Help me understand what this number means. Maybe with a line of comment.

src/internal/components/ScaledTextInput.tsx
```diff
@@ -0,0 +1,53 @@
...
+      }
+    } else {
+      setContainerScale(1);
+    }
+  }, [props.value]);
+
+  return (
+    <>
+      <div ref={inputContainer}>
+        <div
```
Zizzamia, line 30: Can this be done with just 1 `<div>` instead of 2?

src/internal/components/ScaledTextInput.tsx
Zizzamia: @cpcramer I think we are forced here to have the inline style. As this is a dynamic CSS story, which I don't think Tailwind has an easy solution for it.

@cpcramer this is also very similar on how CSS animation are built, where you have a basic class, but then use inline style to do crazy animaiton tricks.

src/internal/components/ScaledTextInput.tsx
```diff
@@ -0,0 +1,53 @@
...
+          style={{
+            transform: `scale(${containerScale})`,
+            width: `${100 / containerScale}%`,
+          }}
+          data-testid="scaled-text-input-container"
+        >
+          <TextInput {...props} />
+        </div>
+      </div>
+      <span
```
Zizzamia, line 41: This is dark magic, @alessey tell me more!!! I rememer this from back in the day, can you please add some comments before the `return` to help explain how we use the combo of `inputContainer` and `offscreenRef` to make this magic work.

@alessey the html wizards 🧙

alessey:
closing in favor of increasing the padding between the value and select and keeping the current scroll behavior, cc: @mindapivessa

PR #1234, opened 2024-09-11
feat: added type LifecycleStatusDataShared

Zizzamia:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/swap/components/SwapProvider.tsx
+++ b/src/swap/components/SwapProvider.tsx
@@ -122,12 +122,18 @@ export function SwapProvider({
       setLifeCycleStatus({
         statusName: 'init',
         statusData: {
-          isMissingRequiredField: false,
+          isMissingRequiredField:
+            lifeCycleStatus.statusData.isMissingRequiredField,
           maxSlippage,
         },
       });
     }
-  }, [hasHandledSuccess, lifeCycleStatus.statusName, maxSlippage]);
+  }, [
+    hasHandledSuccess,
+    lifeCycleStatus.statusData,
+    lifeCycleStatus.statusName,
+    maxSlippage,
+  ]);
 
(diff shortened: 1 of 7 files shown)


src/swap/components/SwapProvider.tsx
```diff
@@ -122,12 +122,18 @@ export function SwapProvider({
       setLifeCycleStatus({
         statusName: 'init',
         statusData: {
-          isMissingRequiredField: false,
+          isMissingRequiredField:
+            lifeCycleStatus.statusData.isMissingRequiredField,
           maxSlippage,
         },
       });
```
alessey, line 129: should these use state updater functions?  i.e. setLifecycleStatus(status => ({ statusName: 'init', ...status }))?

Zizzamia: > If lifecycleStatus is the component state, should it include loading, error, pendingTransaction instead of reacting to lifecycle changes to set them in context? Or would only transient data live in lifecycle and persisted data live in context?

I am been asking the same question to myself for the last 2 weeks 😆 and kind of letting the PRs inform how to evolve this.

So far, I am realizing that the `lifecycle` data, needs to have both `transient` and `persisted` data. By also making the overall `context` super slip.

The main reason I think it needs both, is because it will give a true deep heart beat of what's going on inside the component to the developer.

Honestly, I am leaning more and more to expose everything and than let the developer pick and choose how to use them, for any custom behavior they want to create.

Zizzamia: > should these use state updater functions? i.e. setLifecycleStatus(status => ({ statusName: 'init', ...status }))?

I don't follow, help me understand.

PR #1237, opened 2024-09-11
fix: update lifecyclestatus

alessey:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/swap/types.ts
```diff
@@ -72,11 +84,13 @@ type LifecycleStatusDataShared = {
 export type LifeCycleStatus =
   | {
       statusName: 'init';
-      statusData: LifecycleStatusDataShared;
+      statusData: LifecycleStatusDataBase;
```
Zizzamia, line 87: I know the word Base here is correct, but I would like to avoid using it, so there is no confusion with Base chain.

PR #1238, opened 2024-09-11
feat: re-type `walletCapabilities` object

0xAlec:
**What changed? Why?**
- use `viem` native typing for `walletCapabilities` object
- move `useCapabilitiesSafe` to `internal/hooks` and call on a per-component basis
- memoize `useCapabilitiesSafe`

now developers can use capabilities that are not explicitly supported by OnchainKit since we are using the native `viem` typing

**Notes to reviewers**

**How has it been tested?**


src/constants.ts
```diff
@@ -0,0 +1,6 @@
+// Capabilities
+export enum Capabilities {
+  AtomicBatch = 'atomicBatch',
+  AuxiliaryFunds = 'auxiliaryFunds',
+  PaymasterService = 'paymasterService',
+}
```
0xAlec, line 6: we'll still need to use this enum for any capabilities we support in OnchainKit but now developers can use non-OCK supported capabilities

```
const { walletCapabilities } = useOnchainKit()

if (walletCapabilities[someOtherCapability]){
  ...
}
```

src/OnchainKitConfig.ts
```diff
@@ -9,11 +9,7 @@ export const ONCHAIN_KIT_CONFIG: OnchainKitConfig = {
   chain: baseSepolia,
   rpcUrl: null,
   schemaId: null,
-  walletCapabilities: {
-    hasAtomicBatch: false,
-    hasAuxiliaryFunds: false,
-    hasPaymasterService: false,
-  },
+  walletCapabilities: {},
```
Zizzamia, line 12: We should not expose this at all here, but instead being part of his own Hook.

Also this change will trigger a braking change, so we need to time this merge correctly.

src/OnchainKitProvider.tsx
```diff
@@ -30,11 +30,7 @@ export function OnchainKitProvider({
       chain: chain,
       rpcUrl: rpcUrl ?? null,
       schemaId: schemaId ?? null,
-      walletCapabilities: walletCapabilities ?? {
-        hasAtomicBatch: false,
-        hasAuxiliaryFunds: false,
-        hasPaymasterService: false,
-      },
+      walletCapabilities: walletCapabilities ?? {},
```
Zizzamia, line 33: This should not be store here anymore!

But instead our `useCapabilities` hook should memoize this info.

And long story short we should have a

```
import { useCapabilities } from `@coinbase/onchainkit`;
```

and just use that around the library to get the data.

src/constants.ts
Zizzamia: For each of them, we should probably add a comment that links on where the EIP is.

Just because I can imagine a convo in 2 months where someone else tries to figure out what those are and they take a few to find the correct docs.

src/internal/hooks/useCapabilitiesSafe.ts
```diff
@@ -0,0 +1,23 @@
+import { useMemo } from 'react';
+import type { WalletCapabilities } from 'viem';
+import { useAccount } from 'wagmi';
+import { useCapabilities } from 'wagmi/experimental';
+import type { UseCapabilitiesSafeParams } from '../../types';
+
+export function useCapabilitiesSafe({
```
Zizzamia, line 7: I still feel like we should rename this `useCapabilities`, as the fact that we are not exporting this externally adding the word Safe just feels odd.

src/transaction/hooks/useSendWalletTransactions.tsx
```diff
@@ -18,7 +19,7 @@ export const useSendWalletTransactions = ({
     if (!transactions) {
       return;
     }
-    if (walletCapabilities.hasAtomicBatch) {
+    if (walletCapabilities[Capabilities.AtomicBatch]?.supported) {
```
Zizzamia, line 22: Yeah I like the `enum`, is very clean

src/transaction/components/TransactionProvider.tsx
```diff
@@ -77,7 +78,9 @@ export function TransactionProvider({
     : TRANSACTION_TYPE_CONTRACTS;
 
   // Retrieve wallet capabilities
-  const { walletCapabilities } = useOnchainKit();
+  const walletCapabilities = useCapabilitiesSafe({
+    chainId: chainId || 84532,
```
Zizzamia, line 82: `84532` should this just be `baseSepolia` from Viem?

src/internal/hooks/useCapabilitiesSafe.ts
0xAlec: I think it will make more sense once we add the functionality to debounce or stop subsequent calls to non-EIP5792 wallets on this hook

0xAlec: (which isn't supported in the vanilla Wagmi hook - hence `unsafe`)

0xAlec: I'd also say the vanilla Wagmi hook is `unsafe` because it does not return an empty object for non-EIP5792 wallets

src/transaction/components/TransactionProvider.tsx
```diff
@@ -77,7 +79,9 @@ export function TransactionProvider({
     : TRANSACTION_TYPE_CONTRACTS;
 
   // Retrieve wallet capabilities
-  const { walletCapabilities } = useOnchainKit();
+  const walletCapabilities = useCapabilitiesSafe({
+    chainId: chainId || baseSepolia.id,
```
Zizzamia, line 83: For follow up PR, we need to make sure we handle this correctly where:
- we prioritize the `chainId` from the Transaction component
- if that not present we use the `chainId` from `useOnchainKit()`
- which if not present at OnchainKit config defaults to BaseSepolia.

We can not cut a release if this is not present! Just FYI
cc @0xAlec @alessey @cpcramer @abcrane123

src/transaction/types.ts
```diff
@@ -6,12 +6,13 @@ import type {
...
   TransactionReceipt,
 } from 'viem';
+import type { WalletCapabilities as ViemWalletCapabilities } from 'viem';
 import type { Config } from 'wagmi';
 import type {
   SendTransactionMutateAsync,
   WriteContractMutateAsync,
 } from 'wagmi/query';
-import type { WalletCapabilities as OnchainKitWalletCapabilities } from '../types';
+// 🌲☀🌲
```
Zizzamia, line 15: Not big deal, but this is an easter egg that should alway be as first line only :)

So you probably want to remove it from here

Zizzamia (approved):
Merging the PR as it is, and then fixing this in a quick follow.

PR #1239, opened 2024-09-11
Adjust status functionality for transaction component

abcrane123:
**What changed? Why?**
- when executing multiple transactions with EOA, we were emitting `transactionPending` -> `transactionLegacyExecuted` -> `transactionLegacyExecuted` 
- correct emission: `transactionPending` -> `transactionLegacyExecuted` -> `transactionPending` -> `transactionLegacyExecuted` 
- we were also emitting two `success` statuses: the first one contained both transaction receipts, the second one contained only a single transaction receipt. 

**Notes to reviewers**

**How has it been tested?**


src/transaction/components/TransactionProvider.tsx
```diff
@@ -212,9 +220,23 @@ export function TransactionProvider({
     }
   }, [transactionStatus]);
 
+  // Reset status to pending when executing multiple transactions with EOA
+  useEffect(() => {
+    if (
+      transactionStatus === 'pending' &&
+      lifeCycleStatus.statusName === 'transactionLegacyExecuted'
```
alessey, line 227: shouldn't lifeCycleStatus already have been set by the useEffect above that also checks transactionStatus pending?

src/transaction/components/TransactionProvider.tsx
```diff
@@ -76,6 +76,10 @@ export function TransactionProvider({
     ? TRANSACTION_TYPE_CALLS
     : TRANSACTION_TYPE_CONTRACTS;
 
+  const [prevStatusEmitted, setPrevStatusEmitted] = useState(
```
0xAlec, line 79: nit: can we use `previousStatus, setPreviousStatus` instead

src/transaction/components/TransactionProvider.tsx
abcrane123: yes but transaction status doesn't actually change. it is `pending` until it reaches the `success` state so it wasn't getting reset to `transactionPending` after we change it to `transactionLegacyExecuted`

src/transaction/components/TransactionProvider.tsx
```diff
@@ -191,15 +195,19 @@ export function TransactionProvider({
         transactionReceipts: lifeCycleStatus.statusData.transactionReceipts,
       });
     }
-    // Emit Status
-    onStatus?.(lifeCycleStatus);
+    // Emit Status on status change
+    if (prevStatusEmitted !== lifeCycleStatus.statusName) {
+      onStatus?.(lifeCycleStatus);
+      setPrevStatusEmitted(lifeCycleStatus.statusName);
+    }
```
0xAlec, line 202: are there any other use cases aside from subsequent transactions that we want to emit duplicate statuses for?

e.g. `init`?

I'd also add some comments here on why this is an edge case (multiple pending transactions)

0xAlec: also nit but if we can make this something like

```
if (previousStatus !== currentStatus) {
 // some comment here on edge case
}
``

src/transaction/components/TransactionProvider.tsx
Zizzamia: @abcrane123 this approach adds more complexity to our logic, and makes it harder to mantain later.

What we should do instead, is add a persistant value inside statusData that keep info regarding the pending part.

With @alessey we are actually exploring this way of thinking for Swap and make sure become the new standard https://github.com/coinbase/onchainkit/pull/1237

Make sure to sync with @alessey and follow the same approach.

PR #1243, opened 2024-09-12
chore: export getAddress

abcrane123:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


alessey (approved):
looks like this must have been an oversight as its available in the docs https://onchainkit.xyz/identity/get-address

0xAlec (approved):

cpcramer (approved):

PR #1248, opened 2024-09-12
chore: Update Swap Slippage v1 API for getting the Swap Quote

cpcramer:
**What changed? Why?**
 Adjust slippage for V1 API (aggregator). V1 expects slippage in tenths of a percent (e.g., 30 = 3%). This was implemented when creating the transaction but not when fetching the quote 

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/api/getSwapQuote.ts
+++ b/src/api/getSwapQuote.ts
@@ -14,7 +14,7 @@ import { getAPIParamsForToken } from './utils/getAPIParamsForToken';
  * Retrieves a quote for a swap from Token A to Token B.
  */
 export async function getSwapQuote(
-  params: GetSwapQuoteParams,
+  params: GetSwapQuoteParams
 ): Promise<GetSwapQuoteResponse> {
   // Default parameters
   const defaultParams = {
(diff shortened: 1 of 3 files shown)


0xAlec (approved):

src/api/getSwapQuote.ts
```diff
@@ -37,8 +37,14 @@ export async function getSwapQuote(
...
+    // Adjust slippage for V1 API (aggregator)
+    // V1 expects slippage in tenths of a percent (e.g., 30 = 3%)
+    if (params.useAggregator) {
+      slippagePercentage = (Number(params.maxSlippage) * 10).toString();
+    }
     apiParams = {
-      slippagePercentage: params.maxSlippage,
+      slippagePercentage: slippagePercentage,
       ...apiParams,
     };
```
alessey, line 49: if we need this logic in a 3rd place, we should extract to a utility

alessey (approved):

src/api/getSwapQuote.ts
abcrane123: agreed

abcrane123 (approved):

PR #1249, opened 2024-09-12
fix: update swap lifecyclestatus

alessey:
**What changed? Why?**
* remove context states and use lifecyclestatus as the source of truth
* persist all lifecycle status data (except errors)

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/swap/components/SwapProvider.tsx
+++ b/src/swap/components/SwapProvider.tsx
@@ -18,8 +18,8 @@ import { useFromTo } from '../hooks/useFromTo';
 import { useResetInputs } from '../hooks/useResetInputs';
 import type {
   LifeCycleStatus,
+  LifeCycleStatusUpdate,
   SwapContextType,
-  SwapError,
   SwapProviderReact,
 } from '../types';
 import { isSwapError } from '../utils/isSwapError';
@@ -52,9 +52,6 @@ export function SwapProvider({
   const { useAggregator } = experimental;
   // Core Hooks
   const accountConfig = useConfig();
-  const [loading, setLoading] = useState(false);
-  const [error, setError] = useState<SwapError>();
-  const [isTransactionPending, setPendingTransaction] = useState(false);
   const [lifeCycleStatus, setLifeCycleStatus] = useState<LifeCycleStatus>({
     statusName: 'init',
     statusData: {
@@ -62,6 +59,30 @@ export function SwapProvider({
       maxSlippage: config.maxSlippage,
     },
   }); // Component lifecycle
+
+  // Update lifecycle status, statusData will be persisted for the full lifeCycle
+  const updateLifeCycleStatus = useCallback(
+    (newStatus: LifeCycleStatusUpdate) => {
+      setLifeCycleStatus((prevStatus: LifeCycleStatus) => {
+        // do not persist errors
+        const persistedStatusData =
+          prevStatus.statusName === 'error'
+            ? (({ error, code, message, ...statusData }) => statusData)(
+                prevStatus.statusData,
+              )
+            : prevStatus.statusData;
+        return {
+          statusName: newStatus.statusName,
+          statusData: {
+            ...persistedStatusData,
+            ...newStatus.statusData,
+          },
+        } as LifeCycleStatus;
+      });
+    },
+    [],
+  );
+
   const [hasHandledSuccess, setHasHandledSuccess] = useState(false);
   const { from, to } = useFromTo(address);
   const { sendTransactionAsync } = useSendTransaction(); // Sending the transaction (and approval, if applicable)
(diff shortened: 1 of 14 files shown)


src/swap/components/SwapProvider.tsx
```diff
@@ -52,16 +52,37 @@ export function SwapProvider({
...
+          statusName: newStatus.statusName,
+          statusData: {
+            ...persistedStatusData,
+            ...newStatus.statusData,
+          },
+        } as LifeCycleStatus;
+      });
+    },
+    [],
+  );
```
alessey, line 84: This is the main update, creating a setState callback function here that persists all statusData except errors, this data can be overwritten, but will persist until init is called again.

src/swap/components/SwapProvider.tsx
```diff
@@ -118,30 +123,42 @@ export function SwapProvider({
...
+        amountFrom: from.amount,
+        amountTo: to.amount,
+        tokenFrom: from.token,
+        tokenTo: to.token,
+        // token is missing
+        isMissingRequiredField:
+          !from.token || !to.token || !from.amount || !to.amount,
+      },
+    });
+  }, [from, to, updateLifeCycleStatus]);
```
alessey, line 161: It seemed like we should be triggering an amountChange on toggle since from and to switch, thoughts?

src/swap/components/SwapProvider.tsx
```diff
@@ -228,54 +239,42 @@ export function SwapProvider({
...
       return;
     }
-    const maxSlippage = lifeCycleStatus.statusData.maxSlippage;
-    setLifeCycleStatus({
-      statusName: 'init',
-      statusData: {
-        isMissingRequiredField: false,
-        maxSlippage,
-      },
-    });
```
alessey, line 276: I don't believe this is needed, unless it is possible to swap even if the quote fails.  Status init doesn't make sense at this point when you are in the middle of the process.

src/swap/utils/processSwapTransaction.ts
```diff
@@ -105,32 +84,30 @@ export async function processSwapTransaction({
...
     data: transaction.data,
   });
-
+  updateLifeCycleStatus({
+    statusName: 'transactionApproved',
+    statusData: {
+      transactionHash: txHash,
+      transactionType: useAggregator ? 'ERC20' : 'Permit2',
+    },
+  });
```
alessey, line 101: Currently we are showing confirm in wallet until the receipt, seemed like we should emit the transaction status and then you'll get success after receipt.

src/swap/components/SwapProvider.tsx
```diff
@@ -73,26 +94,10 @@ export function SwapProvider({
...
-    if (lifeCycleStatus.statusName === 'amountChange') {
-      setError(undefined);
-    }
-    if (lifeCycleStatus.statusName === 'transactionPending') {
-      setLoading(true);
-      setPendingTransaction(true);
-    }
-    if (lifeCycleStatus.statusName === 'transactionApproved') {
-      setPendingTransaction(false);
-    }
```
alessey, line 90: no more reacting to state changes to set state

src/swap/components/SwapProvider.tsx
Zizzamia: very very clever, and can't wait for next PRs to have this abstract even more, so we can use it in other components.

src/swap/components/SwapProvider.tsx
Zizzamia: so smoooooooth

src/swap/components/SwapProvider.tsx
Zizzamia: You right! Great point

src/swap/components/SwapProvider.tsx
```diff
@@ -151,7 +168,6 @@ export function SwapProvider({
       dToken?: Token,
       // biome-ignore lint/complexity/noExcessiveCognitiveComplexity: TODO Refactor this component
     ) => {
-      const maxSlippage = lifeCycleStatus.statusData.maxSlippage;
```
Zizzamia, line 154: so clean

src/swap/components/SwapProvider.tsx
Zizzamia: Yeah makes sense.

src/swap/types.ts
```diff
@@ -69,34 +66,69 @@ export type LifeCycleStatus =
...
     }
   | {
       statusName: 'success';
       statusData: {
         transactionReceipt: TransactionReceipt;
-      } & LifecycleStatusDataShared;
+      } & LifeCycleStatusDataShared;
     };
 
+// make all keys in T optional if they are in K
```
Zizzamia, line 93: Dark magic, love it!!!!

src/swap/types.ts
```diff
@@ -69,34 +66,69 @@ export type LifeCycleStatus =
...
+              >;
+            } // make all keys in LifeCycleStatusDataShared optional
+          : {
+              statusData: PartialKeys<
+                D,
+                keyof D & keyof LifeCycleStatusDataShared
+              >;
+            })
+    : never
+  : never;
```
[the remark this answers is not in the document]
Zizzamia: As this complexity will be shared across many experiences, this is a complexity that is fair to pay.

Plus, for everyone reading, we tested all other options and we still found this trade-off being the most simple overall.

src/swap/utils/getSwapMessage.ts
```diff
@@ -32,23 +29,25 @@ export function getSwapMessage({
     return SwapMessage.INSUFFICIENT_BALANCE;
   }
   // handle pending transaction
-  if (isTransactionPending) {
+  if (lifeCycleStatus.statusName === 'transactionPending') {
```
Zizzamia, line 32: so clean

src/swap/utils/processSwapTransaction.ts
Zizzamia: I lean towards your point, but can we ask also @0xAlec @cpcramer @mykcryptodev @abcrane123 what they think about this.

Zizzamia (approved):

Zizzamia:
Great job @alessey !!! 
Was not an easy refactor, and looking forward to your next PRs in keep polishing this.

src/swap/utils/processSwapTransaction.ts
0xAlec: +1

PR #1254, opened 2024-09-13
feat: set `v2` as default API for `Swap`

0xAlec:
**What changed? Why?**
- turn off `v1` aggregator API (0x and 1inch)
- switch on `v2` API by default (Uniswap)

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/swap/components/Swap.tsx
+++ b/src/swap/components/Swap.tsx
@@ -17,7 +17,7 @@ export function Swap({
     maxSlippage: DEFAULT_MAX_SLIPPAGE,
   },
   className,
-  experimental = { useAggregator: true },
+  experimental = { useAggregator: false },
   onError,
   onStatus,
   onSuccess,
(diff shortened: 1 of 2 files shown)


cpcramer (approved):

abcrane123 (approved):

alessey (approved):

src/swap/components/Swap.tsx
```diff
@@ -17,7 +17,7 @@ export function Swap({
     maxSlippage: DEFAULT_MAX_SLIPPAGE,
   },
   className,
-  experimental = { useAggregator: true },
+  experimental = { useAggregator: false },
```
Zizzamia, line 20: If we are switching, why keep all this extra code?

If we are confident in how it works, we should just remove all `experimental` code.

Zizzamia: Are we thinking to do that later? And later when?

0xAlec: let's leave this option in for developers for now. will share the migration plan in Slack

PR #1262, opened 2024-09-16
feat: implement two slash in docs

alessey:
**What changed? Why?**
Added twoslash to all codeblocks

image

**Notes to reviewers**
Twoslash needs to be [explicitly triggered](https://github.com/wevm/vocs/blob/main/src/vite/plugins/mdx.ts#L107) on each codeblock

**How has it been tested?**

### DIFF
--- a/site/docs/pages/guides/lifecycle-status.mdx
+++ b/site/docs/pages/guides/lifecycle-status.mdx
@@ -31,10 +31,13 @@ can tap into the rhythm and customize the behavior of your components. 🎶
 The Lifecycle Status is a TypeScript object that provides easy access to the `statusName` and `statusData` properties, 
 allowing you to stay informed and responsive.
 
-```ts
+```tsx twoslash
+import { useCallback } from 'react';
+import { Transaction } from '@coinbase/onchainkit/transaction';
+// ---cut-before---
 import type { LifeCycleStatus } from '@coinbase/onchainkit/transaction';
 
-const handleOnStatus = useCallback(status: LifeCycleStatus) => {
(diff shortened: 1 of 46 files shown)


site/docs/pages/guides/lifecycle-status.mdx
```diff
@@ -31,10 +31,13 @@ can tap into the rhythm and customize the behavior of your components. 🎶
...
+```tsx twoslash
+import { useCallback } from 'react';
+import { Transaction } from '@coinbase/onchainkit/transaction';
+// ---cut-before---
 import type { LifeCycleStatus } from '@coinbase/onchainkit/transaction';
 
-const handleOnStatus = useCallback(status: LifeCycleStatus) => {
+const handleOnStatus = useCallback((status: LifeCycleStatus) => {
   console.log('LifecycleStatus', status);
 }, []);
```
alessey, line 42: TODO: update case when new version

site/docs/styles.css
```diff
@@ -38,7 +40,7 @@
 
 .vocs_CodeGroup {
   border-radius: 8px;
-  overflow: hidden;
+  overflow: visible;
```
alessey, line 43: allow twoslash popups to overflow code boxes

Zizzamia: niiice

Zizzamia: Hopefullly this will not cause a regression with Mobile Web, but if it does we need to re-think this.

Zizzamia (approved):
Niiiice!

Great job

PR #1264, opened 2024-09-16
feat: batched Swap transactions from ERC-20

0xAlec:
**What changed? Why?**
- add `useAwaitCalls` hook
- add support for atomic batching methods in `processSwapTransaction`

this will allow smart wallets to execute a swap from an ERC-20 in a single transaction (as opposed to 3 previously)

also, updated the inaccurate Swap lifecycle statuses.

**Notes to reviewers**

**How has it been tested?**
playground and unit tests

### DIFF
--- a/src/swap/components/SwapProvider.tsx
+++ b/src/swap/components/SwapProvider.tsx
@@ -58,7 +58,7 @@ export function SwapProvider({
   // Core Hooks
   const accountConfig = useConfig();
   const [callsId, setCallsId] = useState<Hex>();
-  const { data } = useCallsStatus({
+  const callsStatus = useCallsStatus({
     id: callsId || '0x',
     query: {
       refetchInterval: (query) => {
@@ -79,7 +79,7 @@ export function SwapProvider({
   const awaitCallsStatus = useAwaitCalls({
     accountConfig,
     config,
-    data,
+    callsStatus,
     setLifecycleStatus,
   });
 
@@ -355,7 +355,6 @@ export function SwapProvider({
     lifecycleStatus,
     sendCallsAsync,
     sendTransactionAsync,
-    setCallsId,
     to.token,
     updateLifecycleStatus,
     useAggregator,
--- a/src/swap/components/SwapProvider.tsx
+++ b/src/swap/components/SwapProvider.tsx
@@ -5,9 +5,14 @@ import {
   useEffect,
   useState,
 } from 'react';
+import type { Hex } from 'viem';
+import { base } from 'viem/chains';
 import { useAccount, useConfig, useSendTransaction } from 'wagmi';
+import { waitForTransactionReceipt } from 'wagmi/actions';
+import { useCallsStatus, useSendCalls } from 'wagmi/experimental';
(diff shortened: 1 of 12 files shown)


src/swap/components/SwapProvider.tsx
```diff
@@ -52,6 +57,17 @@ export function SwapProvider({
...
+  const [callsId, setCallsId] = useState<Hex>();
+  const callsStatus = useCallsStatus({
+    id: callsId || '0x',
+    query: {
+      refetchInterval: (query) => {
+        return query.state.data?.status === 'CONFIRMED' ? false : 1000;
+      },
+      enabled: callsId !== undefined,
+    },
+  });
```
Zizzamia, line 69: I feel like we need to abstract this. Have this kind of logic in the Provider makes it a bit harder to understand.

src/swap/components/SwapProvider.tsx
```diff
@@ -52,6 +57,17 @@ export function SwapProvider({
...
+  const walletCapabilities = useCapabilitiesSafe({ chainId: base.id }); // Swap is only available on Base
```
Zizzamia, line 70: Do we want to force it like this? It feels a bit odd, like I will expect a `chainId` option at component level that defaults to `base.id` and than if it's not that there will be some logging explain what's going on.

Zizzamia:
Before I continue review this PR, how can we split this PR in sub-parts? it seems we are trying to achive 3~4 things in one hit.

And we are touch way too much core logic to just let this PR go in one go. Especially with 20+ apps running with this component as their main Swap logic.

src/swap/components/SwapProvider.tsx
```diff
@@ -296,14 +325,16 @@ export function SwapProvider({
...
+        setCallsId,
         swapTransaction: response,
+        updateLifecycleStatus,
         useAggregator,
+        walletCapabilities,
       });
-
-      // TODO: refresh balances
     } catch (err) {
+      console.log('Error', err);
```
alessey, line 337: missed this console

src/swap/utils/executeSwapTransactions.ts
```diff
@@ -0,0 +1,31 @@
...
+}: ExecuteSwapTransactionParams) {
+  if (walletCapabilities[Capabilities.AtomicBatch]?.supported) {
+    // For batched transactions, we'll use `SwapProvider` to listen for calls to emit the `success` state
+    updateLifecycleStatus({
+      statusName: 'transactionPending',
+    });
+    const callsId = await sendCallsAsync({
+      calls: transactions,
+    });
+    setCallsId(callsId);
```
alessey, line 22: shouldn't there be a transactionApproved status here?

src/swap/components/SwapProvider.tsx
```diff
@@ -52,6 +57,17 @@ export function SwapProvider({
   const { useAggregator } = experimental;
   // Core Hooks
   const accountConfig = useConfig();
+  const [callsId, setCallsId] = useState<Hex>();
```
alessey, line 60: should this be stored in lifecycle?

cpcramer (approved):

PR #1270, opened 2024-09-17
feat: add utils for batch Swap transactions

0xAlec:
**What changed? Why?**
- add `executeSingleTransactions` (EOA)
- add `executeSwapTransactions` (atomic batch with `useSendCalls`, and single with `executeSingleTransactions`)

these utils will be used for batching all 3 Swap transactions (from ERC-20) for smart wallets

**Notes to reviewers**

**How has it been tested?**
main PR playground - https://github.com/coinbase/onchainkit/pull/1264

### DIFF
--- a/src/swap/types.ts
+++ b/src/swap/types.ts
@@ -103,8 +103,8 @@ export type LifecycleStatus =
   | {
       statusName: 'transactionApproved';
       statusData: {
-        transactionHash: Hex;
-        transactionType: 'ERC20' | 'Permit2';
+        transactionHash?: Hex;
+        transactionType: 'ERC20' | 'Permit2' | 'Batched';
       } & LifecycleStatusDataShared;
     }
   | {
--- a/src/swap/types.ts
+++ b/src/swap/types.ts
@@ -1,5 +1,10 @@
 import type { Dispatch, ReactNode, SetStateAction } from 'react';
-import type { Address, Hex, TransactionReceipt } from 'viem';
+import type {
+  Address,
+  Hex,
+  TransactionReceipt,
+  WalletCapabilities,
+} from 'viem';
 import type {
   Config,
   UseBalanceReturnType,
@@ -8,6 +13,25 @@ import type {
 import type { SendTransactionMutateAsync } from 'wagmi/query';
 import type { BuildSwapTransaction, RawTransactionData } from '../api/types';
 import type { Token } from '../token/types';
+import type { Call } from '../transaction/types';
+
+export type SendSwapTransactionParams = {
+  config: Config;
+  // biome-ignore lint: cannot find module 'wagmi/experimental/query'
+  sendCallsAsync: any;
+  sendTransactionAsync: SendTransactionMutateAsync<Config, unknown>;
+  setCallsId: Dispatch<SetStateAction<Hex | undefined>>;
+  updateLifecycleStatus: (state: LifecycleStatusUpdate) => void;
+  transactions: Call[];
+  walletCapabilities: WalletCapabilities;
+};
+
+export type SendSingleTransactionsParams = {
+  config: Config;
+  sendTransactionAsync: SendTransactionMutateAsync<Config, unknown>;
+  updateLifecycleStatus: (state: LifecycleStatusUpdate) => void;
+  transactions: Call[];
+};
 
 /**
  * Note: exported as public Type
--- a/src/swap/types.ts
+++ b/src/swap/types.ts
@@ -20,17 +20,17 @@ export type SendSwapTransactionParams = {
   // biome-ignore lint: cannot find module 'wagmi/experimental/query'
   sendCallsAsync: any;
   sendTransactionAsync: SendTransactionMutateAsync<Config, unknown>;
-  setCallsId: Dispatch<SetStateAction<Hex | undefined>>;
+  setCallsId: Dispatch<SetStateAction<Hex | undefined>>; // For atomic batched transactions only, used in `useCallsStatus`
+  transactions: Call[]; // A list of transactions to execute
   updateLifecycleStatus: (state: LifecycleStatusUpdate) => void;
-  transactions: Call[];
-  walletCapabilities: WalletCapabilities;
+  walletCapabilities: WalletCapabilities; // EIP-5792 wallet capabilities
 };
 
 export type SendSingleTransactionsParams = {
   config: Config;
   sendTransactionAsync: SendTransactionMutateAsync<Config, unknown>;
+  transactions: Call[]; // A list of transactions to execute
   updateLifecycleStatus: (state: LifecycleStatusUpdate) => void;
-  transactions: Call[];
 };
 
 /**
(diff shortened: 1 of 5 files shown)


src/swap/utils/sendSwapTransactions.ts
```diff
@@ -0,0 +1,31 @@
...
+  transactions,
+}: SendSwapTransactionParams) {
+  if (walletCapabilities[Capabilities.AtomicBatch]?.supported) {
+    // For batched transactions, we'll use `SwapProvider` to listen for calls to emit the `success` state
+    updateLifecycleStatus({
+      statusName: 'transactionPending',
+    });
+    const callsId = await sendCallsAsync({
+      calls: transactions,
+    });
```
0xAlec, line 21: atomic batched transactions does not have a `transactionApproved` status since the approval transaction is part of the batch

0xAlec: (discussed in API meeting)

alessey: we need to modify the [swap messages](https://github.com/coinbase/onchainkit/blob/main/src/swap/utils/getSwapMessage.ts#L20-L26), currently transactionPending shows a confirm in wallet message, we only show Swap in progress on transactionApproved.        I think in this case, we want to show swap in progress without being in the approved status.  Could possibly add a walletApproved status that would also switch to the swap in progress message, thoughts?

0xAlec: alternatively, we can have a third type for `transactionApproved`, but that's not how it's typically used today (`transactionApproved` is emitted after the transactions are confirmed on the blockchain)

```
statusName: 'transactionApproved',
statusData: {
   transactionType: 'batched'
}
```

0xAlec: IMO we should get rid of the `Confirm in Wallet` state entirely, and the UI states should all just say `Transaction Pending` until the transaction has been confirmed, especially with the introduction of session keys

0xAlec: https://github.com/coinbase/onchainkit/pull/1270/commits/3e8f7080a97972a0a58e587c96f480fc12fd9206

how does this look? now we use `transactionApproved` as a wallet approval in both cases

0xAlec: and we are now using `transactionApproved` to mean a wallet approval

src/swap/utils/sendSingleTransactions.ts
```diff
@@ -0,0 +1,59 @@
...
+    }
+    // 2nd to last transaction is the `ERC20` approval
+    if (i === transactions.length - 2) {
+      updateLifecycleStatus({
+        statusName: 'transactionApproved',
+        statusData: {
+          transactionHash: txHash,
+          transactionType: 'ERC20',
+        },
+      });
```
alessey, line 41: Going forward, we may want to persist these individual transaction approvals so lifecycleState ends up as [{transactionHash, transactionType}, ...]

alessey (approved):

src/swap/types.ts
```diff
@@ -8,6 +13,25 @@ import type {
 import type { SendTransactionMutateAsync } from 'wagmi/query';
 import type { BuildSwapTransaction, RawTransactionData } from '../api/types';
 import type { Token } from '../token/types';
+import type { Call } from '../transaction/types';
+
+export type SendSwapTransactionParams = {
```
cpcramer, line 18: nit: Add the comment 

```
/**
 */
```

src/swap/types.ts
```diff
@@ -8,6 +13,25 @@ import type {
...
+  // biome-ignore lint: cannot find module 'wagmi/experimental/query'
+  sendCallsAsync: any;
+  sendTransactionAsync: SendTransactionMutateAsync<Config, unknown>;
+  setCallsId: Dispatch<SetStateAction<Hex | undefined>>;
+  updateLifecycleStatus: (state: LifecycleStatusUpdate) => void;
+  transactions: Call[];
+  walletCapabilities: WalletCapabilities;
+};
+
+export type SendSingleTransactionsParams = {
```
cpcramer, line 29:
```
/**
 */
```

src/swap/types.ts
```diff
@@ -8,6 +13,25 @@ import type {
...
+import type { Call } from '../transaction/types';
+
+export type SendSwapTransactionParams = {
+  config: Config;
+  // biome-ignore lint: cannot find module 'wagmi/experimental/query'
+  sendCallsAsync: any;
+  sendTransactionAsync: SendTransactionMutateAsync<Config, unknown>;
+  setCallsId: Dispatch<SetStateAction<Hex | undefined>>;
+  updateLifecycleStatus: (state: LifecycleStatusUpdate) => void;
+  transactions: Call[];
```
cpcramer, line 25: nit: alphabetical

src/swap/types.ts
```diff
@@ -8,6 +13,25 @@ import type {
...
+  walletCapabilities: WalletCapabilities;
+};
+
+export type SendSingleTransactionsParams = {
+  config: Config;
+  sendTransactionAsync: SendTransactionMutateAsync<Config, unknown>;
+  updateLifecycleStatus: (state: LifecycleStatusUpdate) => void;
+  transactions: Call[];
```
cpcramer, line 33: nit: alphabetical

cpcramer (approved):

src/swap/types.ts
0xAlec: these are not exported as public types

src/swap/types.ts
0xAlec: these are not exported as public types

src/swap/types.ts
0xAlec: done @abcrane123

src/swap/utils/sendSingleTransactions.ts
0xAlec: let's discuss with Leo offline and will add in a future PR - relatively trivial change

cpcramer (approved):

abcrane123 (approved):

src/swap/utils/sendSingleTransactions.ts
```diff
@@ -0,0 +1,59 @@
...
+  for (let i = 0; i < transactions.length; i++) {
+    const tx = transactions[i];
+    updateLifecycleStatus({
+      statusName: 'transactionPending',
+    });
+    const txHash = await sendTransactionAsync(tx);
+
+    if (transactions.length === 3) {
+      // Permit2 has 3 transactions, 2nd to last is the `Permit2` approval
+      if (i === transactions.length - 3) {
```
Zizzamia, line 23: Why do this kind of math? is the index always different?

src/swap/utils/sendSingleTransactions.ts
Zizzamia: Oh I see, `transactionHash` get overridden each time.

I important is for us to store those over time? Or is this more information we give to the devs and the devs can decide to store or not?

In the Transaction expeirence we use the word `transactionHashList: Address[];`, which i can see here not quite working because the different types of transaction.

@0xAlec @alessey so first question to agree on, are those transactionHash, important to store?

alessey: besides just being transparent as to what is going on behind the scenes, i'm not really sure what you would do with them.

PR #1272, opened 2024-09-17
feat: add `atomicBatch` Swap transactions with ERC-20 `from` token

0xAlec:
**What changed? Why?**

- add `useAwaitCalls` hook
- add support for atomic batching
- add 2 other `transactionType` for `transactionApproved` lifecycleStatus
  - `Batched` (for atomic batched transactions with a `callId` instead of a `txHash`)
  - `Swap` (for the swap transaction)

we will now use `transactionApproved` to explicitly indicate to the developer a **_wallet approval_** by the end user

these changes will allow smart wallets to execute a swap from an ERC-20 in a single transaction (as opposed to 3 previously)

**Notes to reviewers**

**How has it been tested?**
playground and unit tests


src/swap/utils/sendSwapTransactions.test.ts
```diff
@@ -61,21 +59,20 @@ describe('sendSwapTransactions', () => {
...
-    expect(setCallsId).toHaveBeenCalledWith('callsId');
     expect(updateLifecycleStatus).toHaveBeenCalledTimes(2);
     expect(updateLifecycleStatus).toHaveBeenNthCalledWith(1, {
       statusName: 'transactionPending',
     });
     expect(updateLifecycleStatus).toHaveBeenNthCalledWith(2, {
       statusName: 'transactionApproved',
       statusData: {
+        callsId: 'callsId',
         transactionType: 'Batched',
```
0xAlec, line 76: @alessey if we want to use `lifecycle`, let's emit the `callsId` as part of `transactionApproved`

0xAlec: I think this is fine since with a `callsId` as a developer, you can still use this to query the calls status yourself using the RPC `wallet_getCallsReceipt` (granted you know which wallet server to use)

0xAlec: I would imagine we can include a field here based on the Wagmi connector, e.g. `walletServer` or `walletConnector` if the developer wants to use these wallet-specific fields

src/swap/components/SwapProvider.tsx
```diff
@@ -60,7 +65,7 @@ export function SwapProvider({
     },
   }); // Component lifecycle
 
-  // Update lifecycle status, statusData will be persisted for the full lifecycle
+  // Update lifecycle status, statusData will be persisted for the full lifeCycle
```
alessey, line 68: its lifecycle now

0xAlec: oops - my cherry pick loaded the old version

pushed a fix

src/swap/hooks/useAwaitCalls.ts
```diff
@@ -0,0 +1,46 @@
...
+    },
+  });
+
+  return useCallback(async () => {
+    if (data?.status === 'CONFIRMED' && data?.receipts) {
+      const transactionReceipt = await waitForTransactionReceipt(
+        accountConfig,
+        {
+          confirmations: 1,
+          hash: data.receipts[data.receipts.length - 1].transactionHash,
```
alessey, line 33: how many receipts would be here?

src/swap/hooks/useAwaitCalls.ts
```diff
@@ -0,0 +1,46 @@
...
+        },
+      );
+      setLifecycleStatus({
+        statusName: 'success',
+        statusData: {
+          isMissingRequiredField: false,
+          maxSlippage: config.maxSlippage,
```
alessey, line 40: i don't think you should need to pass these shared fields

src/swap/components/SwapProvider.tsx
```diff
@@ -86,9 +91,17 @@ export function SwapProvider({
   const [hasHandledSuccess, setHasHandledSuccess] = useState(false);
   const { from, to } = useFromTo(address);
   const { sendTransactionAsync } = useSendTransaction(); // Sending the transaction (and approval, if applicable)
+  const { sendCallsAsync } = useSendCalls(); // Atomic Batch transactions (and approval, if applicable)
```
cpcramer, line 94: In what situations is `approval` applicable?

src/swap/utils/processSwapTransaction.test.ts
```diff
@@ -120,10 +134,9 @@ describe('processSwapTransaction', () => {
       statusName: 'transactionPending',
     });
     expect(updateLifecycleStatus).toHaveBeenNthCalledWith(4, {
-      statusName: 'transactionApproved',
+      statusName: 'success',
```
alessey, line 137: shouldn't we still be emitting transactionApproved here before success?

src/swap/types.ts
```diff
@@ -345,3 +348,12 @@ export type Transaction = {
   to: Address; // The recipient address
   value: bigint; // The value of the transaction
 };
+
+export type UseAwaitCallsParams = {
+  accountConfig: Config;
+  config: {
+    maxSlippage: number;
+  };
```
cpcramer, line 356: Here we can use the type `SwapConfig`:

```
type SwapConfig = {
  maxSlippage: number; // Maximum acceptable slippage for a swap. (default: 10) This is as a percent, not basis points;
};
```


https://github.com/coinbase/onchainkit/blob/main/src/swap/types.ts#L189-191

src/swap/hooks/useAwaitCalls.ts
0xAlec: these are required in the typing - should we make them optional?

cc: @cpcramer

src/swap/components/SwapProvider.tsx
0xAlec: when `from` is ERC-20

src/swap/utils/processSwapTransaction.test.ts
0xAlec: good catch, updated in this PR

src/swap/hooks/useAwaitCalls.ts
0xAlec: only 1 I believe... I'm not sure which cases have more than 1 receipt

0xAlec: but default to returning the last in those cases

src/swap/components/SwapProvider.tsx
0xAlec: the approval can be in the batch, but not all batches will contain approvals

src/swap/types.ts
```diff
@@ -103,8 +105,9 @@ export type LifecycleStatus =
   | {
       statusName: 'transactionApproved';
       statusData: {
+        callsId?: Hex;
         transactionHash?: Hex;
-        transactionType: 'ERC20' | 'Permit2' | 'Batched';
+        transactionType: 'Batched' | 'ERC20' | 'Permit2' | 'Swap';
```
Zizzamia, line 110: Where are we documenting the difference between them?

I feel like we should add some comments here to expand how those types are different.

src/swap/types.ts
```diff
@@ -151,11 +154,16 @@ export type LifecycleStatusUpdate = LifecycleStatus extends infer T
...
 export type ProcessSwapTransactionParams = {
+  chainId: number; // The chain ID
   config: Config;
-  updateLifecycleStatus: (state: LifecycleStatusUpdate) => void;
+  // biome-ignore lint: cannot find module 'wagmi/experimental/query'
+  sendCallsAsync: any;
   sendTransactionAsync: SendTransactionMutateAsync<Config, unknown>;
-  swapTransaction: BuildSwapTransaction;
+  swapTransaction: BuildSwapTransaction; // The response from the Swap API
+  switchChainAsync: SwitchChainMutateAsync<Config, unknown>;
```
Zizzamia, line 163: why do we need this?

src/swap/types.ts
```diff
@@ -348,3 +356,10 @@ export type Transaction = {
   to: Address; // The recipient address
   value: bigint; // The value of the transaction
 };
+
+export type UseAwaitCallsParams = {
+  accountConfig: Config;
+  config: SwapConfig;
+  lifecycleStatus: LifecycleStatus;
+  setLifecycleStatus: React.Dispatch<React.SetStateAction<LifecycleStatus>>;
```
Zizzamia, line 364: I believe we don't use `setLifecycleStatus` like this anymore. At this point passing around `setLifecycleStatus` is an antipattern we should avoid at all coast.

Right @alessey ?

src/swap/utils/getTransactionType.ts
```diff
@@ -0,0 +1,21 @@
+import type { Call } from '../../transaction/types';
+
+export function getTransactionType(
+  transactions: Call[],
+  index: number,
+): 'ERC20' | 'Permit2' | 'Swap' | null {
```
Zizzamia, line 6: We are using this type in a few places, maybe we want to abstract this.

src/swap/utils/getTransactionType.ts
```diff
@@ -0,0 +1,21 @@
...
+    return 'Permit2';
+  }
+  // 2nd to last transaction is the `ERC20` approval
+  if (index === transactions.length - 2) {
+    return 'ERC20';
+  }
+  // Last transaction is the swap itself
+  if (index === transactions.length - 1) {
+    return 'Swap';
+  }
```
Zizzamia, line 18: What about `batch`?

src/swap/utils/processSwapTransaction.ts
```diff
@@ -1,56 +1,46 @@
...
   const { transaction, approveTransaction, quote } = swapTransaction;
+  const transactions: Call[] = [];
 
   // for swaps from ERC-20 tokens,
   // if there is an approveTransaction present,
   // request approval for the amount
   // for V1 API, `approveTx` will be an ERC-20 approval against the Router
   // for V2 API, `approveTx` will be an ERC-20 approval against the `Permit2` contract
   if (approveTransaction?.data) {
-    updateLifecycleStatus({
```
Zizzamia, line 25: We lost all the `transactionPending`, what's the thinking behind this decision?

Zizzamia:
> we will now use transactionApproved to explicitly indicate to the developer a wallet approval by the end user

How this state changes as behavior between EOA and Smart Wallet? Also is this a Breaking changes? It feels like it.

src/swap/types.ts
0xAlec: added comment documenting the different types

src/swap/types.ts
0xAlec: modified to use `updateLifecycleStatus`

src/swap/utils/getTransactionType.ts
0xAlec: added shared type

src/swap/utils/getTransactionType.ts
0xAlec: we handle batch differently - it will return type `Batched` with a `callsId` field.

src/swap/utils/processSwapTransaction.ts
0xAlec: we still have `transactionPending` - we have moved this to the `sendSwapTransactions` function

`processSwapTransaction` is now only responsible for building the `transactions` array

0xAlec:
> > we will now use transactionApproved to explicitly indicate to the developer a wallet approval by the end user
> 
> How this state changes as behavior between EOA and Smart Wallet? Also is this a Breaking changes? It feels like it.

I don't think this is a breaking change. Behavior is the same for EOA. For Smart Wallets now, all 3 transactions are batched to a single pop-up.

EOA:
**pending -> approved -> pending -> approved -> pending -> approved -> success**

Smart Wallet:
before:
**pending -> approved -> pending -> approved -> pending -> approved -> success**

now:
**pending -> approved -> success**

src/swap/hooks/useAwaitCalls.ts
abcrane123: what is the difference between the receipts contained in data object returned from `useCallsStatus` and the receipt returned from `waitForTransactionReceipt` here?

0xAlec: the former is a different object, we're expecting a `TransactionReceipt`

```
export type WalletCallReceipt<quantity = Hex, status = Hex> = {
  logs: {
    address: Hex
    data: Hex
    topics: Hex[]
  }[]
  status: status
  blockHash: Hex
  blockNumber: quantity
  gasUsed: quantity
  transactionHash: Hex
}
```

abcrane123: is there any data the user gets from second receipt that they wouldn't be able to get from first receipt ? just asking because i remember a conversation with Wilson where he brought up that it might not be worth the extra latency from second request. sending you the slack link in DM.

src/swap/utils/getTransactionType.ts
alessey: It is a little weird to need to infer the type from the position in the array when we know the type when adding it to the array.  If something changed in processSwap seems like this could get out of sync.  Could possibly create a transactionTypes array at the same time as the transactions array and reference those types by index in sendSingle.  It is nice to be able to directly pass the transactions without modification.

src/swap/components/SwapProvider.tsx
```diff
@@ -6,14 +6,19 @@ import {
   useState,
 } from 'react';
 import { useAccount, useConfig, useSendTransaction } from 'wagmi';
+import { useChainId } from 'wagmi';
+import { useSwitchChain } from 'wagmi';
```
cpcramer, line 10:
```suggestion
import { useChainId, useSwitchChain } from 'wagmi';
```

src/swap/types.ts
```diff
@@ -103,8 +105,9 @@ export type LifecycleStatus =
   | {
       statusName: 'transactionApproved';
       statusData: {
+        callsId?: Hex;
         transactionHash?: Hex;
-        transactionType: 'ERC20' | 'Permit2' | 'Batched';
+        transactionType: SwapTransactionType;
```
cpcramer, line 110: nice

src/swap/utils/getTransactionType.ts
abcrane123: i agree with Adam here but would be okay if you add as fast follow

cpcramer (approved):

src/swap/utils/sendSingleTransactions.ts
```diff
@@ -11,36 +11,15 @@ export async function sendSingleTransactions({
   let transactionReceipt: TransactionReceipt | undefined;
 
   // Execute the non-batched transactions sequentially
-  for (let i = 0; i < transactions.length; i++) {
-    const tx = transactions[i];
+  for (const { transaction, transactionType } of transactions) {
```
abcrane123, line 14: nice

abcrane123 (approved):

PR #1281, opened 2024-09-19
feat: Add SwapToast component

abcrane123:
**What changed? Why?**
- add `SwapToast` component and display on success

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/internal/utils/getToastPosition.ts
+++ b/src/internal/utils/getToastPosition.ts
@@ -1,0 +1,12 @@
+export function getToastPosition(position: string) {
+  if (position === 'bottom-right') {
+    return 'bottom-5 left-3/4';
+  }
+  if (position === 'top-right') {
+    return 'top-[100px] left-3/4';
+  }
+  if (position === 'top-center') {
+    return 'top-[100px] left-2/4';
+  }
+  return 'bottom-5 left-2/4';
+}
--- a/src/swap/components/SwapProvider.tsx
+++ b/src/swap/components/SwapProvider.tsx
@@ -70,6 +70,8 @@ export function SwapProvider({
     },
   }); // Component lifecycle
 
+  const [isToastVisible, setIsToastVisible] = useState(false);
+
   // Update lifecycle status, statusData will be persisted for the full lifecycle
   const updateLifecycleStatus = useCallback(
     (newStatus: LifecycleStatusUpdate) => {
@@ -117,6 +119,7 @@ export function SwapProvider({
     if (lifecycleStatus.statusName === 'success') {
       onSuccess?.(lifecycleStatus.statusData.transactionReceipt);
       setHasHandledSuccess(true);
+      setIsToastVisible(true);
     }
     // Emit Status
     onStatus?.(lifecycleStatus);
@@ -378,6 +381,8 @@ export function SwapProvider({
     lifecycleStatus,
     updateLifecycleStatus,
     to,
+    isToastVisible,
+    setIsToastVisible,
   });
 
   return <SwapContext.Provider value={value}>{children}</SwapContext.Provider>;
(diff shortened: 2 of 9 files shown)


src/swap/components/SwapToast.tsx
```diff
@@ -0,0 +1,94 @@
...
+    return null;
+  }
+
+  return (
+    <div
+      className={cn(
+        background.default,
+        'flex animate-enter items-center justify-between rounded-lg',
+        'p-2 shadow-[0px_8px_24px_0px_rgba(0,0,0,0.12)]',
+        '-translate-x-2/4 fixed z-20',
```
alessey, line 61: the animate in from the right seems weird, also, should this be pegged to the swap component vs. the window?

src/swap/components/SwapProvider.tsx
```diff
@@ -70,6 +70,9 @@ export function SwapProvider({
     },
   }); // Component lifecycle
 
+  const [isToastVisible, setIsToastVisible] = useState(false);
+  const [transactionHash, setTransactionHash] = useState('');
```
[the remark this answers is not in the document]
abcrane123: yeah i like that but we would do `setToastContent(transactionHash)` right ?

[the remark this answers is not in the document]
abcrane123: hmm well we don't really need to store the whole component because theyre rendering `SwapToast` as a child (following the pattern of the rest of the components) right? also we could use the toast for other information by checking the `isToastVisible` value along with an error state, for example, stored in `lifeCycleStatus`. The reason i had to do the extra state for `transactionHash` is because we immediately reset the `lifeCycleStatus` to `init` after calling onsuccess. 

curious what the rest of the team thinks about this @0xAlec @cpcramer 

my thinking is that if we want to do something like you're suggesting we should follow the same pattern with TransactionToast and maybe that should be a follow up.

src/swap/components/SwapToast.tsx
abcrane123: so will investigate but it appears like it was an existing issue with how we style our toasts

src/swap/components/SwapProvider.tsx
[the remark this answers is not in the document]
alessey: I think it makes sense regardless to not have the toast showing content directly from lifecycle.  If we want to toast an error, we don't want the toast to clear if the error clears, but I don't want to end up with a bunch of separate states to hold different things we want to show in the toast.   my initial thought was that success toast could store transactionHash in its internal state and encapsulate the chainExplorer logic, and then just get displayed as a child of the toast.

0xAlec (approved):

alessey (approved):

src/internal/utils/getToastPosition.ts
```diff
@@ -0,0 +1,12 @@
+export function getToastPosition(position: string) {
```
cpcramer, line 1: nice

src/swap/types.ts
```diff
@@ -213,6 +213,10 @@ export type SwapContextType = {
   handleToggle: () => void;
   updateLifecycleStatus: (state: LifecycleStatusUpdate) => void; // A function to set the lifecycle status of the component
   to: SwapUnit;
+  isToastVisible?: boolean;
+  setIsToastVisible?: (visible: boolean) => void;
```
cpcramer, line 217: I think a larger discussion, but how are we determining when state should live / be handled in the SwapContext vs. LifecycleStatus shared? cc @alessey

alessey: I think the toasts are a case where we cannot rely on lifecycle status in case we want to display a toast longer than a lifecycle status persists.  Although I would prefer to encapsulate the toast state within its own component or a single toast state.

alessey: re: https://github.com/coinbase/onchainkit/pull/1281#discussion_r1773436801

PR #1285, opened 2024-09-22
feat: integrate Coinbase Onramp for funding EOA wallets

steveviselli-cb:
**What changed? Why?**

The current `WalletDropdownFundLink` always takes users to `https://keys.coinbase.com/fund` regardless of wether the connected wallet is actually a Coinbase Smart Wallet or not. This PR updates the `WalletDropdownFundLink` to only take users to `https://keys.coinbase.com/fund` if the connected wallet is a Coinbase Smart Wallet. Otherwise it will send users to Coinbase Onramp i.e. `https://pay.coinbase.com/buy`.

**Notes to reviewers**

I refactored `WalletDropdownFundLink` into a few smaller components to reduce component complexity and make testing much easier. But I did not modify the original code for rendering a Coinbase Smart Wallet fund link.

I'm using the combination of Coinbase Wallet Wagmi connector ID (defined [here](https://github.com/wevm/wagmi/blob/472993b3c0d5941e524e67a0f51b32af5bb9e48f/packages/connectors/src/coinbaseWallet.ts#L90)) and atomicBatch capability support to determine if the connected wallet is a Coinbase Smart Wallet.

For users in the US, the Coinbase Onramp UI will give users the option to continue as a guest if they do not have a Coinbase account. You can't see it in the below video because I'm in Canada.

**How has it been tested?**

Tested in playground.

https://github.com/user-attachments/assets/fd74f9b1-a536-4afd-b509-83236501429a


src/wallet/components/WalletDropdownFundLinkButton.tsx
```diff
@@ -0,0 +1,82 @@
+import { useCallback, useMemo } from 'react';
```
steveviselli-cb, line 1: This whole component was just copy/pasted out of the original fund link component.

src/fund/utils/getCoinbaseSmartWalletFundUrl.ts
```diff
@@ -0,0 +1,21 @@
+import { version } from '../../version';
```
steveviselli-cb, line 1: This code was just extracted out of the original wallet dropdown fund link and moved into a util.

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -14,69 +12,38 @@ export function WalletDropdownFundLink({
...
-      <>
-        <div className="-translate-y-1/2 absolute top-1/2 left-4 flex h-[1.125rem] w-[1.125rem] items-center justify-center">
-          {iconSvg}
-        </div>
-        <span className={cn(themeText.body, 'pl-6')}>{text}</span>
-      </>
-    ),
-    [iconSvg, text],
-  );
+  if (fundingUrl) {
```
0xAlec, line 19: I'm not sure I like this design approach with the conditional rendering of 3 different components (and prop destructuring)

I think we should keep this file, and switch the `fundingUrl` and `handleClick` behavior based on the presence of `walletCapabilities`

cc: @alessey @alissacrane-cb what do you guys think?

src/wallet/components/WalletDropdownFundLinkEOAWallet.tsx
```diff
@@ -0,0 +1,38 @@
...
+      addresses: { [address]: [chain.name.toLowerCase()] },
+    });
+  }, [projectId, address, chain]);
+
+  return (
+    // The Coinbase Onramp widget is not very responsive, so we need to override the popup size.
+    <WalletDropdownFundLinkButton
+      {...props}
+      popupHeightOverride={720}
+      popupWidthOverride={460}
```
alessey, line 34: will this only work at these sizes?  would be nice to respect popupSize in both these implementations.  Is this viewable on mobile?

src/wallet/components/WalletDropdownFundLink.tsx
steveviselli-cb: I'm no FE expert, but the React docs say that using an effect to update state or cache expensive calculations is an anti-pattern (https://react.dev/learn/you-might-not-need-an-effect). That's why I wrapped getting the URLs in useMemo calls.

The conditional rendering is really for maintainability reasons. The fund link component was already getting kinda complex, I didn't want to add even more logic to it. You can imagine that if we add more funding integrations that component could get really out of hand.

I feel like separating the logic of getting the right URL from the logic of actually rendering the button that handles dealing with the popup / new tab opening is a clear separation of concern, and makes it way easier to keep our unit test coverage at 100%.

src/wallet/components/WalletDropdownFundLinkEOAWallet.tsx
steveviselli-cb: Unfortunately yeah, the popup size doesn't even matter, Onramp's view is a hardcoded size, so it'll either have a bunch of empty space, or a bunch of overflow if the popup size doesn't match.

I don't think anyone should be using popup on mobile...

src/wallet/components/WalletDropdownFundLink.tsx
steveviselli-cb: Oh and I can get rid of the props spread if you want, I was mainly doing it because those props should be a pass through for the wallet specific components. And those components are not exported, and the props are all strongly typed, so I'm not really sure why it would be an anti-pattern here. But I'm happy to make it all verbose if that's what we want. We should probably turn on the linting rule for no prop spreading if that's the case.

src/wallet/components/WalletDropdownFundLinkEOAWallet.tsx
alessey: we'll have to communicate somehow that popupSize is ignored for EOA Wallets as otherwise this will seem like a bug

steveviselli-cb: @alessey yeah good call, I added a jsdoc note to the type, will also add it to the doc site later.

src/wallet/components/WalletDropdownFundLink.tsx
steveviselli-cb: @alessey @alissacrane-cb would love your thoughts on this architecture!

alessey: I agree with the goal of extracting the funding url logic, but I think you could reduce boiler plate and keep your separation of concerns having a useFundingUrl hook which called the appropriate util.

steveviselli-cb: @alessey yeah I like that idea, I've refactored this.

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -14,69 +10,39 @@ export function WalletDropdownFundLink({
...
+  const defaultFundingUrl = useGetFundingUrl();
+
+  // If the fundingUrl prop is undefined, fallback to the default funding URL
+  const fundingUrlToRender = fundingUrl ?? defaultFundingUrl?.url;
+  const popupHeightOverride = fundingUrl
+    ? undefined
+    : defaultFundingUrl?.popupHeight;
+  const popupWidthOverride = fundingUrl
+    ? undefined
+    : defaultFundingUrl?.popupWidth;
```
alessey, line 24: const popupHeightOverride = fundingUrl?.popupHeight would end up the same right?

src/fund/hooks/useGetFundingUrl.test.ts
```diff
@@ -0,0 +1,136 @@
...
+vi.mock('../../fund/utils/getCoinbaseSmartWalletFundUrl', () => ({
+  getCoinbaseSmartWalletFundUrl: vi.fn(),
+}));
+
+vi.mock('../../fund/utils/getOnrampBuyUrl', () => ({
+  getOnrampBuyUrl: vi.fn(),
+}));
+
+describe('useGetFundingUrl', () => {
+  it('returns Coinbase Smart Wallet fund URL if connected wallet is a Coinbase Smart Wallet', () => {
```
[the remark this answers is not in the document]
steveviselli-cb: Fixed!

src/wallet/components/WalletDropdownFundLink.tsx
steveviselli-cb: @alessey Na, even if they provide a fundingUrl prop we still fetch the defaultFunding URL (can't conditionally call a hook) so we'd risk overriding the popup size when the use provides a fundingUrl prop.

alessey: true, you could pass in fundingUrl to your useFundingUrl to short circuit the whole thing, which would reduce this component to a single line hook call and may remove the need for it.

steveviselli-cb: Yeah but that would be a really weird interface i.e. `useFundingUrl(fundingUrl: string): string`. Why would you pass a fundingUrl to a hook to get a fundingUrl? This might not be super pretty, but from an interface perspective I think it makes the most sense.

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -14,69 +10,39 @@ export function WalletDropdownFundLink({
   rel,
   target,
   text = 'Fund wallet',
+  fundingUrl,
```
alessey, line 13: since all these defaults were moved to WalletDropdownFundLinkButton, do we need them here?  could { fundingUrl, ...rest }.  This is less explicit, but imo nice as it shows this component only acts on fundingUrl.  @0xAlec better than the destructure spread?

src/wallet/components/WalletDropdownFundLink.tsx
alessey: true, with those names it would be odd, the purpose would be ensuring fundingUrl is a consistent data type.

src/wallet/components/WalletDropdownFundLink.tsx
alessey: perhaps, personally I don't consider prop spreading an anti-pattern, especially with typescript.

steveviselli-cb: @alessey I feel the same, but every other component in OnchainKit avoid doing this and I think consistency is very important. Probably something to align on as a team rather than in a PR, and then either add the no props spread eslint rule or don't and say its ok.

src/wallet/components/WalletDropdownFundLink.tsx
steveviselli-cb: @alessey can you think of better names? Even with other names, the point of the param is to basically tell the hook to do nothing because we can't conditionally call the hook. Do you think this is a blocker for the PR?

alessey: useFundingUrlConfig could work.  I do like the idea of being able to ditch this wrapper component and simplify the override logic, but no, I do not think this is a blocker.

src/wallet/components/WalletDropdownFundLink.tsx
steveviselli-cb: @alessey ok updated the props!

alessey (approved):

src/OnchainKitProvider.tsx
```diff
@@ -16,6 +16,7 @@ export function OnchainKitProvider({
   children,
   rpcUrl,
   schemaId,
+  projectId,
```
cpcramer, line 19: nit: alphabetize params

src/OnchainKitProvider.tsx
```diff
@@ -28,10 +29,11 @@ export function OnchainKitProvider({
       chain: chain,
       rpcUrl: rpcUrl ?? null,
       schemaId: schemaId ?? null,
+      projectId: projectId ?? null,
     };
     setOnchainKitConfig(onchainKitConfig);
     return onchainKitConfig;
-  }, [address, apiKey, chain, rpcUrl, schemaId]);
+  }, [address, apiKey, chain, rpcUrl, schemaId, projectId]);
```
cpcramer, line 36: alphabetical

src/OnchainKitProvider.tsx
```diff
@@ -28,10 +29,11 @@ export function OnchainKitProvider({
       chain: chain,
       rpcUrl: rpcUrl ?? null,
       schemaId: schemaId ?? null,
+      projectId: projectId ?? null,
```
cpcramer, line 32: alphabetical

src/fund/constants.ts
```diff
@@ -0,0 +1,4 @@
+// The recommended height of a Coinbase Onramp popup window.
+export const ONRAMP_WIDGET_HEIGHT = 720;
```
cpcramer, line 2: Can we specify that this is a popup in the naming?

Something like: 

> export const `ONRAMP_POPUP_HEIGHT`

src/fund/constants.ts
```diff
@@ -0,0 +1,4 @@
...
+// The recommended width of a Coinbase Onramp popup window.
+export const ONRAMP_WIDGET_WIDTH = 460;
```
cpcramer, line 4:
```suggestion
export const ONRAMP_POPUP_WIDTH = 460;
```

```suggestion
export const ONRAMP_POPUP_WIDGET_WIDTH = 460;
```

src/fund/hooks/useGetFundingUrl.ts
```diff
@@ -0,0 +1,45 @@
...
+ * user to keys.coinbase.com, otherwise it will send them to pay.coinbase.com.
+ * @returns the funding URL and optional popup dimensions if the URL requires them
+ */
+export function useGetFundingUrl(): UseGetFundingUrlResponse | undefined {
+  const { projectId, chain: defaultChain } = useOnchainKit();
+  const { address, chain: accountChain } = useAccount();
+  const isCoinbaseSmartWallet = useIsWalletACoinbaseSmartWallet();
+
+  // If the connected wallet's active chain is not included in the Wagmi config, accountChain will be undefined. If this
+  // is the case, fall back to the default chain specified in the OnchainKit config.
```
cpcramer, line 21: Love all the comments in general 😄

src/fund/types.ts
```diff
@@ -0,0 +1,81 @@
+/**
+ * Props used to get an Onramp buy URL by directly providing a CDP project ID.
+ * See https://docs.cdp.coinbase.com/onramp/docs/api-initializing#generating-the-coinbase-onramp-buysell-url
+ *
+ * Note: exported as public Type
+ */
+export type GetOnrampUrlWithProjectIdParams = {
```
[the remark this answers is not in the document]
steveviselli-cb: @cpcramer what I'm trying to make clear here is that it's specific to Coinbase Onramp the product. There could be multiple funding providers, actually there are, there's the keys.coinbase.com one and the Onramp one. Each need different params. That's why I think the distinction is important here. "Fund" is the name of the component that is abstracting away the details of what funding provider you are using. Onramp is one of those providers. Does that make sense?

src/fund/constants.ts
steveviselli-cb: Fixed!

src/fund/constants.ts
steveviselli-cb: Fixed!

src/wallet/hooks/useIsWalletACoinbaseSmartWallet.ts
```diff
@@ -0,0 +1,28 @@
...
+  });
+
+  if (
+    connector?.id === COINBASE_WALLET_SDK_CONNECTOR_ID &&
+    walletCapabilities.atomicBatch?.supported === true
+  ) {
+    return true;
+  }
+
+  return false;
```
0xAlec, line 27: nit: can just return this conditional instead

steveviselli-cb: @0xAlec yeah good call, done.

0xAlec (approved):

PR #1286, opened 2024-09-23
feat: Add Swap USD values

cpcramer:
**What changed? Why?**
Add Swap USD values. 

- `formatUSD` function to format the output string.
- Add `fromAmountUSD` and `setFromAmountUSD`
- Add `toAmountUSD` and `setToAmountUSD`

**Notes to reviewers**

**How has it been tested?**


src/swap/components/SwapProvider.tsx
```diff
@@ -272,12 +278,22 @@ export function SwapProvider({
...
           response.to.decimals,
         );
+        const formattedFromAmountUSD = formatTokenAmount(
+          response.fromAmountUSDC,
+          response.to.decimals, // make const = 6?, OR response.to.decimals
+        );
+        const formattedToAmountUSD = formatTokenAmount(
+          response.fromAmountUSDC,
+          response.to.decimals, // make const = 6?, OR response.to.decimals
+        );
```
cpcramer, line 288: Might not need because we have `getRoundedAmount`

cpcramer:
TODO: Balance: 0.01039493 Max is not showing

src/swap/components/SwapAmountInput.tsx
```diff
@@ -64,6 +63,14 @@ export function SwapAmountInput({
   const hasInsufficientBalance =
     type === 'from' && Number(source.balance) < Number(source.amount);
 
+  const formatUSD = (amount: string) => {
+    if (!amount) {
+      return '~$0.00';
+    }
+    const roundedAmount = Number(getRoundedAmount(amount, 2) || 0);
```
alessey, line 70: shouldn't getRoundedAmount always return a value?   when would we hit the 0 fallback?

cpcramer: Good callout, updated!

0xAlec (approved):

abcrane123 (approved):

PR #1293, opened 2024-09-24
feat: add `isSponsored` prop for `Swap`

0xAlec:
**What changed? Why?**
- add `isSponsored` prop

**Notes to reviewers**

**How has it been tested?**


src/swap/components/SwapProvider.tsx
```diff
@@ -48,6 +48,7 @@ export function SwapProvider({
     maxSlippage: DEFAULT_MAX_SLIPPAGE,
   },
   experimental,
+  isSponsored,
```
cpcramer, line 51: default to `false`?

0xAlec: default for booleans are false

src/swap/components/Swap.tsx
```diff
@@ -21,6 +21,7 @@ export function Swap({
   onError,
   onStatus,
   onSuccess,
+  isSponsored,
```
alessey, line 24: could default to false here and not have to handle undefined

0xAlec: I don't think we need special handling for undefined since it's falsey

0xAlec: but I can add a default here

src/swap/components/SwapProvider.tsx
0xAlec: added in latest diff

src/swap/components/Swap.tsx
0xAlec: added in latest diff

cpcramer: I still like explicitly defining as false

src/swap/types.ts
```diff
@@ -267,6 +269,7 @@ export type SwapProviderReact = {
   experimental: {
     useAggregator: boolean; // Whether to use a DEX aggregator. (default: true)
   };
+  isSponsored?: boolean; // Whether the swap is sponsored. (default: false)
```
cpcramer, line 272: nit: starting the comment description with `// An optional...`

src/swap/components/Swap.tsx
```diff
@@ -48,6 +49,7 @@ export function Swap({
       onError={onError}
       onStatus={onStatus}
       onSuccess={onSuccess}
+      isSponsored={isSponsored}
```
cpcramer, line 52: nit: alphabetic imports

src/swap/components/Swap.tsx
```diff
@@ -21,6 +21,7 @@ export function Swap({
   onError,
   onStatus,
   onSuccess,
+  isSponsored = false,
```
cpcramer, line 24: nit: alphabetic

src/swap/types.ts
0xAlec: done

src/swap/types.ts
```diff
@@ -282,6 +285,7 @@ export type SwapReact = {
   experimental?: {
     useAggregator: boolean; // Whether to use a DEX aggregator. (default: true)
   };
+  isSponsored?: boolean; // Whether the swap is sponsored. (default: false)
```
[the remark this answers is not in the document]
0xAlec: done

src/swap/components/Swap.tsx
0xAlec: done - this was from when we called it `sponsored`

src/swap/components/Swap.tsx
0xAlec: done - this was from when we called it `sponsored`

cpcramer (approved):

PR #1314, opened 2024-09-27
Add calls to transaction component docs

mykcryptodev:
**What changed? Why?**
Describe how to pass "calls" to the <Transaction /> component because it currently only demonstrates how to pass "contracts"

**Notes to reviewers**
This change is documentation-only

**How has it been tested?**
Manually viewed it on my local instance of the docs.

### DIFF
--- a/site/docs/pages/transaction/transaction.mdx
+++ b/site/docs/pages/transaction/transaction.mdx
@@ -281,7 +281,8 @@ Using `calls` instead of `contracts` is a great way to bundle several operations
 
 The `calls` prop is an array of `Call` objects, where each `Call` is defined as:
 
-```typescript
+```tsx twoslash
+// @noErrors: 2304 Cannot find name 'Hex'
 type Call = {
   to: Hex;
   data?: Hex;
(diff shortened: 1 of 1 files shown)


0xAlec (approved):

site/docs/pages/transaction/transaction.mdx
```diff
@@ -273,6 +273,70 @@ import { Transaction, TransactionButton, TransactionSponsor } from "@coinbase/on
...
+};
+```
+
+- `to`: The address you are executing the call on
+- `data`: The encoded function data to pass along in the call (optional)
+- `value`: The amount of ether to send along with the call, denominated in wei (optional)
+
+Here's an example of how to use the `calls` prop with the Transaction component:
+
+```tsx
```
alessey, line 298: if you add twoslash it'll enable twoslash for this block, you can use // @noErrors 123 to suppress any errors you don't want to resolve for the example

mykcryptodev: @alessey good call! now using twoslash 🫡

alessey (approved):

PR #1322, opened 2024-09-27
feat: add fund button

steveviselli-cb:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/styles/theme.ts
```diff
@@ -28,7 +28,7 @@ export const pressable = {
   secondary:
     'cursor-pointer bg-ock-secondary active:bg-ock-secondary-active hover:bg-[var(--bg-ock-secondary-hover)]',
   shadow: 'shadow-ock-default',
-  disabled: 'opacity-[0.38]',
+  disabled: 'opacity-[0.38] pointer-events-none',
```
steveviselli-cb, line 31: Removes hover effects for disabled buttons.

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -1,34 +1,79 @@
+import { useCallback, useMemo } from 'react';
 import { useGetFundingUrl } from '../../fund/hooks/useGetFundingUrl';
+import { getPopupSize } from '../../fund/utils/getPopupSize';
+import { openPopup } from '../../internal/utils/openPopup';
+import { cn, pressable, text as themeText } from '../../styles/theme';
+import { useIcon } from '../hooks/useIcon';
 import type { WalletDropdownFundLinkReact } from '../types';
-import { WalletDropdownFundLinkButton } from './WalletDropdownFundLinkButton';
 
 export function WalletDropdownFundLink({
```
steveviselli-cb, line 9: Ended up merging this back into a single component, you guys were right @0xAlec @alessey

src/internal/components/Spinner.tsx
```diff
@@ -13,7 +13,7 @@ export function Spinner({ className }: SpinnerReact) {
       <div
         className={cn(
           'animate-spin border-2 border-gray-200 border-t-3',
-          'rounded-full border-t-blue-500 px-2.5 py-2.5',
```
steveviselli-cb, line 16: Confirmed with @mindapivessa that this was the wrong color.

fakepixels:
Nit - can we increase the sidebar width so smart wallet doesn't become 2 lines? @steveviselli-cb 
Screenshot 2024-09-28 at 11 31 32 AM

src/fund/utils/getPopupSize.ts
```diff
@@ -0,0 +1,29 @@
...
+
+type PopupSize = {
+  height: number;
+  width: number;
+};
+
+/**
+ * Gets the appropriate popup dimensions for the given size and funding URL.
+ */
+export function getPopupSize(
```
alessey, line 16: since this is specific to funding, should this be getFundingPopupSize?

steveviselli-cb:
> Nit - can we increase the sidebar width so smart wallet doesn't become 2 lines?

@fakepixels that's just because I resized the browser window to be small for the video, it's not normally like that.

src/fund/utils/getPopupSize.ts
steveviselli-cb: @alessey Yeah sure I can do that! I figured that having it under `/fund` would make including fund in the name redundant, but probably better to be explicit.

src/wallet/components/WalletDropdownFundLink.tsx
```diff
@@ -1,34 +1,79 @@
...
+      }
+    },
+    [fundingUrlToRender, popupSize, target],
+  );
 
-  if (fundingUrlToRender) {
+  const overrideClassName = cn(
+    pressable.default,
+    // Disable hover effects if there is no funding URL
+    !fundingUrlToRender && 'pointer-events-none',
```
alessey, line 42: with this change, https://github.com/coinbase/onchainkit/pull/1322/files#diff-639b39dc7c8fa5ec11491e030f9add3c94620cae4c45a8e0a9412103659e41dfR31, this shouldn't be necessary right?

mindapivessa:
@steveviselli-cb will smart wallet be a part of a different PR?

steveviselli-cb:
> will smart wallet be a part of a different PR?

@mindapivessa nope smart wallet is in this one too. You can see it in the second video.

src/wallet/components/WalletDropdownFundLink.tsx
steveviselli-cb: @alessey na it is necessary. I don't use `pressable.disabled` on the actual link because that would make the background color change like this:

Screenshot 2024-10-01 at 6 36 25 AM

So I disable the pointer events on the link to prevent the hover effect from `pressable.default` which is on the link, and then apply `pressable.disabled` to the span inside the link so that the text is dimmed to look disabled.

src/fund/utils/getPopupSize.ts
steveviselli-cb: @alessey rename done.

cpcramer (approved):

PR #1339, opened 2024-10-01
fix: swap error message

0xAlec:
**What changed? Why?**
see comments

**Notes to reviewers**

**How has it been tested?**
image

### DIFF
--- a/src/swap/constants.ts
+++ b/src/swap/constants.ts
@@ -18,7 +18,7 @@ export enum SwapMessage {
   FETCHING_BALANCE = 'Fetching balance...',
   INCOMPLETE_FIELD = 'Complete the fields to continue',
   INSUFFICIENT_BALANCE = 'Insufficient balance',
-  LOW_LIQUIDITY = 'Liquidity too low for the token',
+  LOW_LIQUIDITY = 'Insufficient liquidity for this trade.',
   SWAP_IN_PROGRESS = 'Swap in progress...',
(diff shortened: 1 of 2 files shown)


src/swap/constants.ts
```diff
@@ -18,7 +18,7 @@ export enum SwapMessage {
   FETCHING_BALANCE = 'Fetching balance...',
   INCOMPLETE_FIELD = 'Complete the fields to continue',
   INSUFFICIENT_BALANCE = 'Insufficient balance',
-  LOW_LIQUIDITY = 'Liquidity too low for the token',
+  LOW_LIQUIDITY = 'Insufficient liquidity for this trade.',
```
0xAlec, line 21: changed to use the same copy as uniswap

src/swap/utils/getSwapMessage.ts
```diff
@@ -8,6 +8,11 @@ export function getSwapMessage({
   lifecycleStatus,
   to,
 }: GetSwapMessageParams) {
+  // handle specific error codes
+  if (lifecycleStatus.statusName === 'error') {
+    return getErrorMessage(lifecycleStatus.statusData);
+  }
+
```
0xAlec, line 15: INCOMPLETE_FIELD would previously override the error

alessey (approved):

fakepixels (approved):

PR #1349, opened 2024-10-02
feat: [`Pay`] components

0xAlec:
**What changed? Why?**
- add `PayButton`, `Pay`, and `PayStatus`
- add corresponding tests
- add `useGetPayStatus`
- modify `useCommerceContracts`callback to take an `address` as an argument

**Notes to reviewers**

**How has it been tested?**
playground PR next


src/pay/components/PayProvider.tsx
```diff
@@ -0,0 +1,281 @@
...
+            ) || coinbaseWallet({ preference: 'smartWalletOnly' }),
+        });
+        connectedAddress = accounts[0];
+        connectedChainId = _connectedChainId;
+      }
+
+      // This shouldn't ever happen, but added this to make Typescript happy
+      /* v8 ignore next 3 */
+      if (!connectedAddress) {
+        return;
```
0xAlec, line 180: I think maybe we can add an `updateLifecycleStatus` error message here, but almost always there will be `connectedAddress` returned from `connectAsync`

src/pay/components/Pay.tsx
```diff
@@ -0,0 +1,30 @@
...
+    return null;
+  }
+
+  return (
+    <PayProvider
+      chargeHandler={chargeHandler}
+      onStatus={onStatus}
+      productId={productId}
+    >
+      <div className={cn(className, 'flex w-full flex-col gap-2')}>
```
alessey, line 25: className last

src/pay/components/PayProvider.tsx
```diff
@@ -0,0 +1,281 @@
...
+    try {
+      // Open Coinbase Commerce receipt
+      if (lifecycleStatus.statusName === PAY_LIFECYCLESTATUS.SUCCESS) {
+        window.open(
+          `https://commerce.coinbase.com/pay/${chargeId}/receipt`,
+          '_blank',
+          'noopener,noreferrer',
+        );
+        return;
+      }
```
alessey, line 147: may be more straightforward to have separate callbacks for handleFundingFlow and handleViewReceipt and then call different callbacks in the button based on the status

src/pay/components/PayProvider.tsx
```diff
@@ -0,0 +1,281 @@
...
+
+      // Open keys.coinbase.com for payment
+      await writeContractsAsync({
+        contracts,
+      });
+    } catch (error) {
+      if (
+        (error as Error).message?.includes('User denied connection request')
+      ) {
+        setErrorMessage(USER_REJECTED_ERROR);
```
alessey, line 246: shouldn't this also updateLifecycleStatus?

src/pay/components/PayProvider.tsx
0xAlec: I'll refactor this as a fast follow once we have USDC magic spend

0xAlec: I don't think it makes sense to add a `handleFundingFlow.ts` just to delete it in a week or so

alessey (approved):

abcrane123 (approved):

PR #1354, opened 2024-10-03
OCK theming

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/OnchainKitProvider.tsx
```diff
@@ -27,13 +28,15 @@ export function OnchainKitProvider({
...
+      config: config ?? null,
       projectId: projectId ?? null,
       rpcUrl: rpcUrl ?? null,
       schemaId: schemaId ?? null,
     };
     setOnchainKitConfig(onchainKitConfig);
     return onchainKitConfig;
-  }, [address, apiKey, chain, projectId, rpcUrl, schemaId]);
+  }, [address, apiKey, chain, config, projectId, rpcUrl, schemaId]);
+
```
alessey, line 39: thoughts on adding a useTheme hook instead of accessing through useOnchainKit?   I'm on the fence with that, but throwing it out there.

site/docs/pages/guides/tailwind.mdx
```diff
@@ -52,64 +52,64 @@ To override default colorscheme, you need to modify the following css variables:
...
+    --ock-bg-primary: theme(colors.indigo.600);
+    --ock-bg-primary-hover: theme(colors.indigo.700);
+    --ock-bg-primary-active: theme(colors.indigo.800);
+    --ock-bg-secondary: theme(colors.slate.200);
+    --ock-bg-secondary-hover: theme(colors.slate.300);
+    --ock-bg-secondary-active: theme(colors.slate.400);
+    --ock-bg-error: theme(colors.rose.600);
+    --ock-bg-warning: theme(colors.orange.600);
+    --ock-bg-success: theme(colors.lime.300);
+    --ock-bg-default-reverse: theme(colors.gray.950);
```
alessey, line 82: i like this, but could this be a breaking change if users are already overriding our variables?

cpcramer: Yes, it will be a breaking change!  I've been working alongside Mind, I think it's a big improvement that we've wanted to do for some time now. With the new theming work it makes sense to apply these updates now!

src/OnchainKitProvider.tsx
cpcramer: I feel it makes sense to pass the `config` here because the users are passing the parameters directly in the `OnchainKitProvider`.

playground/nextjs-app-router/components/AppProvider.tsx
```diff
@@ -164,7 +188,15 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
...
-      {children}
+      <OnchainKitProvider
+        apiKey={ENVIRONMENT_VARIABLES[ENVIRONMENT.API_KEY]}
+        chain={base}
+        config={{ theme: componentTheme }}
+        projectId={ENVIRONMENT_VARIABLES[ENVIRONMENT.PROJECT_ID]}
+        schemaId="0xf8b05c79f090979bf4a80270aba232dff11a10d9ca55c4f88de95317970f0de9"
+      >
+        {children}
+      </OnchainKitProvider>
```
cpcramer, line 199: I know this is strange. It was the only way I was able to get the componentSwitcher to work in playground. 

Updating the OnchainKitProvider directly caused strange React rendering errors. 

I've implemented this solution to unblock QA and testing in general

src/OnchainKitProvider.tsx
cpcramer: Open to using a useTheme hook if you feel that it would be better!

src/useTheme.ts
```diff
@@ -0,0 +1,76 @@
+import { useEffect, useState } from 'react';
+import type { ComponentTheme, UseThemeReact } from './types';
+import { useOnchainKit } from './useOnchainKit';
+
+export function useTheme(): UseThemeReact {
```
cpcramer, line 5: Theme selection is very complex since we have so many variations based on Theme and Mode while also keeping a clean input for the developers - one `default` theme.

src/types.ts
```diff
@@ -18,14 +18,32 @@ export type isEthereumOptions = {
...
  */
 export type OnchainKitConfig = {
   address: Address | null; // Address is optional as we may not have an address for new users
   apiKey: string | null; // ApiKey for Coinbase Developer Platform APIs
-  rpcUrl: string | null; // RPC URL for onchain requests. Defaults to using CDP Node if the API Key is set
   chain: Chain; // Chain must be provided as we need to know which chain to use
+  appearance?: {
+    mode?: Mode; // Optionally determines color scheme based on OS preference or user selection
+    theme?: ComponentTheme; // Optionally sets the visual style for components
+  };
```
cpcramer, line 45: appearance as a whole is optional. Internally we default:

```
mode: 'auto'
theme: 'default-light'
```

src/types.ts
```diff
@@ -18,14 +18,32 @@ export type isEthereumOptions = {
...
+export type ComponentTheme = 'base' | 'cyberpunk' | 'default' | 'minimal';
+
+// Internal theme options, including light/dark variants for 'default'
+export type UseThemeReact =
+  | 'base'
+  | 'cyberpunk'
+  | 'default'
+  | 'minimal'
+  | 'default-light'
+  | 'default-dark';
```
cpcramer, line 33: We use explicit light or dark class naming internally

src/types.ts
```diff
@@ -18,14 +18,32 @@ export type isEthereumOptions = {
   isMainnetOnly?: boolean; // If the chainId check is only allowed on mainnet
 };
 
+export type Mode = 'auto' | 'light' | 'dark';
+
+// External theme options for users
+export type ComponentTheme = 'base' | 'cyberpunk' | 'default' | 'minimal';
```
cpcramer, line 24: Keep theme options simple for dev experience

playground/nextjs-app-router/app/globals.css
```diff
@@ -59,11 +59,24 @@
...
+  :root,
+  .default-light,
+  .default-dark,
+  .base,
+  .cyberpunk,
+  .minimal {
+    --ock-bg-primary: red;
+  }
+} */
+
```
cpcramer, line 74: How to customize your theme or create a new one - you'll need to override the specific `--ock` variable

src/fund/components/FundButton.tsx
```diff
@@ -18,6 +19,7 @@ export function FundButton({
   target,
   text: buttonText = 'Fund',
 }: FundButtonReact) {
+  const componentTheme = useTheme();
```
cpcramer, line 22: If there is time before the launch or as a fast follow - look to add Theme in an outer layer. Perhaps directly on `cn`

alessey:
swap settings variable not visible in base and minimal themes

src/useTheme.ts
```diff
@@ -0,0 +1,76 @@
...
+      if (newTheme !== currentTheme) {
+        setCurrentTheme(newTheme);
+      }
+    };
+
+    updateTheme(); // Set initial theme
+    mediaQuery.addEventListener('change', updateTheme); // Listen for OS theme changes
+
+    return () => mediaQuery.removeEventListener('change', updateTheme); // Cleanup listener
+  }, [theme, mode, currentTheme]);
```
alessey, line 73: this is the only thing that needs to be watched for changes, you could extract this to its own hook with state and a useeffect which would simplify this hook

alessey: if you set your default/auto defaults in onchainkitprovider, and extract the preferred mode to a hook, i think you should be able to reduce this to 

export function useTheme(): UseThemeReact {
  const preferredMode = usePreferredColorScheme();
  const { appearance } = useOnchainKit();
  const { theme, mode } = appearance || {};  

  if (theme === 'cyberpunk' || theme === 'base' || theme === 'minimal') {
    return theme;
  }

  // If the theme doesn't support light/dark variants, return it as is
  const currentMode = !mode || mode === 'auto' ? preferredMode : mode;
  
  return `${theme}-${currentMode}` as UseThemeReact;
}

playground/nextjs-app-router/components/form/component-mode.tsx
```diff
@@ -0,0 +1,34 @@
...
+        </SelectTrigger>
+        <SelectContent>
+          <SelectItem value="auto">Auto</SelectItem>
+          <SelectItem value="light">Light</SelectItem>
+          <SelectItem value="dark">Dark</SelectItem>
+        </SelectContent>
+      </Select>
+    </div>
+  );
+}
```
abcrane123, line 34: should these files be `ComponentMode.tsx` and `ComponentTheme.tsx` to keep consistent with others or is there a reason you named them differently?

cpcramer: I'm not sure why, but we have been using snake-case in this `form` file. So this format matches that.
Screenshot 2024-10-10 at 2 19 18 PM

src/transaction/components/Transaction.tsx
```diff
@@ -35,7 +38,9 @@ export function Transaction({
       onStatus={onStatus}
       onSuccess={onSuccess}
     >
-      <div className={cn(className, 'flex w-full flex-col gap-2')}>
+      <div
+        className={cn(componentTheme, className, 'flex w-full flex-col gap-2')}
```
abcrane123, line 42: `className` should be last here so it can override other styles

src/useTheme.ts
abcrane123: yeah +1 i see a few ways this could be refactored because it is quite complex

src/transaction/components/Transaction.tsx
cpcramer: Good eye, will fix in the follow up!

alessey (approved):
🥇

0xAlec (approved):

PR #1409, opened 2024-10-14
nft api

alessey:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/nft/components/NFTErrorFallback.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+  return (
+    <div
+      className={cn(
+        componentTheme,
+        color.foreground,
+        background.default,
+        border.defaultActive,
+        border.radius,
+        'flex w-full max-w-[500px] flex-col items-center justify-center border px-6 py-4',
```
cpcramer, line 15: Do we still need the class `border` if we have `border.defaultActive` and `border.raidus`?

src/nft/components/mint/NFTCollectionTitle.tsx
```diff
@@ -12,5 +12,15 @@ export function NFTCollectionTitle({ className }: NFTCollectionTitleReact) {
     return null;
   }
 
-  return <div className={cn('pt-4 pb-1', text.title1, className)}>{name}</div>;
+  return (
+    <div
+      className={cn(
+        'overflow-hidden text-ellipsis pt-4 pb-1',
+        text.title1,
+        className,
```
cpcramer, line 20: Nit: I feel like the order of classNames should always be:

`themes -> hard coded tailwind classes -> className`

src/nft/components/NFTErrorFallback.tsx
alessey: border radius only sets the radius and defaultActive only sets the color

src/nft/components/mint/NFTCreator.tsx
```diff
@@ -18,7 +18,7 @@ export function NFTCreator({ className }: NFTCreatorReact) {
   return (
     <div className={cn('flex justify-between pb-2', className)}>
       <Identity
-        className="space-x-2 px-0"
+        className="px-0 [&>div]:space-x-2"
```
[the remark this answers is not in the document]
alessey: it was in my tests, is there a reason it wouldn't?

src/nft/components/view/NFTLastSoldPrice.tsx
```diff
@@ -12,14 +12,18 @@ export function NFTLastSoldPrice({
...
-  if (!amount || !currency || !amountUSD) {
+  if (
+    !lastSoldPrice?.amount ||
+    !lastSoldPrice?.currency ||
+    !lastSoldPrice?.amountUSD
+  ) {
     return null;
   }
 
+  const { amount, currency, amountUSD } = lastSoldPrice;
```
cpcramer, line 25: I think restructuring should still work here if the values are undefined?

```
const { amount, currency, amountUSD } = useNFTContext();
if (!amount || !currency || !amountUSD) {
  return null;
}
```

alessey: only if i default lastSoldPrice to {}

alessey: `  const {
    lastSoldPrice: { amount, currency, amountUSD } = {},
  } = useNFTContext();
` I always feel those destructure defaults are kind of messy

src/nft/utils/ipfs.ts
```diff
@@ -0,0 +1,31 @@
+export function IsValidIpfsUrl(url: string): boolean {
```
cpcramer, line 1: Should we put these two functions in their own files? 

`/utils/IsValidIpfsUrl/.ts`

`/utils/convertIpfsToHttps.ts`

src/nft/types.ts
```diff
@@ -130,8 +135,8 @@ export type NFTMintCardReact = {
   contractAddress: Hex; // Contract address of the NFT
   tokenId?: string; // Token ID of the NFT only required for ERC1155
   isSponsored?: boolean; // Optional boolean to determine if the mint is sponsored by paymaster
-  useNFTData: UseNFTData; // Required hook to supply NFT data
-  buildMintTransaction: BuildMintTransaction; // Required function that builds the mint transaction
+  useNFTData?: UseNFTData; // Required hook to supply NFT data
+  buildMintTransaction?: BuildMintTransaction; // Required function that builds the mint transaction
```
cpcramer, line 139: A little confused why these params are marked as optional the comments state that they are required

src/nft/components/mint/NFTCreator.tsx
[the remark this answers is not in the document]
cpcramer: Cool, I wasn't able to get this syntax to work some time ago. I thought it was a tailwind thing

src/nft/components/view/NFTLastSoldPrice.tsx
cpcramer: oo yeah I don't like that

src/nft/utils/ipfs.ts
alessey: i can, they were both so related i figured a single ipfs file would work well

src/nft/types.ts
alessey: good catch, updated the params, but not the comments, they used to be required.

alessey: fixed

src/nft/components/mint/NFTCollectionTitle.tsx
alessey: I may have a few places like this, I can revisit this in a separate PR

src/api/types.ts
```diff
@@ -143,3 +144,65 @@ export type RawTransactionData = {
...
+};
+
+type MintTransaction = {
+  call_data: {
+    data: Address;
+    to: Address;
+    from: Address;
+    value: string;
+  };
+};
```
abcrane123, line 206: nit but can we add descriptions for these

src/nft/hooks/useMintDetails.ts
```diff
@@ -0,0 +1,31 @@
...
+        tokenId,
+      });
+
+      if (isNFTError(mintDetails)) {
+        throw mintDetails;
+      }
+
+      return mintDetails;
+    },
+    enabled: true,
```
abcrane123, line 27: i think the default is true here so you might not need this. is there a case where we want to disable ?

src/nft/hooks/useTokenDetails.ts
```diff
@@ -0,0 +1,29 @@
...
+        tokenId,
+      });
+
+      if (isNFTError(tokenDetails)) {
+        throw tokenDetails;
+      }
+
+      return tokenDetails;
+    },
+    enabled: true,
```
abcrane123, line 25: same as prev comment, is there a case where this should be disabled?

src/nft/utils/ipfs.ts
abcrane123: i think it is fine in one file

src/api/types.ts
alessey: updated

src/nft/hooks/useTokenDetails.ts
alessey: nope, updated

src/nft/components/NFTErrorFallback.tsx
```diff
@@ -0,0 +1,23 @@
...
+      )}
+      data-testid="ockNFTMintCardFallback_Container"
+    >
+      <div>Sorry, we had an unhandled error</div>
```
0xAlec, line 19: nit, this copy should be `Sorry, please try again later.` to be consistent with our other 500 responses

src/nft/components/mint/NFTAssetCost.tsx
```diff
@@ -18,19 +18,19 @@ export function NFTAssetCost({ className }: NFTAssetCostReact) {
     return null;
   }
 
-  if (price?.amount === 0) {
+  if (Number(price?.amount) === 0) {
```
0xAlec, line 21: `amount` is in wei right? you'll get an overflow casting this to Number

0xAlec: should use native BigInt here

src/nft/utils/ipfs.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @returns https:// URL for the media
+ * @link https://docs.ipfs.tech/how-to/address-ipfs-on-web/ - for more details.
+ */
+export function convertIpfsToHttps(url?: string): string | undefined {
+  if (!url) {
+    return undefined;
+  }
+
+  const IPFS_GATEWAY_URL =
+    'https://b891d14d436694bb9a7feeba91730b95.ipfscdn.io/ipfs/';
```
0xAlec, line 24: is there rate limiting on this gateway?

src/nft/components/mint/NFTAssetCost.tsx
alessey: no, its in ETH, converted on the BE

src/nft/utils/ipfs.ts
alessey: i'm not sure, this is the one wallet uses

alessey: the only time I've seen that we've had to handle ipfs urls on the FE is for cb pre-mints, shouldn't happen often

alessey: could switch to just ipfs.io

cpcramer (approved):

PR #1414, opened 2024-10-14
chore: TokenSelectButton and Base theme cleanup

cpcramer:
**What changed? Why?**
- Fix token select button background and hover background for visibility.
- Update Base theme colors.
- Add back in .light and .dark internally to map to the respective themes.

Screenshot 2024-10-14 at 3 47 22 PM

**Notes to reviewers**

**How has it been tested?**


src/styles/tailwind-base.css
```diff
@@ -2,6 +2,7 @@
 
 @layer base {
   :root,
+  .light,
```
cpcramer, line 5: @avidreder I think we want to add back in light and dark classes to correspond to the respective theming. This should help with backwards compatibility for those still using root level html classes for light and dark mode.

fakepixels (approved):

0xAlec (approved):

PR #1415, opened 2024-10-15
chore: Add Noto Sans Mono font-family to Hacker theme

cpcramer:
**What changed? Why?**
Add `Noto Sans Mono` fontFamily to the Hacker theme. 

**Notes to reviewers**

**How has it been tested?**
Screenshot 2024-10-14 at 7 22 06 PM

Screenshot 2024-10-14 at 7 22 21 PM

Screenshot 2024-10-14 at 7 22 37 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,6 @@
     "@rainbow-me/rainbowkit": "^2.1.3",
     "@tanstack/react-query": "^5",
     "clsx": "^2.1.1",
-    "geist": "^1.3.1",
     "graphql": "^14 || ^15 || ^16",
     "graphql-request": "^6.1.0",
     "permissionless": "^0.1.29",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -37,6 +37,7 @@
     "@rainbow-me/rainbowkit": "^2.1.3",
     "@tanstack/react-query": "^5",
     "clsx": "^2.1.1",
+    "geist": "^1.3.1",
```
cpcramer, line 40: I don't love adding a new dependency but I think it's the only way to support this theme

fakepixels (approved):

alessey:
can we dynamically load the correct fonts for a theme?

0xAlec (approved):

PR #1419, opened 2024-10-15
feat: add resetOnComplete prop to Transaction

abcrane123:
**What changed? Why?**
- add prop that enables resetting of state after txn completion

**Notes to reviewers**

**How has it been tested?**


alessey:
is the use case here that a user passes in calls or contracts then submits them, and wants to pass in more calls/contracts to the same button?  should it just reset state when calls/contracts changes?

cpcramer (approved):

cpcramer:
Can you add a changeset

alessey (approved):

PR #1424, opened 2024-10-16
chore: Add browser window check in usePreferredColorScheme

cpcramer:
**What changed? Why?**
Add browser window check to avoid server side errors when the browser window does not exist. 

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/.changeset/grumpy-weeks-smile.md
+++ b/.changeset/grumpy-weeks-smile.md
@@ -3,3 +3,4 @@
 ---
 
 **feat**: add formatting for scientific notation for Swap amounts. by @0xAlec #1423
+-**fix**: Added browser window check in usePreferredColorScheme to fix server side errors. By @cpcramer #1424


src/internal/hooks/usePreferredColorScheme.ts
```diff
@@ -1,19 +1,22 @@
...
-    mediaQuery.matches ? 'dark' : 'light',
-  );
+  const [colorScheme, setColorScheme] = useState<'light' | 'dark'>('light');
 
   useEffect(() => {
-    function updateColorScheme(event: MediaQueryListEvent) {
-      setColorScheme(event.matches ? 'dark' : 'light');
-    }
+    // Check if window is defined (client-side)
+    if (typeof window !== 'undefined' && window.matchMedia) {
```
0xAlec, line 8: if you move this into the `useEffect`, do you still need to check the type? doesn't Next.js automatically use client-side rendering for all code inside React hooks?

cpcramer: TIL

0xAlec (approved):

alessey (approved):

PR #1450, opened 2024-10-18
feat: add promises to transactions and calls

alessey:
**What changed? Why?**
* Add ability to pass a promise into the transaction component that resolves to calls or contracts
* Added new building transaction status
* Added new onError state when promise is rejected

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/transaction/components/Transaction.tsx
+++ b/src/transaction/components/Transaction.tsx
@@ -18,11 +18,12 @@ export function Transaction({
 }: TransactionReact) {
   const isMounted = useIsMounted();
   const componentTheme = useTheme();
+  const { chain } = useOnchainKit();
+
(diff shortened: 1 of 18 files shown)


fakepixels (approved):

src/transaction/components/Transaction.tsx
```diff
@@ -18,11 +18,12 @@ export function Transaction({
 }: TransactionReact) {
   const isMounted = useIsMounted();
   const componentTheme = useTheme();
+  const { chain } = useOnchainKit();
```
alessey, line 21: this fixes the hooks order console error

cpcramer (approved):

0xAlec (approved):

PR #1451, opened 2024-10-18
feat: default Wagmi configuration in OnchainKit context

0xAlec:
**What changed? Why?**
- add `AppConfig` type
- add `name` and `logo` fields to `AppConfig`
- add `getDefaultConfig` which returns a Wagmi config with settings provided by `AppConfig`
- add logic for using the parent `WagmiProvider` if provided, otherwise default to the value returned by `getDefaultConfig`

**Notes to reviewers**

**How has it been tested?**


src/getDefaultConfig.ts
```diff
@@ -0,0 +1,41 @@
...
+        appLogoUrl:
+          config?.appearance?.logo ||
+          'https://onchainkit.xyz/favicon/48x48.png?v4-19-24',
+        preference: 'smartWalletOnly',
+      }),
+    ],
+    storage: createStorage({
+      storage: cookieStorage,
+    }),
+    ssr: true,
```
0xAlec, line 29: my understanding is this is broken for ssr frameworks like `Next.js` with the default value 

```
Turning on the ssr property means that content from the external stores will be hydrated on the client after the initial mount.
```

for non-SSR apps, is there any drawback to this approach? cc: @alessey @alissacrane-cb 


[link](https://wagmi.sh/react/guides/ssr)

src/OnchainKitProvider.tsx
```diff
@@ -48,6 +56,32 @@ export function OnchainKitProvider({
...
+      console.error('Error fetching config, using OnchainKit defaults:', error);
+      throw error;
+    }
+  }
+
+  // If WagmiProvider is not found, return the context with defaulted parent providers
+  if (!providedConfig) {
+    return (
+      <WagmiProvider config={getDefaultConfig({ apiKey, config })}>
+        <QueryClientProvider client={new QueryClient()}>
```
0xAlec, line 76: `create wagmi` wraps these in a `useState` but not sure the benefits of doing so

```
  const [config] = useState(() => getConfig())
  const [queryClient] = useState(() => new QueryClient())
```

0xAlec: should the Wagmi configuration (and/or the query client) be memo-ized?

src/OnchainKitProvider.tsx
```diff
@@ -48,6 +56,32 @@ export function OnchainKitProvider({
...
+    if (!(error instanceof WagmiProviderNotFoundError)) {
+      console.error('Error fetching config, using OnchainKit defaults:', error);
+      throw error;
+    }
+  }
+
+  // If WagmiProvider is not found, return the context with defaulted parent providers
+  if (!providedConfig) {
+    return (
+      <WagmiProvider config={getDefaultConfig({ apiKey, config })}>
```
alessey, line 75: could you pass onchainkticonfig.config here or move the defaults to a consts file so they aren't duplicated?

src/OnchainKitProvider.tsx
0xAlec:
```
  const defaultConfig = useMemo(() => getDefaultConfig({ apiKey, config }), [apiKey, config]);
  const queryClient = useMemo(() => new QueryClient(), []);

  if (!providedConfig) {
    return (
      <WagmiProvider config={defaultConfig}>
        <QueryClientProvider client={queryClient}>
          <OnchainKitContext.Provider value={value}>
            {children}
          </OnchainKitContext.Provider>
        </QueryClientProvider>
      </WagmiProvider>
    );
  }
```

if we do this, are these values created even if they aren't used? not sure on the client-side implications, since if a parent `WagmiProvider` is provided these would be essentially duplicates and take up storage/memory

alessey: you could split this default provider out to another component to solve that

src/createWagmiConfig.ts
```diff
@@ -0,0 +1,38 @@
...
+import type { CreateWagmiConfigParams } from './types';
+
+// createWagmiConfig returns a WagmiConfig (https://wagmi.sh/react/api/createConfig) using OnchainKit provided settings.
+// This function should only be used if the user does not provide WagmiProvider as a parent in the React context.
+export const createWagmiConfig = ({
+  apiKey,
+  appName,
+  appLogoUrl,
+}: CreateWagmiConfigParams) => {
+  console.log('Created default Wagmi config');
```
alessey, line 13: intentional?

0xAlec: good catch, ty

src/OnchainKitProvider.tsx
```diff
@@ -48,6 +57,46 @@ export function OnchainKitProvider({
...
+  const defaultQueryClient = useMemo(() => {
+    // IMPORTANT: Don't create a new QueryClient if one already exists
+    // This prevents the user-provided QueryClient from being overriden
+    return providedQueryClient || new QueryClient();
+  }, [providedQueryClient]);
+
+  // If a dependency is missing, return a context with default parent providers
+  if (!providedWagmiConfig || !providedQueryClient) {
+    return (
+      <WagmiProvider config={defaultConfig}>
```
alessey, line 90: doesn't this leave the possibility of having two wagmiProviders or queryClientProviders if one or the other wasn't set up?

0xAlec: no because I think `defaultConfig` would default to the provided one for the respective dependency

0xAlec: but there might be two `WagmiProviders` you're right

0xAlec: unsure if there's any adverse effects there, will do a bit more testing

PR #1428, opened 2024-10-16
feat: add `Checkout` component to Playground

0xAlec:
**What changed? Why?**
- add `Checkout` component to Playground
- add option for `chargeId` or `productId`

**Notes to reviewers**

**How has it been tested?**
locally + playground

PR #1458, opened 2024-10-20
fix: `isSponsored` in `Checkout` component

0xAlec:
**What changed? Why?**
- was not passing in the proper paymaster field for `capabilities` previously

**Notes to reviewers**

**How has it been tested?**
cherry picked to https://github.com/coinbase/onchainkit/pull/1428 and tested locally


fakepixels:
Should this be an option? I think we should default this to true at launch?

PR #1460, opened 2024-10-21
feat: add custom error and success states to txn button

abcrane123:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/transaction/components/TransactionButton.tsx
+++ b/src/transaction/components/TransactionButton.tsx
@@ -60,12 +60,12 @@ export function TransactionButton({
     transactionId,
   });
 
-  const { errorText, successText, pendingText } = useMemo(() => {
+  const { errorText, successText, pendingContent } = useMemo(() => {
     const successText = successOverride?.text ?? 'View transaction';
     const errorText = errorOverride?.text ?? 'Try again';
-    const pendingText = pendingOverride?.text;
+    const pendingContent = pendingOverride?.text ?? <Spinner />;
 
-    return { successText, errorText, pendingText };
+    return { successText, errorText, pendingContent };
   }, [errorOverride, pendingOverride, successOverride]);
 
   const successHandler = useCallback(() => {
@@ -109,7 +109,7 @@ export function TransactionButton({
       return errorText;
     }
     if (displayPendingState) {
-      return pendingText;
+      return pendingContent;
     }
     return idleText;
   }, [
@@ -117,7 +117,7 @@ export function TransactionButton({
     errorMessage,
     errorText,
     idleText,
-    pendingText,
+    pendingContent,
     receipt,
     successText,
   ]);
@@ -148,15 +148,9 @@ export function TransactionButton({
       disabled={isDisabled}
       data-testid="ockTransactionButton_Button"
     >
-      {displayPendingState && !pendingText ? (
-        <Spinner />
-      ) : (
-        <span
-          className={cn(text.headline, color.inverse, 'flex justify-center')}
-        >
-          {buttonContent}
-        </span>
-      )}
+      <div className={cn(text.headline, color.inverse, 'flex justify-center')}>
+        {buttonContent}
+      </div>
     </button>
   );
 }
--- a/src/transaction/components/TransactionButton.tsx
+++ b/src/transaction/components/TransactionButton.tsx
@@ -61,12 +61,8 @@ export function TransactionButton({
   });
 
   const { errorText, successText, pendingText } = useMemo(() => {
-    const successText = successOverride?.text
-      ? successOverride?.text
-      : 'View transaction';
-
-    const errorText = errorOverride?.text ? errorOverride?.text : 'Try again';
-
+    const successText = successOverride?.text ?? 'View transaction';
+    const errorText = errorOverride?.text ?? 'Try again';
     const pendingText = pendingOverride?.text;
 
     return { successText, errorText, pendingText };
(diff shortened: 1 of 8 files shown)


fakepixels (approved):

src/transaction/types.ts
```diff
@@ -75,6 +86,7 @@ export type TransactionButtonReact = {
 
 export type TransactionContextType = {
   chainId?: number; // The chainId for the transaction.
+  customStates?: CustomStates;
```
cpcramer, line 89: Can we add a comment description

cpcramer:
Add a changeset

src/transaction/components/TransactionButton.tsx
```diff
@@ -50,41 +52,69 @@ export function TransactionButton({
...
 
+  const { errorText, successText } = useMemo(() => {
+    const successText = successOverride?.text
+      ? successOverride?.text
+      : 'View transaction';
+
+    const errorText = errorOverride?.text ? errorOverride?.text : 'Try again';
+
+    return { successText, errorText };
+  }, [errorOverride, successOverride]);
```
0xAlec, line 63: I also feel like this might be cleaner if the top-level prop is not necessarily an override but the state itself, then we can set the default behavior using a default value:

e.g.
```
success: { text: 'View Transaction', onClick = () => { return window.open(basescan) } };
```

but more of a nit

0xAlec: then we can eliminate lines 55-87 (around 30~ LOC)

src/transaction/types.ts
```diff
@@ -15,6 +15,11 @@ import type {
 
 export type Call = { to: Hex; data?: Hex; value?: bigint };
 
+type TransactionButtonOverride = {
+  text?: string;
+  onClick?: (receipt?: TransactionReceipt) => void;
+};
```
cpcramer, line 21: Define in `TransactionButton` if this is used for a singular purpose

src/transaction/components/TransactionButton.tsx
abcrane123: the issue is the default success state can be either `showCallsStatus({ id: transactionId });` or `window.open(...)` depending on wallet type

src/transaction/types.ts
abcrane123: it will be exported

src/transaction/components/TransactionButton.tsx
```diff
@@ -30,61 +34,106 @@ export function TransactionButton({
...
+    const successText = successOverride?.text
+      ? successOverride?.text
+      : 'View transaction';
+
+    const errorText = errorOverride?.text ? errorOverride?.text : 'Try again';
+
+    const pendingText = pendingOverride?.text;
+
+    return { successText, errorText, pendingText };
+  }, [errorOverride, pendingOverride, successOverride]);
```
alessey, line 73: you should be able to simplify these to const errorText = errorOverride?.text ?? 'Try again', you could also set pendingText = pendingOverride?.text ?? <Spinner/> which would probably simplify render some.

src/transaction/types.ts
```diff
@@ -15,6 +15,11 @@ import type {
 
 export type Call = { to: Hex; data?: Hex; value?: bigint };
 
+type TransactionButtonOverride = {
+  text?: string;
```
alessey, line 19: we are stuck with the text prop name, but any reason not to allow a ReactNode here?  this would allow someone to use their own custom loading icon for the pending state, etc.

src/transaction/types.ts
```diff
@@ -71,6 +75,9 @@ export type TransactionButtonReact = {
   className?: string; // An optional CSS class name for styling the button component.
   disabled?: boolean; // A optional prop to disable the submit button
   text?: string; // An optional text to be displayed in the button component.
+  errorOverride?: TransactionButtonOverride; // Optional overrides for text and onClick handler in error state (default is resubmit txn)
+  successOverride?: TransactionButtonOverride; // Optional overrides for text and onClick handler in success state (default is view txn on block explorer)
+  pendingOverride?: { text: string }; // Optional overrides for text in pending state (default is loading spinner)
```
alessey, line 80: Pick<TransactionButtonOverride, 'text'> would keep these types more consistent

src/transaction/types.ts
abcrane123: yeah i agree i'll update type but keep `text` name for consistency since i don't think it is worth causing a breaking change

src/transaction/components/TransactionButton.tsx
abcrane123: not sure what you mean about the `pendingText` line but i updated the others

alessey: if you had const pendingText = pendingOverride?.text ?? <Spinner/> i think you'd be able to remove to. remove [this logic](https://github.com/coinbase/onchainkit/pull/1460/files#diff-bb9ce23f9b647b4d1b981336417d591dcf4aae8ac2b3d60757fe8f80f2ad43a1R151-R158)  and just display buttonContent always

abcrane123: ohh i see yeah i didn't do that because buttonContent is rendered inside this `span` element so i'd have to have `buttonContent` return `successText`, `errorText` and `idleText` within a `span`. do you think that is cleaner?
```
      {displayPendingState && !pendingText ? (
        <Spinner />
      ) : (
        <span
          className={cn(text.headline, color.inverse, 'flex justify-center')}
        >
          {buttonContent}
        </span>
      )}
```

PR #1467, opened 2024-10-21
feat: add prop that exposes callback to return txn state

abcrane123:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/transaction/components/Transaction.tsx
+++ b/src/transaction/components/Transaction.tsx
@@ -13,6 +13,7 @@ export function Transaction({
   children,
   contracts,
   onError,
+  onResetState,
   onStatus,
   onSuccess,
 }: TransactionReact) {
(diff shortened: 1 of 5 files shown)


cpcramer (approved):

alessey:
should the transactions component reset state if calls or contracts changes?  Along with the resetOnComplete prop, this seems like it would handle both the use cases of transact -> new contract -> transact and transacting the same contract repeatedly.

0xAlec:
> should the transactions component reset state if calls or contracts changes? Along with the resetOnComplete prop, this seems like it would handle both the use cases of transact -> new contract -> transact and transacting the same contract repeatedly.

+1 on this, I don't think this is the proper developer experience here. we should just refresh the component state to `init` if one of the provided props changes and (optionally) if a boolean configuration like `resetOnComplete` is specified

PR #1476, opened 2024-10-22
feat: nft part 1

alessey:
**What changed? Why?**
initial nft components, view/mint parent components and providers

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/nft/components/NftMint.tsx
+++ b/src/nft/components/NftMint.tsx
@@ -1,23 +1,23 @@
 import { background, border, cn, color } from '../../styles/theme';
 import { useIsMounted } from '../../useIsMounted';
 import { useTheme } from '../../useTheme';
-import { LifecycleType, type NftMintReact } from '../types';
-import { NftLifecycleProvider } from './NftLifecycleProvider';
-import { NftMintProvider } from './NftMintProvider';
-import { NftProvider } from './NftProvider';
+import { LifecycleType, type NFTMintReact } from '../types';
+import { NFTLifecycleProvider } from './NFTLifecycleProvider';
+import { NFTMintProvider } from './NFTMintProvider';
+import { NFTProvider } from './NFTProvider';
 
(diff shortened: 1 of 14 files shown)


src/nft/components/NftMint.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { LifecycleType, type NftMintReact } from '../types';
+import { NftLifecycleProvider } from './NftLifecycleProvider';
+import { NftMintProvider } from './NftMintProvider';
+import { NftProvider } from './NftProvider';
+
+export function NftMint({
+  children,
+  className,
+  contractAddress,
+  tokenId = '1',
```
0xAlec, line 13: are NFT `tokenIds` 0-indexed?

0xAlec: seems like this is also assuming the NFT is going to be the first mint - is it a common developer workflow to provide the specific `tokenId` you're going to mint? I would assume that's handled by the intrinsic smart contract logic

src/nft/types.ts
```diff
@@ -0,0 +1,230 @@
...
+  callData?: {
+    data: Hex;
+    to: Hex;
+    value: bigint;
+  }[];
+};
+
+type BuildMintTransaction = (
+  props: BuildMintTransactionDataProps,
+) => Promise<Call[]>;
```
0xAlec, line 107: would be nice if this could be `Promise<Call[]> | Promise<ContractFunctionParams[]>`

src/nft/components/NftMint.tsx
alessey: some are 0 indexed, not all, depends on the contract.  I've removed this default and will handle based on contract type in the data/mint hooks.

src/nft/types.ts
alessey: will revisit this in a later PR

PR #1479, opened 2024-10-22
feat: add nft scaffolding

alessey:
**What changed? Why?**
Adding NFT scaffolding/providers

**Notes to reviewers**

**How has it been tested?**


src/nft/types.ts
```diff
@@ -0,0 +1,230 @@
...
+
+/**
+ * Note: exported as public Type
+ */
+export type NFTData = {
+  name?: string;
+  description?: string;
+  imageUrl?: string;
+  animationUrl?: string;
+  mimeType?: string;
```
cpcramer, line 72: nit: Can we add a comment for this param?

alessey: added.  it may change as I do more testing, but that's the current implementation.

src/nft/components/NFTMint.tsx
```diff
@@ -0,0 +1,64 @@
...
+  onSuccess,
+}: NFTMintReact) {
+  const componentTheme = useTheme();
+
+  const isMounted = useIsMounted();
+
+  // prevents SSR hydration issue
+  if (!isMounted) {
+    return null;
+  }
```
cpcramer, line 28: I'm wondering if we should access the theme after the component is mounted? 

```suggestion
  const isMounted = useIsMounted();

  // prevents SSR hydration issue
  if (!isMounted) {
    return null;
  }
  
    const componentTheme = useTheme();

```

cpcramer (approved):

src/nft/components/NFTMint.tsx
alessey: can't because of hooks

src/nft/components/NFTLifecycleProvider.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const context = useContext(NFTLifecycleContext);
+  if (context === emptyContext) {
+    throw new Error(
+      'useNFTLifecycleContext must be used within an NFTView or NFTMint component',
+    );
+  }
+  return context;
+}
+
+export function NFTLifecycleProvider({
```
abcrane123, line 24: i like how you abstracted this away. i think we should follow the same pattern in our other components.

src/nft/components/NFTMint.tsx
```diff
@@ -0,0 +1,64 @@
...
+  useNFTData,
+  useNFTMintData,
+  buildMintTransaction,
+  onStatus,
+  onError,
+  onSuccess,
+}: NFTMintReact) {
+  const componentTheme = useTheme();
+
+  const isMounted = useIsMounted();
```
abcrane123, line 23: do we need this in both NFTMint and NFTView component ? or could we move to wrapper ?

alessey: there's not really any wrapper for these, they are just separate top level components.

alessey: yes, i think we can, let me work something up

alessey: I consolidated the styling and these two checks to a single NFT component used in both cards.  I also removed the NFTMintProvider allowing one data hook to be passed and defined which data applies to each component in types.  The NFTMint card will be required when using the MintButton as it sets lifecycle status separately and allows a user to pass in the buildTransaction function.

0xAlec (approved):

src/nft/components/NFTLifecycleProvider.tsx
0xAlec: +1, this is great

PR #1480, opened 2024-10-22
feat: add heterogeneous calls functionality

abcrane123:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/playground/nextjs-app-router/components/AppProvider.tsx
+++ b/playground/nextjs-app-router/components/AppProvider.tsx
@@ -21,6 +21,7 @@ export enum OnchainKitComponent {
 export enum TransactionTypes {
   Calls = 'calls',
   Contracts = 'contracts',
+  ContractsAndCalls = 'contractsAndCalls',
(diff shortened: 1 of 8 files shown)


jxom:
FWIW, Viem & Wagmi already supports this. See snippet under [Contract Calls](https://viem.sh/experimental/eip5792/sendCalls#contract-calls) for `sendCalls` that demonstrates a **value transfer** call along with a **contract** call.

abcrane123:
> FWIW, Viem & Wagmi already supports this. See snippet under [Contract Calls](https://viem.sh/experimental/eip5792/sendCalls#contract-calls) for `sendCalls` (demonstrates a **value transfer** call along with 2 **contract** calls).
> 
> Separating `contracts` and `calls` may break the sequential call behavior of EIP-5792 (ie. a consumer would not be able to batch a transfer call → contract call → transfer call).

oh amazing thank you for the tip! upgrading our packages now.

jxom:
Nice one!

PR #1483, opened 2024-10-23
feat: add view components

alessey:
**What changed? Why?**
Add nft view components

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/internal/svg/baseSvg.tsx
+++ b/src/internal/svg/baseSvg.tsx
@@ -1,0 +1,18 @@
+export const baseSvg = (
+  <svg
+    data-testid="ock-baseSvg"
+    role="img"
+    aria-label="base"
+    width="100%"
+    height="100%"
+    viewBox="0 0 146 146"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <circle cx="73" cy="73" r="73" fill="#0052FF" />
+    <path
+      d="M73.323 123.729C101.617 123.729 124.553 100.832 124.553 72.5875C124.553 44.343 101.617 21.4463 73.323 21.4463C46.4795 21.4463 24.4581 42.0558 22.271 68.2887H89.9859V76.8864H22.271C24.4581 103.119 46.4795 123.729 73.323 123.729Z"
+      fill="white"
+    />
+  </svg>
+);
--- a/src/nft/components/view/NFTImage.tsx
+++ b/src/nft/components/view/NFTImage.tsx
@@ -1,0 +1,91 @@
+import { useCallback, useEffect, useState } from 'react';
+import { defaultNFTSvg } from '../../../internal/svg/defaultNFTSvg';
+import { cn } from '../../../styles/theme';
+import type { NFTError } from '../../types';
+import { useNFTContext } from '../NFTProvider';
+
+type NFTImageReact = {
+  className?: string;
+  onLoading?: (mediaUrl: string) => void;
+  onLoaded?: () => void;
+  onError?: (error: NFTError) => void;
+};
+
+export function NFTImage({
+  className,
+  onLoading,
+  onLoaded,
+  onError,
+}: NFTImageReact) {
+  const { imageUrl, description } = useNFTContext();
+  const [loaded, setLoaded] = useState(false);
+  const [error, setError] = useState(false);
+  const [transitionEnded, setTransitionEnded] = useState(false);
+
+  const loadImage = useCallback(() => {
+    if (imageUrl) {
+      onLoading?.(imageUrl);
+
+      const img = new Image();
+      img.onload = () => {
+        setLoaded(true);
+        onLoaded?.();
+      };
+
+      img.onerror = (error: string | Event) => {
+        onError?.({
+          error: typeof error === 'string' ? error : error.type,
+          code: 'NmNIc01', // NFT module NFTImage component 01 error
+          message: 'Error loading image',
+        });
+        setError(true);
+      };
+      img.src = imageUrl;
+    }
+  }, [imageUrl, onLoading, onLoaded, onError]);
+
+  useEffect(() => {
+    loadImage();
+  }, [loadImage]);
+
+  const handleRetry = useCallback(async () => {
+    setError(false);
+    loadImage();
+  }, [loadImage]);
+
+  const handleTransitionEnd = () => {
+    setTransitionEnded(true);
+  };
+
+  return (
+    <div
+      className={cn(
+        'relative flex h-[450px] max-h-screen items-center justify-center',
+        className,
+      )}
+    >
+      {error && (
+        <div className="absolute top-[60%] z-10">
+          <button type="button" onClick={handleRetry}>
+            retry
+          </button>
+        </div>
+      )}
+      {!transitionEnded && (
+        <div
+          className={`absolute inset-0 ${loaded ? 'opacity-0' : 'opacity-100'} transition-[opacity] duration-500 ease-in-out`}
+        >
+          {defaultNFTSvg}
+        </div>
+      )}
+      <img
+        data-testid="ockNFTImage"
+        src={imageUrl}
+        alt={description}
+        decoding="async"
+        className={`max-h-[450px] transition-[opacity] duration-500 ease-in-out ${loaded ? 'opacity-100' : 'opacity-0'}`}
+        onTransitionEnd={handleTransitionEnd}
+      />
+    </div>
+  );
+}
--- a/src/nft/components/view/NFTImage.tsx
+++ b/src/nft/components/view/NFTImage.tsx
@@ -53,9 +53,9 @@ export function NFTImage({
     loadImage();
   }, [loadImage]);
 
-  const handleTransitionEnd = () => {
+  const handleTransitionEnd = useCallback(() => {
     setTransitionEnded(true);
-  };
+  }, []);
 
   return (
     <div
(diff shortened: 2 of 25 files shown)


src/nft/components/view/NFTImage.tsx
```diff
@@ -0,0 +1,91 @@
...
+  onLoading?: (mediaUrl: string) => void;
+  onLoaded?: () => void;
+  onError?: (error: NFTError) => void;
+};
+
+export function NFTImage({
+  className,
+  onLoading,
+  onLoaded,
+  onError,
```
0xAlec, line 18: should these just be `onStatus`?

src/nft/components/view/NFTLastSoldPrice.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { useNFTContext } from '../NFTProvider';
+
+type NFTLastSoldPriceProps = {
+  className?: string;
+  label?: string;
+};
+
+export function NFTLastSoldPrice({
+  className,
+  label = 'Mint price',
```
0xAlec, line 12: should we standardize on `label`? we use `text` in our other components

src/nft/components/view/NFTMedia.tsx
```diff
@@ -0,0 +1,80 @@
...
+    );
+  }
+
+  return (
+    <NFTImage
+      onLoading={handleLoading}
+      onLoaded={handleLoaded}
+      onError={handleError}
+    />
+  );
```
0xAlec, line 79: prefer a `switch` here

src/nft/components/view/NFTMedia.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  const mediaType = useMemo(() => {
+    return {
+      isVideo: mimeType?.startsWith('video'),
+      isAudio:
+        mimeType?.startsWith('audio') || mimeType?.startsWith('application'),
+      isImage: mimeType?.startsWith('image'),
+    };
+  }, [mimeType]);
```
0xAlec, line 50: prefer returning an enum here rather than an object with always 2/3 null values

src/nft/components/view/NFTNetwork.tsx
```diff
@@ -0,0 +1,37 @@
...
+  className?: string;
+  label?: string;
+};
+
+const networkMap = {
+  Base: baseSvg,
+} as Record<string, ReactNode>;
+
+export function NFTNetwork({ className, label = 'Network' }: NFTNetworkProps) {
+  const { chain } = useOnchainKit();
```
0xAlec, line 16: I think we should deprecate this top-level `chain` prop, so I don't recommend using it here. 

Is there any other way to tell which chain the NFT is on?

src/nft/components/view/NFTLastSoldPrice.tsx
alessey: label seems more flexible if you ever want to change from string to ReactNode

src/nft/components/view/NFTMedia.tsx
alessey: any particular reason?  performance differences at this point are negligible

src/nft/components/view/NFTImage.tsx
alessey: I think I'd need to basically create a pseudo lifecycle status to change to onStatus.  I could handle lifecycle status in the children, but I was trying to keep the mediaLoaded/success logic centralized in one place.

src/nft/components/view/NFTMedia.tsx
0xAlec: not for performance but for > 2 cases it's cleaner to just use a `switch` compared to multiple `if` statements

0xAlec: definitely a nit though

alessey: updated, it made more sense when i switched to an enum anyways

src/internal/svg/baseSvg.tsx
```diff
@@ -0,0 +1,18 @@
...
+    width="100%"
+    height="100%"
+    viewBox="0 0 146 146"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <circle cx="73" cy="73" r="73" fill="#0052FF" />
+    <path
+      d="M73.323 123.729C101.617 123.729 124.553 100.832 124.553 72.5875C124.553 44.343 101.617 21.4463 73.323 21.4463C46.4795 21.4463 24.4581 42.0558 22.271 68.2887H89.9859V76.8864H22.271C24.4581 103.119 46.4795 123.729 73.323 123.729Z"
+      fill="white"
```
cpcramer, line 15: Is the `fill` always meant to be white here? Is the intention for it to NOT update based off of the current theme?

src/nft/components/view/NFTMintDate.tsx
```diff
@@ -0,0 +1,46 @@
...
+      className={cn(
+        'flex items-center justify-between py-1',
+        text.label2,
+        className,
+      )}
+    >
+      <div>{label}</div>
+      <div>{formattedDate}</div>
+    </div>
+  );
```
cpcramer, line 45: I think these divs are unnecessary and can be removed:

```suggestion
  return (
    <div
      className={cn(
        'flex items-center justify-between py-1',
        text.label2,
        className,
      )}
    >
      {label}
      {formattedDate}
    </div>
  );
```

src/nft/components/view/NFTNetwork.tsx
```diff
@@ -0,0 +1,37 @@
...
+        className,
+      )}
+    >
+      <div>{label}</div>
+      <div className="flex items-center gap-2">
+        <div className="h-4 w-4 object-cover">{networkMap[chain.name]}</div>
+        <div>{chain.name}</div>
+      </div>
+    </div>
+  );
```
cpcramer, line 36: Probably can remove these divs that don't have styling:

```suggestion
return (
  <div
    className={cn(
      'flex items-center justify-between py-1',
      text.label2,
      className,
    )}
  >
    {label}
    <div className="flex items-center gap-2">
      <span className="h-4 w-4 object-cover">{networkMap[chain.name]}</span>
      {chain.name}
    </div>
  </div>
);
```

src/nft/components/view/NFTOwner.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <div
+      className={cn(
+        'flex items-center justify-between',
+        text.label2,
+        className,
+      )}
+    >
+      <div>{label}</div>
```
cpcramer, line 28:
```suggestion
   {label}
```

src/nft/components/view/NFTVideo.tsx
```diff
@@ -0,0 +1,64 @@
...
+      };
+
+      videoRef.current.onloadeddata = () => {
+        onLoaded?.();
+      };
+
+      videoRef.current.onerror = (error: string | Event) => {
+        onError?.({
+          error: typeof error === 'string' ? error : error.type,
+          code: 'NmNVc01', // NFT module NFTVideo component 01 error
```
cpcramer, line 36: Should we add all of the possible error codes in a error / const file and reference from it?

0xAlec: I like this idea

src/internal/svg/baseSvg.tsx
alessey: no, I didn't think the base logo should change

src/nft/components/view/NFTVideo.tsx
alessey: let me revisit that at the end

src/nft/components/view/NFTNetwork.tsx
alessey: this is possible, but I think its easier to see whats going on using an element as a flex-child vs. a text node.

cpcramer (approved):

src/nft/components/view/NFTImage.tsx
```diff
@@ -0,0 +1,91 @@
...
+  }, [loadImage]);
+
+  const handleRetry = useCallback(async () => {
+    setError(false);
+    loadImage();
+  }, [loadImage]);
+
+  const handleTransitionEnd = () => {
+    setTransitionEnded(true);
+  };
```
abcrane123, line 58: nit but should this be useCallback just to keep consistent

src/nft/components/view/NFTImage.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { cn } from '../../../styles/theme';
+import type { NFTError } from '../../types';
+import { useNFTContext } from '../NFTProvider';
+
+type NFTImageReact = {
+  className?: string;
+  onLoading?: (mediaUrl: string) => void;
+  onLoaded?: () => void;
+  onError?: (error: NFTError) => void;
+};
```
abcrane123, line 12: nit move to types file

src/nft/components/view/NFTLastSoldPrice.tsx
```diff
@@ -0,0 +1,41 @@
+import type { ReactNode } from 'react';
+import { cn, text } from '../../../styles/theme';
+import { formatAmount } from '../../../token/utils/formatAmount';
+import { useNFTContext } from '../NFTProvider';
+
+type NFTLastSoldPriceProps = {
+  className?: string;
+  label?: ReactNode;
+};
```
abcrane123, line 9: nit move to types

src/nft/components/view/NFTMintDate.tsx
```diff
@@ -0,0 +1,46 @@
...
+const DATE_OPTIONS = {
+  year: 'numeric',
+  month: 'short',
+  day: 'numeric',
+} as Intl.DateTimeFormatOptions;
+
+type NFTMintDateProps = {
+  className?: string;
+  label?: ReactNode;
+};
```
abcrane123, line 14: nit move to types

abcrane123 (approved):

PR #1489, opened 2024-10-23
feat: IdentityCard and Socials components

cpcramer:
**What changed? Why?**

Add `IdentityCard` and `Socials` components.

- Add copy on click functionality for the `Address` component.
- Playground support - we now show Basenames ENS and regular ENS views.

`IdentityCard`:
Screenshot 2024-10-28 at 1 10 16 PM
Screenshot 2024-10-28 at 1 10 36 PM

`Identity`:
Screenshot 2024-10-28 at 1 11 09 PM

`WalletDropdown`:
Screenshot 2024-10-28 at 1 11 31 PM

**Notes to reviewers**

**How has it been tested?**


src/identity/components/Avatar.tsx
```diff
@@ -70,7 +70,7 @@ export function Avatar({
     <div className="relative">
       <div
         data-testid="ockAvatar_ImageContainer"
-        className={cn('h-8 w-8 overflow-hidden rounded-full', className)}
+        className={cn('h-10 w-10 overflow-hidden rounded-full', className)}
```
alessey, line 73: is this a global change we want to make?

src/identity/components/Socials.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const accountAddress = address ?? contextAddress;
+  const accountChain = chain ?? contextChain;
+
+  if (!accountAddress) {
+    console.error(
+      'Socials: an Ethereum address must be provided to the Socials component.',
+    );
+    return null;
+  }
```
alessey, line 29: is there history behind why Identity doesn't default to the connected address?

src/identity/components/Socials.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+    expect(consoleSpy).toHaveBeenCalledWith(
+      'Socials: an Ethereum address must be provided to the Socials component.',
+    );
+    expect(container.firstChild).toBeNull();
+
+    consoleSpy.mockRestore();
+  });
+
+  it('renders loading state', () => {
```
alessey, line 54: it('should...

src/identity/components/Avatar.tsx
cpcramer: In `Identity` yes!

src/identity/components/Socials.tsx
cpcramer: Hmm I'm not sure what prompted the decision - something Leo implemented

src/identity/components/Address.tsx
```diff
@@ -19,7 +19,10 @@ export function Address({
   const accountAddress = address ?? contextAddress;
 
   return (
-    <span data-testid="ockAddress" className={cn(text.label2, className)}>
+    <span
+      data-testid="ockAddress"
+      className={cn(color.foreground, text.label2, className)}
```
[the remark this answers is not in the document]
cpcramer: Yes, and it was not behaving as expected when switching Themes!

playground/nextjs-app-router/components/demo/Wallet.tsx
```diff
@@ -28,11 +29,12 @@ function WalletComponent() {
           <Name />
         </ConnectWallet>
         <WalletDropdown>
-          <Identity className="px-4 pt-3 pb-2" hasCopyAddressOnClick={true}>
```
alessey, line 31: ?

src/identity/components/Address.tsx
```diff
@@ -18,9 +20,50 @@ export function Address({
...
+        className,
+        'group relative cursor-pointer',
+      )}
+      onClick={handleClick}
+      onKeyDown={(e) => {
+        if (e.key === 'Enter' || e.key === ' ') {
+          e.preventDefault();
+          handleClick();
+        }
+      }}
```
alessey, line 50: extract to handleKeyDown

src/identity/components/IdentityCard.tsx
```diff
@@ -0,0 +1,45 @@
...
+}: IdentityCardReact) {
+  return (
+    <Identity
+      address={address}
+      chain={chain}
+      className={cn(
+        border.radius,
+        background.default,
+        line.default,
+        'items-left flex min-w-[300px] p-4',
```
alessey, line 32: why does this need the min-width?

playground/nextjs-app-router/components/demo/Wallet.tsx
cpcramer: Updated the Identity component to have the Address be clickable rather than the entire Identity component. With the addition of the icons, there are multiple elements that can be clicked on. In a follow up PR I will deprecate this field entirely

src/identity/components/Socials.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const { data: socials, isLoading: isLoadingSocials } = useSocials(
+    {
+      ensName: name ?? '',
+      chain: accountChain,
+    },
+    { enabled: !!name },
+  );
+
+  if (isLoadingName || isLoadingSocials) {
+    return <span className={className} />;
```
alessey, line 45: should this have some sort of loading state?

src/identity/components/IdentityCard.tsx
cpcramer: Design decision from Mind!

src/identity/components/Socials.tsx
```diff
@@ -0,0 +1,113 @@
...
+              href = `https://warpcast.com/${value}`;
+              icon = warpcastSvg;
+              break;
+            }
+            case 'website': {
+              href = value;
+              icon = websiteSvg;
+              break;
+            }
+          }
```
alessey, line 83: if you split this out into a function that returned {href, icon} you could eliminate your object.values loop above and just check to see if your function returned null

src/identity/components/Socials.tsx
cpcramer: I don't think so. Because there is a possibility that the user will not have any socials associated with the address, and in that case we will not render the Socials component. The current setup is cleaner

src/identity/components/Socials.tsx
```diff
@@ -0,0 +1,113 @@
...
+            <a
+              key={platform}
+              href={href}
+              target="_blank"
+              rel="noopener noreferrer"
+              className={cn(
+                pressable.default,
+                border.radius,
+                border.default,
+                'flex items-center justify-center p-2 transition-opacity',
```
alessey, line 95: where is the opacity you're transitioning?

src/identity/utils/getSocials.ts
```diff
@@ -0,0 +1,49 @@
...
+    });
+    return result || null;
+  };
+
+  const [twitter, github, farcaster, website] = await Promise.all([
+    fetchTextRecord('com.twitter'),
+    fetchTextRecord('com.github'),
+    fetchTextRecord('xyz.farcaster'),
+    fetchTextRecord('url'),
+  ]);
```
alessey, line 46: do you want this to reject on any of these failing?  otherwise you should add a .catch to each one so they don't blow up your promise.all

src/identity/components/Socials.tsx
cpcramer: good call this was old and can be removed

src/identity/utils/getSocials.ts
cpcramer: I think if any of these fail then we can just ignore it and return null

src/identity/components/Socials.tsx
cpcramer: Done

playground/nextjs-app-router/components/demo/Wallet.tsx
alessey: you may just be able to stopPropagation on your link clicks and keep this functionality intact

alessey (approved):

PR #1490, opened 2024-10-23
feat: add mint components

alessey:
**What changed? Why?**
Adding Mint components

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/nft/components/mint/NFTAssetCost.tsx
+++ b/src/nft/components/mint/NFTAssetCost.tsx
@@ -3,11 +3,11 @@ import { cn, text } from '../../../styles/theme';
 import { formatAmount } from '../../../token/utils/formatAmount';
 import { useNFTContext } from '../NFTProvider';
 
-type NFTCostProps = {
+type NFTAssetCostReact = {
   className?: string;
 };
 
-export function NFTAssetCost({ className }: NFTCostProps) {
+export function NFTAssetCost({ className }: NFTAssetCostReact) {
   const { price, quantity } = useNFTContext();
 
   if (
--- a/src/nft/components/mint/NFTAssetCost.tsx
+++ b/src/nft/components/mint/NFTAssetCost.tsx
@@ -1,0 +1,40 @@
+import { multiplyFloats } from '../../../internal/utils/multiplyFloats';
+import { cn, text } from '../../../styles/theme';
+import { formatAmount } from '../../../token/utils/formatAmount';
+import { useNFTContext } from '../NFTProvider';
+
+type NFTCostProps = {
+  className?: string;
+};
+
+export function NFTAssetCost({ className }: NFTCostProps) {
+  const { price, quantity } = useNFTContext();
+
+  if (
+    price?.amount === undefined ||
+    !price.currency ||
+    price.amountUSD === undefined
+  ) {
+    return null;
+  }
+
+  if (price?.amount === 0) {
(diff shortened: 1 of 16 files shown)


src/nft/components/mint/NFTCreator.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <Identity
+        className="space-x-2 px-0"
+        address={creatorAddress}
+        schemaId={schemaId}
+      >
+        <Avatar className="h-5 w-5" />
+        <Name>
+          <Badge />
+        </Name>
+      </Identity>
```
0xAlec, line 29: love the composability here

src/nft/components/mint/NFTMintButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+    return null;
+  }
+
+  if (!address) {
+    return (
+      <div className={cn('py-2', className)}>
+        <ConnectWallet className="w-full" />
+      </div>
+    );
+  }
```
0xAlec, line 72: we should align our components to **_always_** render, and only prompt for wallet connection in response to an action (e.g. minting the NFT)

abcrane123: agreed you can see my implementation with SwapButton here https://github.com/coinbase/onchainkit/blob/main/src/swap/components/SwapButton.tsx

alessey: I'm missing the difference

abcrane123: this is the same implementation as our swap button https://github.com/coinbase/onchainkit/blob/main/src/swap/components/SwapButton.tsx

src/nft/components/mint/NFTAssetCost.tsx
```diff
@@ -0,0 +1,40 @@
+import { multiplyFloats } from '../../../internal/utils/multiplyFloats';
+import { cn, text } from '../../../styles/theme';
+import { formatAmount } from '../../../token/utils/formatAmount';
+import { useNFTContext } from '../NFTProvider';
+
+type NFTAssetCostReact = {
+  className?: string;
+};
```
cpcramer, line 8: Should these types be in a general `types.ts` file?

src/nft/components/mint/NFTMintButton.tsx
abcrane123: are you suggesting we change the `onClick` handler to be connect wallet if address isn't connected ?

src/nft/components/mint/NFTTotalCost.tsx
```diff
@@ -0,0 +1,125 @@
...
+            })}
+          </div>
+        </div>
+      </div>
+    );
+  }, [mintFee, price, quantity]);
+
+  if (!price?.amount || !price?.currency || !price?.amountUSD) {
+    return null;
+  }
```
abcrane123, line 88: is there ever a case where an NFT can be free ?

alessey: yea, the NFTAssetCost component handles that use case, for this total cost component I wasn't showing as we already handle Free elsewhere, but can discuss and revisit.

0xAlec (approved):

cpcramer (approved):

PR #1499, opened 2024-10-24
feat: add isSponsored prop to Transaction and playground

abcrane123:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/transaction/components/TransactionProvider.tsx
```diff
@@ -157,6 +163,18 @@ export function TransactionProvider({
...
+    if (isSponsored && paymaster) {
+      return {
+        paymasterService: { url: paymaster },
+        // this needs to be below so devs can override default paymaster
+        // with their personal paymaster in production playgroundd
+        ...transactionCapabilities,
+      };
+    }
+    return transactionCapabilities;
+  }, [isSponsored, paymaster, transactionCapabilities]);
```
0xAlec, line 176: `paymaster` from `useOnchainKit` should be the CDP Paymaster by default, and a 3rd party if override is provided

we should just adjust the playground to use the `OnchainKitProvider` prop instead

0xAlec: also does the spread operator override field values?

0xAlec: ideally to prevent breaking changes we should override the global config paymaster with the `capabilities` prop (if passed in)

src/transaction/components/Transaction.tsx
```diff
@@ -12,6 +12,7 @@ export function Transaction({
   className,
   children,
   contracts,
+  isSponsored,
```
0xAlec, line 15: can you add a `deprecated` docstring to the `capabilities` prop

playground/nextjs-app-router/components/DemoOptions.tsx
```diff
@@ -49,6 +50,7 @@ export default function DemoOptions({
           {commonOptions}
           <Chain />
           <PaymasterUrl />
+          <IsSponsored />
```
0xAlec, line 53: can you add this to other relevant components (Swap and Pay)

0xAlec (changes requested):

src/transaction/components/TransactionProvider.tsx
```diff
@@ -52,17 +53,22 @@ export function useTransactionContext() {
...
   onStatus,
   onSuccess,
 }: TransactionProviderReact) {
   // Core Hooks
   const account = useAccount();
   const config = useConfig();
+  const {
+    config: { paymaster } = { paymaster: undefined },
+  } = useOnchainKit();
+
```
alessey, line 71: with this fix, I should remove https://github.com/coinbase/onchainkit/blob/main/src/nft/components/mint/NFTMintButton.tsx#L37-L39 right?  @0xAlec

playground/nextjs-app-router/components/DemoOptions.tsx
abcrane123: should Pay have `isSponsored` prop ? isn't it always sponsored by Coinbase? or is this a short term thing ?

src/transaction/components/TransactionProvider.tsx
abcrane123: you'll need to add an `isSponsored` prop to the component and pass it to `Transaction`

src/transaction/components/TransactionProvider.tsx
abcrane123: transactionCapabilities will override if provided

playground/nextjs-app-router/components/AppProvider.tsx
```diff
@@ -119,6 +122,8 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
     '0x1D6b183bD47F914F9f1d3208EDCF8BefD7F84E63:1',
   );
 
+  const [isSponsored, setIsSponsoredState] = useState<boolean>(false);
```
cpcramer, line 125: I think we can omit State like so:

```suggestion
  const [isSponsored, setIsSponsored] = useState<boolean>(false);
```

alessey (approved):

cpcramer (approved):

0xAlec (approved):

PR #1501, opened 2024-10-24
fix: nft and general fixes

alessey:
**What changed? Why?**
* Removed Base NFT, split base styles into Mint/View cards as they have different styles
* Renamed NFTView->NFTCard, NFTMint->NFTMintCard
* small type name fixes
* added NFTCollectionTitle
* View links to opensea
* added paymaster
* fix active styles
* 
**Notes to reviewers**

**How has it been tested?**


cpcramer (approved):

src/nft/components/NFTCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const handleOnClick = useCallback(() => {
+    const network = chain?.name.toLowerCase() ?? 'base';
+    const openSeaUrl = `https://opensea.io/assets/${network}/${contractAddress}/${tokenId}`;
+    window.open(openSeaUrl, '_blank', 'noopener,noreferrer');
+  }, [chain, contractAddress, tokenId]);
+
+  // prevents SSR hydration issue
+  if (!isMounted) {
+    return null;
+  }
```
abcrane123, line 35: sorry if i missed this in the other PR but did we decide that we needed this `isMounted` return in both `NFTCard` and `NFTMintCard` ?

src/nft/components/NFTCard.tsx
```diff
@@ -0,0 +1,68 @@
...
+  onSuccess,
+}: NFTCardReact) {
+  const componentTheme = useTheme();
+
+  const isMounted = useIsMounted();
+
+  const { chain } = useAccount();
+
+  const handleOnClick = useCallback(() => {
+    const network = chain?.name.toLowerCase() ?? 'base';
```
abcrane123, line 27: is there somewhere else we can find what `chain` they've configured ? it seems like they would've had to have configured it somewhere right?

src/nft/components/NFTCard.tsx
alessey: Originally I had the same styling in both so I could extract the divs and these hooks to a common component, but, that was incorrect, both these cards are pretty different so I need to handle them separately.

src/nft/components/NFTCard.tsx
abcrane123: ahh i see makes sense

abcrane123 (approved):

PR #1527, opened 2024-10-28
feat: upgrade viem to support heterogeneous calls

abcrane123:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


package.json
```diff
@@ -70,7 +70,7 @@
     "graphql-request": "^6.1.0",
     "jsdom": "^24.1.0",
     "packemon": "3.3.1",
-    "permissionless": "^0.1.29",
+    "permissionless": "^0.2.10",
```
alessey, line 73: can we remove this one as its a dependency too?

src/transaction/components/TransactionProvider.tsx
```diff
@@ -136,32 +107,17 @@ export function TransactionProvider({
...
+    if (walletCapabilities[Capabilities.AtomicBatch]?.supported) {
+      return statusSendCalls;
+    }
+    return statusSendCall;
+  }, [statusSendCall, statusSendCalls, walletCapabilities]);
 
   // Transaction hash for single transaction (non-batched)
-  const singleTransactionHash =
-    writeContractTransactionHash || sendCallTransactionHash;
+  const singleTransactionHash = sendCallTransactionHash;
```
cpcramer, line 120: Do we still need this assignment? 

Can we remove this and update `useSendCall` `data` return value to be named `singleTransactionHash`?

src/transaction/components/Transaction.tsx
```diff
@@ -35,7 +34,6 @@ export function Transaction({
       calls={calls}
       capabilities={capabilities}
       chainId={accountChainId}
-      contracts={contracts}
```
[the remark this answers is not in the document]
abcrane123: i'm leaning towards just removing it to avoid confusion. i.e devs might think they need to pass calls to calls and contracts to contracts which would cause the order of execution to be unreliable. plus it is a super small change on the developer side (just changing prop name to calls).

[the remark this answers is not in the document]
alessey: If you left the calls and contracts error in place we could mark contracts as deprecated for a few releases before removing.  it would at least give the devs a heads up over the next few releases before the forcing them to update due to a breaking change.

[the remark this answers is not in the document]
abcrane123: curious what the rest of the team thinks but personally still leaning towards just removing it since it is such a small change on the dev side.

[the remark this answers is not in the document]
dschlabach: +1 on marking as deprecated for a few releases, then removing. It should be a simple change for the dev but nice to give them a bit of time to switch over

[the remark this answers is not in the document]
dschlabach: Actually, reconsidered this. Given it's a simple prop name change, I'd probably lean towards just making the breaking change now

playground/nextjs-app-router/components/demo/Transaction.tsx
```diff
@@ -114,9 +114,9 @@ function TransactionDemo() {
     <div className="mx-auto grid w-1/2 gap-8">
       <Transaction
         chainId={chainId ?? 84532} // something breaks if we don't have default network?
-        {...transactions}
         isSponsored={isSponsored}
         onStatus={handleOnStatus}
+        calls={transactions as Calls}
```
dschlabach, line 119: any way to avoid this?

src/transaction/components/TransactionProvider.tsx
```diff
@@ -93,75 +83,38 @@ export function TransactionProvider({
...
-    throw new Error(
-      'Transaction: One of contracts or calls must be provided as a prop to the Transaction component.',
-    );
-  }
-  if (calls && contracts) {
+  // Validate `calls` props
+  if (!calls) {
     throw new Error(
-      'Transaction: Only one of contracts or calls can be provided as a prop to the Transaction component.',
+      'Transaction: Calls must be provided as a prop to the Transaction component.',
```
dschlabach, line 89: nit: should we say "calls" (lowercase) to match the prop?

dschlabach (approved):

playground/nextjs-app-router/components/demo/Transaction.tsx
abcrane123: oh yeah unnecessary let me remove

src/transaction/utils/sendSingleTransactions.ts
```diff
@@ -1,18 +1,26 @@
...
+  const calls = transactions?.map((transaction) => {
+    if (isContract(transaction)) {
+      return {
+        data: encodeFunctionData({
+          abi: transaction?.abi,
+          functionName: transaction?.functionName,
+          args: transaction?.args,
+        }),
+        to: transaction?.address,
+      };
```
0xAlec, line 18: nice

src/wallet/utils/isWalletACoinbaseSmartWallet.ts
```diff
@@ -25,7 +25,7 @@ export async function isWalletACoinbaseSmartWallet({
     if (!code) {
       // no code at address, check that the initCode is deploying a Coinbase Smart Wallet
       // factory address is first 20 bytes of initCode after '0x'
-      const factoryAddress = userOp.initCode.slice(0, 42) as Address;
+      const factoryAddress = userOp?.initCode?.slice(0, 42) as Address;
```
alessey, line 28: can update [test](https://github.com/coinbase/onchainkit/blob/main/src/wallet/utils/isWalletACoinbaseSmartWallet.test.ts#L19) and [docs](https://github.com/coinbase/onchainkit/blob/main/site/docs/pages/wallet/is-wallet-a-coinbase-smart-wallet.mdx) and I think you should be able to remove permissionless dependency

package.json
abcrane123: removed

src/transaction/components/TransactionProvider.tsx
```diff
@@ -93,75 +84,45 @@ export function TransactionProvider({
 
   const { switchChainAsync } = useSwitchChain();
 
-  // Validate `calls` and `contracts` props
-  if (!contracts && !calls) {
```
0xAlec, line 97: nit: this is the same conditional

src/transaction/components/TransactionProvider.tsx
```diff
@@ -93,75 +84,45 @@ export function TransactionProvider({
...
-        };
-    return transactionStatuses[transactionType];
-  }, [
-    statusSendCalls,
-    statusWriteContracts,
-    statusSendCall,
-    statusWriteContract,
-    transactionType,
-    walletCapabilities[Capabilities.AtomicBatch],
-  ]);
```
0xAlec, line 160: god bless

src/transaction/types.ts
```diff
@@ -107,17 +104,18 @@ export type SendBatchedTransactionsParams = {
...
 
 export type SendSingleTransactionParams = {
   sendCallAsync: SendTransactionMutateAsync<Config, unknown> | (() => void);
-  transactions: Call[] | ContractFunctionParameters[];
-  transactionType: string;
-  writeContractAsync: WriteContractMutateAsync<Config, unknown> | (() => void);
+  transactions:
+    | Call[]
+    | ContractFunctionParameters[]
+    | (Call | ContractFunctionParameters)[];
```
0xAlec, line 118: is this the same type?

0xAlec:
```suggestion
transactions: (Call | ContractFunctionParameters)[];
```

src/transaction/components/TransactionProvider.tsx
abcrane123: reverted

0xAlec (approved):

PR #1529, opened 2024-10-28
feat: add `onConnect` handler to `<ConnectWallet />`

dschlabach:
**What changed? Why?**
We're adding an `onConnect` handler to `<ConnectWallet />` so that developers can perform arbitrary actions after users click the "Connect Wallet" button. For example, this can be used to support SIWE without requiring a second button click/popup.

While this can be done with wagmi today, this improves DX for devs who wan to perform an action only after a user clicks the `ConnectWallet` button (vs. re-connecting on page reload/re-render).

**Notes to reviewers**

**How has it been tested?**
- Added tests
- Manual testing across browsers, including smart wallets, EOAs, and RainbowKit


src/wallet/components/ConnectWallet.tsx
```diff
@@ -24,12 +25,16 @@ export function ConnectWallet({
...
   withWalletAggregator = false,
+  onConnect,
 }: ConnectWalletReact) {
   // Core Hooks
   const { isOpen, setIsOpen } = useWalletContext();
   const { address: accountAddress, status } = useAccount();
   const { connectors, connect, status: connectStatus } = useConnect();
 
+  // State
+  const [hasClickedConnect, setHasClickedConnect] = React.useState(false);
```
dschlabach, line 36: Ideally we'd be able to avoid this piece of state, but there's no other way to detect whether `status === 'connected'` is true because:

1.  the user has just signed in OR
2. the user has refreshed the page

src/wallet/components/ConnectWallet.tsx
```diff
@@ -80,7 +94,10 @@ export function ConnectWallet({
         <ConnectButton
           className={className}
           connectWalletText={connectWalletText}
-          onClick={() => connect({ connector })}
+          onClick={() => {
+            connect({ connector });
```
dschlabach, line 98: If we didn't need to support the RainbowKit use case, we could just do:

```
connect({ connector }, { onSuccess: onConnect })
```

src/wallet/components/ConnectWallet.tsx
```diff
@@ -80,7 +96,16 @@ export function ConnectWallet({
...
           className={className}
           connectWalletText={connectWalletText}
-          onClick={() => connect({ connector })}
+          onClick={() => {
+            connect(
+              { connector },
+              {
+                onSuccess: () => {
+                  onInitialConnect?.();
+                },
```
dschlabach, line 105: Safer, more reliable way of triggering the callback. We have to use the useEffect and state for RainbowKit support

src/wallet/components/ConnectWallet.tsx
```diff
@@ -58,6 +63,14 @@ export function ConnectWallet({
     setIsOpen(!isOpen);
   }, [isOpen, setIsOpen]);
 
+  // Effects
+  React.useEffect(() => {
```
brendan-defi, line 67: it looks like the convention is to import hooks directly. `React.____` seems to be used for types, but not for hooks.

dschlabach: Thanks! Fixed

yarn.lock
```diff
@@ -2234,9 +2234,9 @@ __metadata:
   linkType: hard
 
 "@emotion/hash@npm:^0.9.0":
-  version: 0.9.1
-  resolution: "@emotion/hash@npm:0.9.1"
-  checksum: cdafe5da63fc1137f3db6e232fdcde9188b2b47ee66c56c29137199642a4086f42382d866911cfb4833cae2cc00271ab45cad3946b024f67b527bb7fac7f4c9d
```
cpcramer, line 2239: Was it intended to update this `yarn.lock` file?

cpcramer:
What are we updating with the `playground/nextjs-app-router/bun.lockb` file?

src/wallet/types.ts
```diff
@@ -20,6 +21,11 @@ export type ConnectWalletReact = {
   /** @deprecated Prefer `ConnectWalletText component` */
   text?: string; // Optional text override for button
   withWalletAggregator?: boolean; // Optional flag to enable the wallet aggregator like RainbowKit
+  /**
+   * Optional callback function to execute when the wallet is connected.
+   * This can be used to perform additional actions after the wallet is connected.
+   */
+  onInitialConnect?: () => void;
```
cpcramer, line 28: nit: let's keep this comment inline with `onInitialConnect`, similar to the other comments in this file

cpcramer:
Let's add a changeset as well!

dschlabach:
@cpcramer - `bun.lockb` and `yarn.lock` files were reset to main. Not sure how they got messed up but I was installing/removing a few packages for testing

src/wallet/types.ts
dschlabach: Fair enough, updated.

BTW the reason for using multiline comments is that the comment then shows up in Intellisense. I think this helps improve DX (internally and externally) but I understand that we already have a convention around `//` style comments.
  
CleanShot 2024-10-30 at 08 09 03@2x

src/wallet/components/ConnectWallet.test.tsx
```diff
@@ -162,6 +167,71 @@ describe('ConnectWallet', () => {
...
+      <ConnectWallet
+        text="Connect Wallet"
+        onInitialConnect={onInitialConnectMock}
+      />,
+    );
+
+    expect(onInitialConnectMock).toHaveBeenCalledTimes(1);
+  });
+
+  it('should not call onConnect callback when component is first mounted', () => {
```
alessey, line 217: onInitialConnect

src/wallet/components/ConnectWallet.test.tsx
```diff
@@ -215,5 +290,32 @@ describe('ConnectWallet', () => {
       fireEvent.click(button);
       expect(openConnectModalMock).toHaveBeenCalled();
     });
+
+    it('should call onConnect callback when connect button is clicked', () => {
```
alessey, line 294: onInitialConnect

cpcramer (approved):

0xAlec (approved):

alessey (approved):

PR #1547, opened 2024-10-30
chore: Update hasCopyAddressOnClick

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


site/docs/pages/swap/types.mdx
```diff
@@ -133,6 +133,14 @@ type SwapQuote = {
 };
 ```
 
+## `SwapConfig`
+
+```ts
+type SwapConfig = {
+  maxSlippage: number; // Maximum acceptable slippage for a swap (e.g., 3 for 3%). This is a percentage, not basis points
+};
+```
```
cpcramer, line 142: Flagged from a user that we had no documentation of the maxSlippage param.

site/docs/pages/swap/types.mdx
```diff
@@ -139,7 +139,9 @@ type SwapQuote = {
 type SwapReact = {
   children: ReactNode;
   className?: string; // Optional className override for top div element.
-  config?: SwapConfig;
+  config?: {
+    maxSlippage: number; // Maximum acceptable slippage for a swap (e.g., 3 for 3%). This is a percentage, not basis points;
```
cpcramer, line 143: Flagged from a user that we had no documentation of the maxSlippage param. SwapConfig is not an exported type and this page does not support twoslash

src/identity/components/IdentityLayout.tsx
```diff
@@ -35,7 +35,9 @@ export function IdentityLayout({
     return {
       avatar: childrenArray.find(findComponent(Avatar)),
       name: childrenArray.find(findComponent(Name)),
-      address: childrenArray.find(findComponent(Address)),
+      address: childrenArray.find(findComponent(Address))
+        ? cloneElement(childrenArray.find(findComponent(Address))!, { onClick })
+        : undefined,
```
alessey, line 40: you could move this childrenArray.find up above your return and not have to loop twice

src/identity/components/Address.tsx
```diff
@@ -36,7 +47,7 @@ export function Address({
           setTimeout(() => setCopyText('Copy'), 2000);
         })
         .catch((err) => {
-          console.error('Failed to copy address: ', err);
+          console.error('Failed to copy: ', err);
         });
     }
   };
```
alessey, line 53: i know this is unchanged, but making a span behave like a button instead of using a button will never be as good, maybe we can revisit this

src/identity/components/Identity.tsx
```diff
@@ -9,34 +8,22 @@ export function Identity({
   chain,
   children,
   className,
-  hasCopyAddressOnClick = false,
+  hasCopyAddressOnClick = true,
```
alessey, line 11: i don't think you need a default here at all since Address defaults true

alessey (approved):

0xAlec (approved):

dschlabach (approved):

PR #1575, opened 2024-11-05
chore: Update OnchainKitProvider to fallback to CB verified schemaId

cpcramer:
**What changed? Why?**
Default to use the `COINBASE_VERIFIED_ACCOUNT_SCHEMA_ID` in the `OnchainKitProvider` when the `schemaID` is missing. This is to prevent the `DisplayBadge` component from throwing an error. 
**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/.changeset/quick-fans-tease.md
+++ b/.changeset/quick-fans-tease.md
@@ -1,0 +1,5 @@
+---
+'@coinbase/onchainkit': patch
+---
+
+-**chore**: Updated `OnchainKitProvider` to fallback to CB verified schemaID. By @cpcramer #1575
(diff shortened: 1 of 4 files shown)


PauloFCP1 (approved):

src/identity/components/DisplayBadge.tsx
```diff
@@ -21,7 +21,7 @@ export function DisplayBadge({ children, address }: DisplayBadgeReact) {
   const attestations = useAttestations({
     address: address ?? contextAddress,
     chain: chain,
-    schemaId: contextSchemaId ?? schemaId,
+    schemaId: contextSchemaId ?? schemaId ?? null,
```
cpcramer, line 24: I had to add this to remove a type error:

```
Type '`0x${string}` | null | undefined' is not assignable to type '`0x${string}`'.
  Type 'undefined' is not assignable to type '`0x${string}`'.ts(2322)
types.ts(193, 3): The expected type comes from property 'schemaId' which is declared here on type 'UseAttestations'
```

alessey (approved):

brendan-defi (approved):

PR #1585, opened 2024-11-05
feat: Global Toast

brendan-defi:
**What changed? Why?**
* created global `<Toast>` component
* added Toast to playground with swap and transaction options

**Notes to reviewers**

**How has it been tested?**


brendan-defi:
closing this to start from scratch as we decided not to export the Toast component

PR #1588, opened 2024-11-05
feat: toast component

brendan-defi:
**What changed? Why?**
* created global <Toast> component

**Notes to reviewers**

**How has it been tested?**


src/internal/components/Toast.tsx
```diff
@@ -0,0 +1,76 @@
...
+  setIsToastVisible: (isToastVisible: boolean) => void;
+  children: React.ReactNode;
+};
+
+export function Toast({
+  className,
+  durationMs = 3000,
+  position = 'bottom-center',
+  isToastVisible,
+  setIsToastVisible,
```
alessey, line 19: nit: these could probably just be visible/setIsVisible as they are in the toast

brendan-defi: happy to change the name, but my intention was to have this state live in the Toast's parent in case developers want to have other triggers close the toast. though i suppose now that we are keeping the component internal, that might not be necessary. 

what do you think, should the state live in this component?

brendan-defi: ahh, nevermind, I'm sorry I misunderstood your point. 

I will remove the "Toast" from the label of these props in the Toast component, and they'll typically be used something like `isVisible={isToastVisible}`.

src/internal/components/Toast.test.tsx
```diff
@@ -0,0 +1,239 @@
+import { fireEvent, render } from '@testing-library/react';
+import '@testing-library/jest-dom';
+import { describe, expect, it, vi } from 'vitest';
+import { Toast } from './Toast';
+
+describe('Toast component', () => {
+  it('bottom-right renders correctly', () => {
```
alessey, line 7: nit: it('should render bottom-right correctly')

brendan-defi: ack, will fix these up

src/transaction/components/TransactionToast.tsx
```diff
@@ -1,13 +1,12 @@
...
+import { useCallback } from 'react';
+import { Toast } from '../../internal/components/Toast';
 import type { TransactionToastReact } from '../types';
 import { useTransactionContext } from './TransactionProvider';
 
 export function TransactionToast({
   children,
   className,
-  durationMs = 3000,
+  durationMs = 8000,
```
alessey, line 9: shouldn't these still be 3000?

brendan-defi: That's what we currently have, but in my testing, 3 seconds seems really short. I'm barely able to read the toast and move my cursor over before it disappears.

alessey: @0xAlec thoughts?

src/swap/components/SwapToast.tsx
```diff
@@ -1,17 +1,16 @@
...
 import { successSvg } from '../../internal/svg/successSvg';
-import { getToastPosition } from '../../internal/utils/getToastPosition';
 import { getChainExplorer } from '../../network/getChainExplorer';
 import type { SwapToastReact } from '../types';
 import { useSwapContext } from './SwapProvider';
 
 export function SwapToast({
   className,
-  durationMs = 3000,
+  durationMs = 8000,
```
dschlabach, line 13: intended to leave?

dschlabach (approved):

src/swap/components/SwapToast.tsx
brendan-defi: i did, but per comments, will revert to 5000ms

src/transaction/components/TransactionToast.tsx
brendan-defi: reverting to 5000ms

abcrane123 (approved):
lgtm ! agree with 5 second comment

PR #1589, opened 2024-11-06
feat: `OnchainKitProvider` default dependencies

0xAlec:
**What changed? Why?**
- add default `WagmiProvider` and `QueryClientProvider` if not provided in the React context

**Notes to reviewers**
please review the default `WagmiProvider` configuration settings 🙏 

**How has it been tested?**
playground and unit test

### DIFF
--- a/playground/nextjs-app-router/components/OnchainProviders.tsx
+++ b/playground/nextjs-app-router/components/OnchainProviders.tsx
@@ -1,8 +1,10 @@
 'use client';
 import { ENVIRONMENT, ENVIRONMENT_VARIABLES } from '@/lib/constants';
 import { OnchainKitProvider } from '@coinbase/onchainkit';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import type { ReactNode } from 'react';
 import { http, createConfig } from 'wagmi';
+import { WagmiProvider } from 'wagmi';
 import { base, baseSepolia } from 'wagmi/chains';
 import { coinbaseWallet } from 'wagmi/connectors';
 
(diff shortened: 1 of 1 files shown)


src/createWagmiConfig.ts
```diff
@@ -0,0 +1,37 @@
...
+  appName,
+  appLogoUrl,
+}: CreateWagmiConfigParams) => {
+  return createConfig({
+    chains: [base, baseSepolia],
+    connectors: [
+      coinbaseWallet({
+        appName,
+        appLogoUrl,
+        preference: 'smartWalletOnly',
```
0xAlec, line 19: should we set this to `all`?

src/createWagmiConfig.ts
```diff
@@ -0,0 +1,37 @@
...
+    ssr: true,
+    transports: {
+      [base.id]: apiKey
+        ? http(`https://api.developer.coinbase.com/rpc/v1/base/${apiKey}`)
+        : http(),
+      [baseSepolia.id]: apiKey
+        ? http(
+            `https://api.developer.coinbase.com/rpc/v1/base-sepolia/${apiKey}`,
+          )
+        : http(),
```
0xAlec, line 34: uses Base networks by default and sets the RPC as the CDP Node

src/OnchainKitProvider.tsx
```diff
@@ -51,6 +56,47 @@ export function OnchainKitProvider({
...
+  if (!providedWagmiConfig && !providedQueryClient) {
+    return (
+      <WagmiProvider config={defaultConfig}>
+        <QueryClientProvider client={defaultQueryClient}>
+          <OnchainKitContext.Provider value={value}>
+            {children}
+          </OnchainKitContext.Provider>
+        </QueryClientProvider>
+      </WagmiProvider>
+    );
```
alessey, line 97: since Wagmi requires QueryClient, the only time 1 dep is missing should be when there is a QueryClient, but no Wagmi?   Seems like an edge case, but would be possible to support this usecase, no?  or, should we adjust the console.errors at least to only state using default if both don't exist and otherwise prompt the user to fix?

0xAlec: if 1 dep is missing, we shouldn't modify the default behavior and allow the app to display the typical error message instead

image

0xAlec: in my mind, if you provide 1 of the 2 deps then it should be the same logic as what it is currently (before this PR) which is `Unhandled Runtime Error`

src/createWagmiConfig.ts
alessey: seems like it'd be a more flexible implementation, unless we just really want to push to smart wallet.

alessey (approved):

dschlabach (approved):

src/createWagmiConfig.test.ts
```diff
@@ -0,0 +1,96 @@
...
+        transports: {
+          [base.id]: expect.any(Function),
+          [baseSepolia.id]: expect.any(Function),
+        },
+      }),
+    );
+    expect(coinbaseWallet).toHaveBeenCalledWith({
+      appName: undefined,
+      appLogoUrl: undefined,
+      preference: 'smartWalletOnly',
```
abcrane123, line 51: Is this going to force the components to use smart wallet ? thinking this might be related to the issue i'm having where i'm unable to connect to EOA

src/createWagmiConfig.ts
abcrane123: we need to add a second connector for EOA 
`
    coinbaseWallet({
      appName: 'OnchainKit',
      preference: 'eoaOnly',
    }),
`

PR #1603, opened 2024-11-08
Fund button - setupOnrampEventListeners util for setting up onSuccess, onEvent and onExit callbacks

rustam-cb:
**What changed? Why?**
Added util function `setupEventSubscriptions` that is used to specify `onEvent`, `onExit` and `onSuccess` callbacks

**Notes to reviewers**
Example of usage of `setupOnrampEventListeners`

```
import { setupOnrampEventListeners } from '@onchainkit/utils';

setupOnrampEventListeners({
  onEvent: (event: EventMetadata) => {
    console.log('Event received', event.eventName);
  },
  onExit: () => {
    console.log('Exit event received');
  },
  onSuccess: () => {
    console.log('Success event received');
  } ,
  host: 'https://somthing.coinbase.com', // Optional: The host of the Coinbase Onramp widget. Defaults to https://pay.coinbase.com.
});
```

**How has it been tested?**
Unit tests added


src/fund/types/events.ts
```diff
@@ -0,0 +1,40 @@
...
+};
+
+export type TransitionViewEvent = {
+  eventName: 'transition_view';
+  pageRoute: string;
+};
+
+export type PublicErrorEvent = {
+  eventName: 'error';
+  // TODO: Public error shape
```
steveviselli-cb, line 13: @rustam-cb we shouldn't just blindly copy-paste TODO comments from our old SDK. We can probably just delete this.

src/fund/utils/subscribeToWindowMessage.ts
```diff
@@ -0,0 +1,85 @@
+import { DEFAULT_ONRAMP_URL } from '../constants';
+import type { JsonObject } from '../jsonTypes';
+
+export enum MessageCodes {
+  LaunchEmbedded = 'launch_embedded',
+  AppReady = 'app_ready',
+  AppParams = 'app_params',
+  SigninSuccess = 'signin_success',
+  Success = 'success',
+  Exit = 'exit',
```
steveviselli-cb, line 10: @rustam-cb does this message still get published? Seems like we have an "event" message that is an exit event. We probably shouldn't have two versions of exit.

Same goes for error and exit. Can you see if those are maybe legacy events that we don't need to support any more?

src/fund/utils/subscribeToWindowMessage.ts
```diff
@@ -0,0 +1,85 @@
...
+  Event = 'event',
+  Error = 'error',
+  PixelReady = 'pixel_ready',
+  OnAppParamsNonce = 'on_app_params_nonce',
```
steveviselli-cb, line 14: @rustam-cb delete these two they are no longer supported.

src/fund/utils/subscribeToWindowMessage.ts
```diff
@@ -0,0 +1,85 @@
...
+ * Subscribes to a message from the parent window.
+ * @param messageCode A message code to subscribe to.
+ * @param onMessage Callback for when the message is received.
+ * @param shouldUnsubscribe Whether to unsubscribe after the first message is received.
+ * @param allowedOrigin The origin to allow messages from.
+ * @param onValidateOrigin Callback to validate the origin of the message.
+ * @returns
+ */
+export function subscribeToWindowMessage(
+  messageCode: MessageCode,
```
steveviselli-cb, line 36: @rustam-cb Seems like the only message code we really care about is event. Maybe we should just delete all the others?

src/fund/utils/setupEventSubscriptions.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+/**
+ * Subscribes to events from the Coinbase Onramp widget.
+ * @param onEvent - Callback for when any event is received.
+ * @param onExit - Callback for when an exit event is received.
+ * @param onSuccess - Callback for when a success event is received.
+ * @returns a function to unsubscribe from the event listener.
+ */
+export function setupEventSubscriptions({
```
steveviselli-cb, line 20: @rustam-cb I think we should include "onramp" in this function name, since this is specific to onramp events, and the fund component is not necessarily coupled to onramp. Maybe `subscribeToOnrampEvents` or `setupOnrampEventListeners`?

steveviselli-cb (commented):
Looking good just a couple small things. Also, since this is a public repo, it would be nice to include some examples of how you would use these utils in the PR description. That would give reviewers and watchers a better idea of what this change is for.

src/fund/jsonTypes.ts
```diff
@@ -0,0 +1,25 @@
+/**
+Matches a JSON object.
+This type can be useful to enforce some input to be JSON-compatible or as a super-type to be extended from. Don't use this as a direct return type as the user would have to double-cast it: `jsonObject as unknown as CustomResponse`. Instead, you could extend your CustomResponse type from it to ensure your type only uses JSON-compatible types: `interface CustomResponse extends JsonObject { … }`.
+@category JSON
+*/
+export type JsonObject = { [Key in string]?: JsonValue };
```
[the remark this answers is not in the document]
rustam-cb: I thought these were a specific set up types and added them separately the same way as it was in CBPay-js so that types.ts does not get bloated but I can move these if that is not a concern.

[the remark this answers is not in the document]
steveviselli-cb: Yeah types bloating doesn't seem like a big concern, the swap component has a 300+ line types file. I would typically advocate for smaller files in other repos, but I think it's better to be consistent in this repo.

src/fund/utils/subscribeToWindowMessage.ts
rustam-cb: `Error` is not used so removing it from here. 

`Exit` is still used.

src/fund/utils/subscribeToWindowMessage.ts
rustam-cb: @steveviselli-cb 
in onramp-widget we do use `app_params` as well.  If not that then we could just remove the `messageCode` from this function fully. 
https://sourcegraph.cbhq.net/github.cbhq.net/payments/onramp-widget/-/blob/src/pages/embed/index.page.tsx?L22

In this case are you suggesting to create a separate type for the messageCode with two items in it? (`event` and `app_params`)

steveviselli-cb: @rustam-cb the whole embeded flow is deprecated. We don't support it any more. The code is still there because customers might still be using old versions of our SDK, but it wont work correctly. Anything to do with embeded onramp should not be included in onchainkit.

I think we can probably remove the messageCode param from the function.

src/fund/utils/subscribeToWindowMessage.ts
steveviselli-cb: Isn't there an exit event?
```
export type ExitEvent = {
  eventName: 'exit';
  // biome-ignore lint/suspicious/noExplicitAny: <explanation>
  error?: any;
};
```
What's the difference?

rustam-cb: I'm a bit confused by these Events and MessageCode and which one we need to deprecate from those. Did you mean to remove `exit` from the MessageCode?

steveviselli-cb: Yeah I think so, can you take a quick look at the history of the old SDK and see why we have both a message code called exit and an event called exit? My impression was that the event was put it to replace the message code, and the message code was kept for backwards compatibility. But we should confirm this and just remove anything that's old.

rustam-cb: I could not find the reason why it was done as the event was just added without much context. I see that messageCode is only used in `embed` files, so I will remove it from the messageCode. 

Its still not fully clear to me what is going on there (not the flow but with all the structure like:

```
{
  eventName: 'something',
  data: {
     eventName: 'somethingELSE'
  }
}
```

But I think as I dig more into it I'll be able to better adjust it.

PR #1610, opened 2024-11-08
feat: Add WalletModal component

cpcramer:
**What changed? Why?**
How to use:
```
 <OnchainKitProvider
        apiKey={ENVIRONMENT_VARIABLES[ENVIRONMENT.API_KEY]}
        chain={base}
        config={{
            appearance: {
              name: 'Cat.town',
              logo: 'cattownlogo.png'
              mode: 'auto',
              theme: 'default',
          },
          wallet: {
              display: 'modal', // 'modal' | 'classic'
              termsUrl: 'https://www.coinbase.com/legal/privacy', 
              privacyUrl: 'https://www.coinbase.com/legal/cookie',
          },
        }}
      >
</OnchainKitProvider>
```
Screenshot 2024-11-20 at 8 36 40 PM

Screenshot 2024-11-20 at 8 36 49 PM

Screenshot 2024-11-20 at 8 36 58 PM

Screenshot 2024-11-20 at 8 37 11 PM

Screenshot 2024-11-20 at 8 37 36 PM

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/wallet/components/WalletModal.test.tsx
+++ b/src/wallet/components/WalletModal.test.tsx
@@ -267,22 +267,6 @@ describe('WalletModal', () => {
     );
   });
 
-  it('updates shouldRender state when isOpen changes', () => {
-    const { rerender } = render(
-      <WalletModal isOpen={false} onClose={mockOnClose} />,
-    );
-
-    expect(screen.queryByTestId('ockModalOverlay')).not.toBeInTheDocument();
-
-    rerender(<WalletModal isOpen={true} onClose={mockOnClose} />);
-    expect(screen.getByTestId('ockModalOverlay')).toBeInTheDocument();
-
-    rerender(<WalletModal isOpen={false} onClose={mockOnClose} />);
-    const overlay = screen.getByTestId('ockModalOverlay');
-    fireEvent.transitionEnd(overlay);
-    expect(screen.queryByTestId('ockModalOverlay')).not.toBeInTheDocument();
-  });
-
   it('handles WalletConnect connection errors', () => {
     const mockError = new Error('WalletConnect connection failed');
     const mockOnError = vi.fn();
@@ -374,5 +358,14 @@ describe('WalletModal', () => {
       '_blank',
       'noopener,noreferrer',
     );
+  });
+
+  it('closes modal on Enter key press on overlay', () => {
+    render(<WalletModal isOpen={true} onClose={mockOnClose} />);
+
+    const overlay = screen.getByTestId('ockModalOverlay');
+    fireEvent.keyDown(overlay, { key: 'Enter' });
+
+    expect(mockOnClose).toHaveBeenCalled();
   });
 });
--- a/src/wallet/components/WalletModal.tsx
+++ b/src/wallet/components/WalletModal.tsx
@@ -82,12 +82,6 @@ export function WalletModal({
     };
   }, [isOpen, onClose]);
 
-  const _handleAnimationEnd = () => {
-    if (!isOpen) {
-      setShouldRender(false);
-    }
-  };
-
   const appLogo = config?.appearance?.logo;
   const appName = config?.appearance?.name;
   const privacyPolicyUrl = config?.wallet?.privacyUrl ?? undefined;


playground/nextjs-app-router/package.json
```diff
@@ -17,7 +17,6 @@
     "watch:tailwind": "mkdir -p ./node_modules/@coinbase/onchainkit/src && nodemon --watch ../../src/styles.css --exec 'cp ../../src/styles.css ./node_modules/@coinbase/onchainkit/src/styles.css'"
   },
   "dependencies": {
-    "@coinbase/onchainkit": "^0.35.0",
```
abcrane123, line 20: why are we removing this ?

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+  border,
+  cn,
+  color,
+  line,
+  pressable,
+  text,
+} from '../../styles/theme';
+import { useOnchainKit } from '../../useOnchainKit';
+
+const BaseIcon = () => (
```
abcrane123, line 14: can we move these SVGs to their own files

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+    config?.wallet?.termsUrl ?? 'https://base.org/terms-of-service';
+
+  return (
+    <div
+      className={cn(
+        `${background.default}/50`,
+        'fixed inset-0 z-50 flex items-start justify-center',
+        !isOpen && 'hidden',
+      )}
+      onClick={() => onClose()}
```
abcrane123, line 97: this can just be `onClose`

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+  const { connect } = useConnect();
+  const connectors = useConnectors();
+  const { config } = useOnchainKit();
+
+  const appLogo = config?.appearance?.logo;
+  const appName = config?.appearance?.name;
+  const privacyPolicyUrl =
+    config?.wallet?.privacyUrl ?? 'https://base.org/privacy-policy';
+  const termsOfServiceUrl =
+    config?.wallet?.termsUrl ?? 'https://base.org/terms-of-service';
```
abcrane123, line 88: i would wrap these in a useMemo

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+        fill="white"
+      />
+    </g>
+  </svg>
+);
+
+interface WalletModalProps {
+  isOpen: boolean;
+  onClose: () => void;
+}
```
abcrane123, line 76: should we add an optional `className` prop here?

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+      <div
+        className={cn(
+          border.radius,
+          background.default,
+          line.default,
+          'mt-[116px] h-[390px] w-[323px] p-6 pb-4',
+          'border-x-0 border-t border-b-0',
+          'flex flex-col gap-4',
+        )}
+        onClick={(e) => e.stopPropagation()}
```
abcrane123, line 108: what is the purpose of the onClick here?

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+          </div>
+        )}
+
+        <div className="flex flex-col gap-3">
+          <button
+            onClick={() => {
+              // WalletPreference.SMART_WALLET
+              connect({ connector: connectors[0] });
+              onClose();
+            }}
```
abcrane123, line 141: can you move this to a function defined above and wrap in `useCallback`

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+              </span>
+            </div>
+          </div>
+
+          <button
+            onClick={() => {
+              // WalletPreference.EOA
+              connect({ connector: connectors[1] });
+              onClose();
+            }}
```
abcrane123, line 183: see comment above

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+          <button
+            onClick={() => {
+              const walletConnectConnector = connectors.find(
+                (c) => c.name === 'WalletConnect',
+              );
+              if (walletConnectConnector) {
+                connect({ connector: walletConnectConnector });
+                onClose();
+              }
+            }}
```
abcrane123, line 211: same here (move to function defined above wrapped in useCallback)

src/OnchainKitProvider.tsx
```diff
@@ -42,6 +42,11 @@ export function OnchainKitProvider({
           theme: config?.appearance?.theme ?? 'default',
         },
         paymaster: config?.paymaster || defaultPaymasterUrl,
+        wallet: {
+          display: config?.wallet?.display,
```
alessey, line 46: should we just default to classic here since it seems like that's what we are doing in code?

src/wallet/components/ConnectWallet.tsx
```diff
@@ -72,23 +74,17 @@ export function ConnectWallet({
...
-          )}
-        </ConnectButtonRainbowKit.Custom>
+        <div className="flex" data-testid="ockConnectWallet_Container">
+          <ConnectButton
+            className={className}
+            connectWalletText={connectWalletText}
+            onClick={() => setIsOpen(true)}
+            text={text}
+          />
+          <WalletModal isOpen={isOpen} onClose={() => setIsOpen(false)} />
```
alessey, line 86: extract onclose

src/wallet/components/WalletModal.tsx
alessey, line 88: any reason not to default these in config?

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+    config?.wallet?.privacyUrl ?? 'https://base.org/privacy-policy';
+  const termsOfServiceUrl =
+    config?.wallet?.termsUrl ?? 'https://base.org/terms-of-service';
+
+  return (
+    <div
+      className={cn(
+        `${background.default}/50`,
+        'fixed inset-0 z-50 flex items-start justify-center',
+        !isOpen && 'hidden',
```
alessey, line 95: instead of just hiding it, wouldn't we want to return null and not render this at all?

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+                  className="h-full w-full object-cover"
+                />
+              </div>
+            )}
+            {appName && (
+              <h2
+                className={cn(
+                  text.headline,
+                  color.foreground,
+                  'font-dmSans font-semibold',
```
alessey, line 126: should this be using theme fonts?

src/wallet/components/WalletModal.tsx
alessey, line 211: extract to usecallback

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+                onClose();
+              }
+            }}
+            className={cn(
+              border.radiusInner,
+              line.default,
+              text.label2,
+              pressable.default,
+              color.inverse,
+              color.foreground,
```
alessey, line 218: does foreground overwrite inverse?

playground/nextjs-app-router/package.json
cpcramer: We aren't, this is just for my local dev

src/wallet/components/WalletModal.tsx
cpcramer: Yep! Sorry, I was planning to. This PR wasn't really ready for review, just ready for component QA!

src/wallet/components/WalletModal.tsx
cpcramer: Good call

src/wallet/components/WalletModal.tsx
cpcramer: Good call, I added a check if it is open early in the component. If it's not opened then I'm now just returning null!

src/wallet/components/WalletModal.tsx
cpcramer: Good call

src/wallet/components/WalletModal.tsx
cpcramer: Extracting and memoizing all connectors now

src/wallet/components/WalletModal.tsx
cpcramer: I think this is a simple enough computation where the cost of recalculating is negligible vs creating and maintaining the memoized value

playground/nextjs-app-router/components/AppProvider.tsx
```diff
@@ -228,6 +228,11 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
             theme: componentTheme === 'none' ? undefined : componentTheme,
           },
           paymaster: paymasters?.[chainId || 8453]?.url,
+          wallet: {
+            display: 'modal',
+            termsUrl: 'https://www.coinbase.com/legal/privacy', // URL to the terms of service for the wallet modal
+            privacyUrl: 'https://www.coinbase.com/legal/cookie', // URL to the privacy policy for the wallet modal
```
alessey, line 234: switched?

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+  useEffect(() => {
+    if (isOpen) {
+      // Immediately render the modal when opening
+      setShouldRender(true);
+    } else {
+      // When closing, wait for animation to complete before removing from DOM
+      const timer = setTimeout(() => {
+        setShouldRender(false);
+      }, 100);
```
alessey, line 39: I think you should be able to use a transitionEnd event here instead.

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,261 @@
...
+  }, [isOpen]);
+
+  const appLogo = config?.appearance?.logo;
+  const appName = config?.appearance?.name;
+  const privacyPolicyUrl = config?.wallet?.privacyUrl ?? undefined;
+  const termsOfServiceUrl = config?.wallet?.termsUrl ?? undefined;
+
+  const handleSmartWalletConnector = useCallback(() => {
+    // WalletPreference.SMART_WALLET
+    connect({ connector: connectors[0] });
```
alessey, line 51: does this only work if a user uses our wagmi config?

src/wallet/components/ConnectWallet.tsx
```diff
@@ -63,32 +61,22 @@ export function ConnectWallet({
...
-        </ConnectButtonRainbowKit.Custom>
+        <div className="flex" data-testid="ockConnectWallet_Container">
+          <ConnectButton
+            className={className}
+            connectWalletText={connectWalletText}
+            onClick={() => setIsOpen(true)}
+            text={text}
+          />
+          <WalletModal isOpen={isOpen} onClose={handleClose} />
+        </div>
```
dschlabach, line 79: Does this break our `onConnect` callback?

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,261 @@
...
+    if (walletConnectConnector) {
+      connect({ connector: walletConnectConnector });
+      onClose();
+    }
+  }, [connect, connectors, onClose]);
+
+  const handleKeyDown = (e: React.KeyboardEvent) => {
+    if (e.key === 'Enter' || e.key === ' ') {
+      onClose();
+    }
```
alessey, line 74: wouldn't this generally close on esc vs. enter/space?

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,261 @@
...
+      className={cn(
+        'fixed inset-0 z-50 flex items-center justify-center',
+        isOpen ? 'animate-fadeIn' : 'animate-fadeOut',
+        'bg-black/70',
+        className,
+      )}
+      onClick={onClose}
+      onKeyDown={handleKeyDown}
+      role="presentation"
+      data-testid="modal-overlay"
```
alessey, line 97: ockModalOverlay?

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,261 @@
...
+    >
+      <div
+        className={cn(
+          border.default,
+          border.radius,
+          background.default,
+          line.default,
+          'w-[323px] p-6 pb-4',
```
alessey, line 105: should this be full screen on mobile?

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,261 @@
...
+        role="dialog"
+        aria-modal="true"
+      >
+        <button
+          type="button"
+          onClick={onClose}
+          className={cn(
+            'absolute top-4 right-4',
+            'flex items-center justify-center',
+            'h-[12px] w-[12px]',
```
alessey, line 122: h-3 w-3

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,261 @@
...
+            type="button"
+            onClick={handleCoinbaseWalletConnector}
+            className={cn(
+              border.default,
+              border.radiusInner,
+              line.default,
+              text.label2,
+              pressable.alternate,
+              color.foreground,
+              'h-[40px] w-[275px] px-4 py-2.5',
```
alessey, line 199: h-10

dschlabach:
Getting this flicker on disconnect

https://github.com/user-attachments/assets/552e76a6-cd43-4617-8da8-bfeded026803

src/OnchainKitConfig.ts
```diff
@@ -15,6 +16,11 @@ export const ONCHAIN_KIT_CONFIG: OnchainKitConfig = {
       theme: null,
     },
     paymaster: null,
+    wallet: {
+      display: null,
+      termsUrl: DEFAULT_TERMS_URL,
+      privacyUrl: DEFAULT_PRIVACY_URL,
```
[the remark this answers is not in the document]
cpcramer: Ah I agree provider is better. I think I misunderstood your previous comment about defaulting in the "config"

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,261 @@
...
+      onClose();
+    }
+  }, [connect, connectors, onClose]);
+
+  const handleKeyDown = (e: React.KeyboardEvent) => {
+    if (e.key === 'Enter' || e.key === ' ') {
+      onClose();
+    }
+  };
+
```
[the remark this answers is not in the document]
cpcramer: TIL trap focus - great call

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,261 @@
...
+              'relative h-full w-full transition-colors',
+              '[&>svg>path]:hover:fill-[var(--ock-icon-color-foreground-muted)]',
+            )}
+          >
+            {closeSvg}
+          </div>
+        </button>
+
+        {(appLogo || appName) && (
+          <div className="mt-3 flex w-[275px] flex-col items-center gap-3 self-stretch p-2">
```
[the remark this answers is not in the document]
cpcramer: I don't believe there is a way around them if we want specific styling outside of the tailwind built in styles

src/wallet/components/WalletModal.tsx
cpcramer: I was getting a lint error that required close on enter + space

src/wallet/components/WalletModal.tsx
cpcramer: I like how it currently looks on mobile:

Screenshot 2024-11-21 at 9 43 58 AM

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+        <div
+          className={cn(
+            color.foregroundMuted,
+            text.legal,
+            'flex flex-col items-center justify-center gap-1 px-4',
+            'mt-4 w-[275px] text-center leading-3',
+          )}
+        >
+          <span className="font-normal text-[10px] leading-[13px]">
```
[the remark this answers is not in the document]
cpcramer: Agreed, would love if our linter could enforce this

src/OnchainKitProvider.tsx
```diff
@@ -47,6 +47,11 @@ export function OnchainKitProvider({
           theme: config?.appearance?.theme ?? 'default',
         },
         paymaster: config?.paymaster || defaultPaymasterUrl,
+        wallet: {
+          display: config?.wallet?.display ?? 'classic',
+          termsUrl: config?.wallet?.termsUrl,
+          privacyUrl: config?.wallet?.privacyUrl,
```
brendan-defi, line 53: do we need to provide default values?

src/wallet/components/WalletModal.tsx
dschlabach: I like the drawer pattern on mobile but obviously that can be a future thing

https://vaul.emilkowal.ski/

src/wallet/components/WalletModal.tsx
cpcramer: good call

src/wallet/components/WalletModal.tsx
0xAlec: I don't think this should be hardcoded to the 1st connector...

cpcramer: @0xAlec What do you recommend?

src/constants.ts
```diff
@@ -4,3 +4,5 @@ export enum Capabilities {
   AuxiliaryFunds = 'auxiliaryFunds',
   PaymasterService = 'paymasterService',
 }
+export const DEFAULT_PRIVACY_URL = 'https://base.org/privacy-policy';
+export const DEFAULT_TERMS_URL = 'https://base.org/terms-of-service';
```
0xAlec, line 8: I don't think these are the right privacy and terms of service to be using as defaults?

src/wallet/components/WalletModal.tsx
0xAlec: you also need to handle the case where the user did not provide an appropriate connector

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+  const handleSmartWalletConnector = useCallback(() => {
+    // WalletPreference.SMART_WALLET
+    connect({ connector: connectors[0] });
+    onClose();
+  }, [connect, connectors, onClose]);
+
+  const handleCoinbaseWalletConnector = useCallback(() => {
+    // WalletPreference.EOA
+    connect({ connector: connectors[1] });
```
0xAlec, line 56: this will throw an index out of error range if the user doesn't provide more than one connector

src/wallet/components/WalletModal.tsx
0xAlec: you can use the `provider` from a connector and check the CB Wallet SDK `appMetadata` field to check what `preference` is set to

cpcramer: Sweet, thanks! I'll refactor this

src/constants.ts
cpcramer: Will double check with Mind

src/wallet/components/ConnectWallet.tsx
```diff
@@ -72,23 +84,22 @@ export function ConnectWallet({
...
-                text={text}
-              />
-            </div>
+        <div className="flex" data-testid="ockConnectWallet_Container">
+          <ConnectButton
+            className={className}
+            connectWalletText={connectWalletText}
+            onClick={() => {
+              handleOpenModal();
+              setHasClickedConnect(true);
```
dschlabach, line 95: Another way to do this (and to remove the state for `hasClickedConnect` would be to just set if the modal _was_ open and the `status === 'connected'`, but I think it's fine to leave for now. 

Previously we didn't have full control over the modal with Rainbowkit

src/wallet/components/WalletModal.tsx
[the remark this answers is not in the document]
dschlabach: Yeah that's the syntax I'm just saying it's generally bad practice to hardcode pixel values like this especially with Tailwind.

It's fine for some one-offs but it makes it a lot harder for other devs to reason about/change in the future. Plus, if you think of Tailwind as a design system, it's also nice to keep all of our spacing, widths, and sizes within that system.

Probably this case is different, but if there's spacing/width/height values that we're using frequently, we should add them to the Tailwind config so that it's part of the design system

cc @mindapivessa

src/wallet/components/WalletModal.tsx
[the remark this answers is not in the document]
dschlabach: I can spend some time looking at the linter, but Biome generally has poor support for linting/formatting Tailwind.

But do we use text-10px frequently? If so, we can add it to the Tailwind spacing system so that it can be used across the codebase

cc @mindapivessa

src/wallet/components/ConnectWallet.tsx
```diff
@@ -24,11 +25,13 @@ export function ConnectWallet({
...
   // but for now we will keep it for backward compatibility.
   text = 'Connect Wallet',
-  withWalletAggregator = false,
   onConnect,
 }: ConnectWalletReact) {
+  const { config = { wallet: { display: undefined } } } = useOnchainKit();
+
   // Core Hooks
   const { isOpen, setIsOpen } = useWalletContext();
+  const [isModalOpen, setIsModalOpen] = useState(false);
```
dschlabach, line 34: Still getting an issue trapping focus on the modal

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,296 @@
...
+  const privacyPolicyUrl = config?.wallet?.privacyUrl ?? undefined;
+  const termsOfServiceUrl = config?.wallet?.termsUrl ?? undefined;
+
+  const handleCoinbaseWalletConnection = useCallback(() => {
+    try {
+      const cbConnector = coinbaseWallet({
+        preference: 'all',
+      });
+
+      connect({ connector: cbConnector });
```
dschlabach, line 61: Could you update this component to accept `onConnect` as a prop, then pass it to the `onSuccess` callback here? That would be a lot cleaner than the solution I initially implemented in `ConnectWallet` where we have to track `hasClickedConnect`

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,296 @@
...
+  const handleWalletConnectConnector = useCallback(() => {
+    try {
+      const walletConnectConnector = connectors.find(
+        (c) => c.type === 'walletConnect',
+      );
+      if (!walletConnectConnector) {
+        console.error('WalletConnect connector not configured');
+        return;
+      }
+      connect({ connector: walletConnectConnector });
```
dschlabach, line 84: And same here... it'd look like this:

```
connect(
              { connector },
              {
                onSuccess: () => {
                  onConnect?.();
                },
              },
            );
```

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,296 @@
...
+      onClose();
+    } catch (error) {
+      console.error('WalletConnect connection error:', error);
+    }
+  }, [connect, connectors, onClose]);
+
+  const handleKeyDown = useCallback(
+    (e: React.KeyboardEvent | KeyboardEvent) => {
+      if (['Escape', 'Enter', ' '].includes(e.key)) {
```
dschlabach, line 93: We should remove enter from this - if you press "Enter" while focused on one of the buttons in the modal, then the modal is closed instead of the button being pressed

src/wallet/components/WalletModal.test.tsx
```diff
@@ -0,0 +1,314 @@
...
+        onClose={mockOnClose}
+        className={customClass}
+      />,
+    );
+
+    const overlay = screen.getByTestId('ockModalOverlay');
+    expect(overlay).toHaveClass(customClass, 'bg-black/70');
+  });
+
+  it('closes modal on Enter, Space, or Escape key press', () => {
```
dschlabach, line 213: Prob need to update this now

dschlabach (approved):

playground/nextjs-app-router/components/AppProvider.tsx
```diff
@@ -228,6 +228,11 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
             theme: componentTheme === 'none' ? undefined : componentTheme,
           },
           paymaster: paymasters?.[chainId || 8453]?.url,
+          wallet: {
+            display: 'modal',
+            termsUrl: 'https://www.coinbase.com/legal/cookie', // URL to the terms of service for the wallet modal
+            privacyUrl: 'https://www.coinbase.com/legal/privacy', // URL to the privacy policy for the wallet modal
+          },
```
brendan-defi, line 235: do we also want to give playground users the ability to see `wallet.display.classic`?

playground/nextjs-app-router/components/OnchainProviders.tsx
```diff
@@ -24,6 +24,17 @@ export const config = createConfig({
       appName: 'OnchainKit',
       preference: 'eoaOnly',
     }),
+    walletConnect({
+      projectId:
+        ENVIRONMENT_VARIABLES[ENVIRONMENT.WALLETCONNECT_PROJECT_ID] ?? '',
```
brendan-defi, line 29: let's add this env var to `.env.local.example`

src/wallet/components/WalletModal.tsx
```diff
@@ -0,0 +1,333 @@
...
+};
+
+export function WalletModal({
+  isOpen,
+  onClose,
+  className,
+  onError,
+}: WalletModalProps) {
+  const modalRef = useRef<HTMLDivElement>(null);
+  const [shouldRender, setShouldRender] = useState(isOpen);
```
brendan-defi, line 33: why do we need `shouldRender` if we have `isOpen`? is this just to handle animation delays?

PR #1612, opened 2024-11-09
feat: Add useOutsideClick hook

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/useClickOutside.ts
+++ b/src/useClickOutside.ts
@@ -10,6 +10,7 @@ export function useClickOutside(
         return;
       }
 
+      // Check if the clicked target is outside of the referenced element
       if (!elRef.current.contains(e.target as Node)) {
         callback();
       }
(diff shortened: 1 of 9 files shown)


src/useClickOutside.ts
```diff
@@ -0,0 +1,33 @@
...
+      if (!elRef.current.contains(e.target as Node)) {
+        callback();
+      }
+    },
+    [callback, elRef],
+  );
+
+  useEffect(() => {
+    // Add click event listener when component mounts
+    document.addEventListener('click', handleClickOutside, { capture: true });
```
alessey, line 23: if you're using capture, should you stopPropagation to skip bubbling altogether?

src/useClickOutside.ts
```diff
@@ -0,0 +1,33 @@
+import { useCallback, useEffect } from 'react';
+
+export function useClickOutside(
```
alessey, line 3: nit: I've generally seen this as a useOutsideClick hook 🤷

cpcramer: updated

src/useClickOutside.ts
cpcramer: Do we want to skip bubbling entirely? It seems beneficial in situations where the parent components might need to handle / know about clicks

I like using capture because it doesn't interrupt event flow, just determines when the handler runs.

dschlabach (approved):

PR #1616, opened 2024-11-12
feat: enable deep linking in playground

dschlabach:
**What changed? Why?**

This PR enables deep linking on the playground. This will allow us to share component configurations with users and team members, leading to better DX and easier debugging.

**Notes to reviewers**
- Paymaster options is left using `localStorage` for now as it contains an RPC URL which is sensitive

**How has it been tested?**


dschlabach:
Moving to #1624

PR #1617, opened 2024-11-13
fix: fix CLI's `npm` build

dschlabach:
**What changed? Why?**
Our CLI template has mismatching versions of `viem` and `wagmi` from `onchainkit`, causing type errors and the build to fail. This PR updates the template's `package.json` to have versions of `viem` and `wagmi` that match `onchainkit`'s.

**Notes to reviewers**
- Also adds CI check that builds package, runs it, installs deps, and builds the project with npm

**How has it been tested?**

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -19,24 +19,21 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v4
-        with:
-          node-version: ${{ matrix.node-version }}
-          cache: 'npm'
-
       - name: Setup Bun
         uses: oven-sh/setup-bun@v1
         with:
           bun-version: latest
-  
+
       - name: Install dependencies
         working-directory: ./create-onchain
         run: bun install
-  
+
       - name: Build CLI
         working-directory: ./create-onchain
         run: bun run build
+
+      - name: Make CLI executable
+        run: chmod +x ./create-onchain/dist/esm/cli.js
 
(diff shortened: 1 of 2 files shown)


create-onchain/templates/next/package.json
```diff
@@ -14,8 +14,8 @@
     "react": "^18",
     "react-dom": "^18",
     "@tanstack/react-query": "^5",
-    "wagmi": "^2.12.25",
-    "viem": "^2.21.40"
+    "viem": "^2.17.4",
+    "wagmi": "^2.11.0"
```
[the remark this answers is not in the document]
dschlabach: This is just the CLI template (matches OnchainKit's current peer deps)

Once @abcrane123 merges her latest upgrades, we can update CLI accordingly

.github/workflows/cli.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Create test project
+        run: |
+          mkdir test-project
+          cd test-project
+          yes "" | ../create-onchain/dist/esm/cli.js
+        env:
+          CI: true
+
+      - name: List directories
+        run: ls -la test-project
```
brendan-defi, line 47: if the purpose here is to confirm that create-onchain worked correctly, should we turn an explicit test for a file that we'd expect? 

something like:
```
test -d test-project/my-onchainkit-app/app/page.tsx || exit 1
test "$(ls -A test-project/my-onchainkit-app)" || exit 1
```

dschlabach: Sorry, that was for testing purposes only - will remove

abcrane123 (approved):

PR #1620, opened 2024-11-13
Feat: toast animations

brendan-defi:
**What changed? Why?**
* fixed stuttering issue with Toast animations
* create two new animation types (`enterUp` and `enterDown`)
* created default animation behavior based on position
* added animation prop to Toast to override default

**Notes to reviewers**

**How has it been tested?**
* [x] full test coverage
* [x] locally in playground


dschlabach (approved):

src/internal/utils/getToastAnimation.ts
```diff
@@ -0,0 +1,16 @@
...
+      return 'animate-enterRight';
+    case 'top-center':
+      return 'animate-enterDown';
+    case 'bottom-right':
+      return 'animate-enterRight';
+    case 'bottom-center':
+      return 'animate-enterUp';
+    default:
+      throw new Error('Invalid toast position');
+  }
```
alessey, line 15: if we aren't trying to reuse this for other animations it could be a simple map in the toast component, i.e. `{ 'top-right': 'animate-enterRight', 'top-center': 'animate-enterDown', ...}`

brendan-defi: sounds good. i was trying to follow the `getToastPosition` pattern, but a simple const is simpler :-)

brendan-defi: all set

alessey:
cc: @alissacrane-cb was there a requirement around animations always coming in from the right?

alessey (approved):

PR #1623, opened 2024-11-14
refactor: refactor `DemoOptions` component

dschlabach:
**What changed? Why?**
This PR refactors the `DemoOptions` component away from a switch/case and to a mapping pattern. I think this is appropriate for 2 reasons:
1. Mapping is easier to read and maintain without stepping through nested logic.
2. (more importantly) it will enable the next playground feature, the ability to copy "share links". This requires us to know _which options_ are associated with _which components_, and this is much easier to do with an accessible mapping available.

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/playground/nextjs-app-router/components/form/checkout-options.tsx
+++ b/playground/nextjs-app-router/components/form/checkout-options.tsx
@@ -3,11 +3,10 @@ import { Label } from '@/components/ui/label';
 import { RadioGroup, RadioGroupItem } from '@/components/ui/radio-group';
 import { useContext } from 'react';
 import { useState } from 'react';
-import { AppContext, CheckoutTypes, OnchainKitComponent } from '../AppProvider';
+import { AppContext, CheckoutTypes } from '../AppProvider';
 
 export function CheckoutOptions() {
   const {
-    activeComponent,
     checkoutTypes,
     setCheckoutTypes,
     checkoutOptions,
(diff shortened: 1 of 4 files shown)


playground/nextjs-app-router/components/form/checkout-options.tsx
```diff
@@ -23,78 +23,78 @@ export function CheckoutOptions() {
...
-                setProductOptions({
-                  ...productOptions,
-                  chargeId: e.target.value,
-                });
-                setCheckoutOptions?.({
-                  ...checkoutOptions,
-                  chargeId: e.target.value,
-                });
-              }}
+  return activeComponent === OnchainKitComponent.Checkout ? (
```
dschlabach, line 26: This is required so the component doesn't return `false` in any case.

playground/nextjs-app-router/components/form/transaction-options.tsx
```diff
@@ -17,40 +17,38 @@ export function TransactionOptions() {
...
-              Calls Callback
-            </SelectItem>
-            <SelectItem value={TransactionTypes.ContractsCallback}>
-              Contracts Callback
-            </SelectItem>
-          </SelectContent>
-        </Select>
-      </div>
-    )
+  return activeComponent === OnchainKitComponent.Transaction ||
```
dschlabach, line 20: This is required so the component doesn't return `false` in any case.

brendan-defi (approved):

playground/nextjs-app-router/components/form/checkout-options.tsx
alessey: if we have to have these checks for each component specific option is there any reason not to just include them all?

dschlabach: Yeah i'm actually not sure why this was included in the first place

alessey (approved):

PR #1624, opened 2024-11-14
feat: enable deep linking to playground

dschlabach:
**What changed? Why?**
This PR adds deep linking capability to the playground. If a user has a bug or we'd like to share component configs internally, this will make that process a lot easier.

Also, refactored some logic in `AppProvider` to a new hook, `useStateWithStorage`, which handles setting state and localStorage and initializing with query params, in the case of a shared link.

**Notes to reviewers**
- It would be helpful to get more eyes on this playground's preview link
- The only param this doesn't support today is the `paymasters` state - unsure if it's ok to expose the `paymasterUrl` in params

**How has it been tested?**
- Locally and in playground


playground/nextjs-app-router/lib/hooks.ts
```diff
@@ -23,3 +24,71 @@ export function useCapabilities() {
...
+
+  // Sync to localStorage only when state changes after initialization
+  useEffect(() => {
+    if (!isInitialized) {
+      return;
+    }
+
+    try {
+      if (state !== undefined) {
+        window.localStorage.setItem(key, serializer(state));
```
alessey, line 84: should we namespace our localStorage items?

playground/nextjs-app-router/components/AppProvider.tsx
```diff
@@ -144,45 +148,12 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
       connect({ connector: connectors[1] });
     }
   }, [connect, connectors, walletType]);
```
alessey, line 150: smart wallet popup is blocked on initial load

playground/nextjs-app-router/lib/url-params.ts
```diff
@@ -0,0 +1,68 @@
...
+  const relevantParams = getComponentQueryParams(activeComponent);
+  const params = new URLSearchParams();
+
+  for (const param of [...relevantParams, ...commonOptions]) {
+    const value = localStorage.getItem(param);
+    if (value) {
+      params.set(param, value);
+    }
+  }
+  return `${window.location.origin}?${params.toString()}`;
```
alessey, line 46: thoughts on only setting params that are not the default to reduce the length of the url?

alessey (approved):

playground/nextjs-app-router/lib/url-params.ts
dschlabach: done! 👍🏻

0xAlec (approved):

PR #1636, opened 2024-11-19
fix: flakey tx test

brendan-defi:
**What changed? Why?**
* updated a test in TransactionProvider
  * previously, test was flakey
  * made test more explicit, and added an additional mock to make sure we test the intended case

**Notes to reviewers**

**How has it been tested?**
via test suite

### DIFF
--- a/src/transaction/components/TransactionProvider.test.tsx
+++ b/src/transaction/components/TransactionProvider.test.tsx
@@ -289,6 +289,12 @@ describe('TransactionProvider', () => {
   });
 
   it('should emit onError when legacy transactions fail', async () => {
+    // need to mock sendWalletTransactions to prevent lifecycleStatus from being
+    // set to 'error' when sendWalletTransactions is called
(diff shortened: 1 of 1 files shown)


src/transaction/components/TransactionProvider.test.tsx
```diff
@@ -309,7 +315,12 @@ describe('TransactionProvider', () => {
     );
     fireEvent.click(button);
     await waitFor(() => {
-      expect(onErrorMock).toHaveBeenCalled();
+      expect(sendWalletTransactionsMock).toHaveBeenCalled();
+      expect(onErrorMock).toHaveBeenCalledWith({
+        code: 'TmTPc01',
+        error: '{}',
+        message: 'Something went wrong. Please try again.',
+      });
```
alessey, line 323: nit: could use expect.objectContaining here to only test pertinent parts of error

alessey (approved):

PR #1648, opened 2024-11-20
feat: Add `SwapLite` util functions

abcrane123:
**What changed? Why?**
- `useResetSwapLiteInputs` : resets to token, from token, USDC, and ETH balances 
- `getSwapLiteQuote`: formats swap quote values
- `useSwapLiteTokens`: returns token information (balance, etc)  - for context, USDC and ETH will always be provided as swap options. users then have the ability to provide one more `from` token to the component

**Notes to reviewers**

**How has it been tested?**


src/core/api/getSwapLiteQuote.ts
```diff
@@ -0,0 +1,71 @@
...
+type GetSwapLiteQuoteParams = Omit<GetSwapQuoteParams, 'from'> & {
+  fromSwapUnit: SwapUnit;
+  from?: Token;
+};
+
+/**
+ * Fetches a quote for a swap, but only if the from and to tokens are different.
+ */
+
+export async function getSwapLiteQuote({
```
dschlabach, line 27: Why not just check if to === from and then return early if that's the case?

abcrane123: we're fetching three quotes (to <> ETH, to <> USDC, to <> customToken) so we wouldn't want to return early. we want to gracefully handle case where any of those tokens overlap.

src/core/api/getSwapLiteQuote.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  let formattedFromAmount = '';
+  if (response && !isSwapError(response)) {
+    formattedFromAmount = response?.fromAmount
+      ? formatTokenAmount(response.fromAmount, response.from.decimals)
+      : '';
+
+    fromSwapUnit.setAmountUSD(response?.fromAmountUSD || '');
+    fromSwapUnit.setAmount(formattedFromAmount || '');
```
alessey, line 61: if this is going to be a core/api, it shouldn't setState, is there a reason a consumer would use this vs. getSwapQuote?

src/core/api/getSwapLiteQuote.ts
abcrane123: you can checkout the implementation of this hook [here](https://github.com/coinbase/onchainkit/pull/1729/files#diff-5fdf01a2dda1a451f5038f451d0d90a5ccbd38f2e31835a0470e631049811cc6R273)

src/core/api/getSwapLiteQuote.ts
abcrane123: oh yeah good catch let me move it to `swap`. no reason to have this here.

src/swap/hooks/useSwapLiteTokens.ts
```diff
@@ -0,0 +1,104 @@
...
+    setAmountUSD: setToAmountUSD,
+    setAmount: setToAmount,
+    token: toToken,
+    loading: toLoading,
+    setLoading: setToLoading,
+    error: toTokenBalanceError,
+  });
+
+  return { fromETH, from, fromUSDC, to };
+};
```
alessey, line 104: is it possible to extract a single useToken hook and use it for each usecase?

src/swap/constants.ts
```diff
@@ -23,3 +26,23 @@ export enum SwapMessage {
...
+
+export const usdcToken: Token = {
+  name: 'USDC',
+  address: '0x833589fcd6edb6e08f4c7c32d4f71b54bda02913',
+  symbol: 'USDC',
+  decimals: 6,
+  image:
+    'https://d3r81g40ycuhqg.cloudfront.net/wallet/wais/44/2b/442b80bd16af0c0d9b22e03a16753823fe826e5bfd457292b55fa0ba8c1ba213-ZWUzYjJmZGUtMDYxNy00NDcyLTg0NjQtMWI4OGEwYjBiODE2',
+  chainId: base.id,
+};
```
0xAlec, line 48: should these belong in https://github.com/coinbase/onchainkit/tree/main/src/token instead?

src/swap/hooks/useSwapLiteTokens.ts
0xAlec: +1, I think we should use a shared util for each input

src/swap/constants.ts
abcrane123: yeah i'll move them

src/swap/hooks/useSwapLiteTokens.ts
abcrane123: done ! let me know what you think

src/swap/hooks/useResetSwapLiteInputs.ts
```diff
@@ -0,0 +1,27 @@
...
+      fromETH.setAmountUSD(''),
+      fromUSDC.balanceResponse?.refetch(),
+      fromUSDC.setAmount(''),
+      fromUSDC.setAmountUSD(''),
+      to.balanceResponse?.refetch(),
+      to.setAmount(''),
+      to.setAmountUSD(''),
+    ]);
+  }, [from, fromETH, fromUSDC, to]);
+};
```
alessey, line 27: this may be more flexible accepting an array of SwapUnits

alessey (approved):

src/swap/hooks/useResetSwapLiteInputs.ts
abcrane123: created a ticket to refactor later

PR #1659, opened 2024-11-22
Css addresss Resolved

harshit-2004:
What changed? Why?

 The import statement for the CSS file was updated from import '../src/styles.css'; to import '../src/styles/index.css';. 

Notes to reviewers
Check if any other parts of the application rely on the old path and may need updates.

How has it been tested?
 The change has been tested by running the application and confirming that the styles are being applied correctly.


dschlabach:
We aren't using storybook at the moment!

PR #1666, opened 2024-11-26
dev: fix playground dev server

dschlabach:
**What changed? Why?**
Dev server hot reloading has stopped working reliably; switching to bun linking and moving output to ./onchainkit in the playground dir to fix

**Notes to reviewers**
This only affects internal dev experience.

**How has it been tested?**
Locally

### DIFF
--- a/playground/nextjs-app-router/onchainkit/package.json
+++ b/playground/nextjs-app-router/onchainkit/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/onchainkit",
-  "version": "0.35.5",
+  "version": "0.35.7",
   "type": "module",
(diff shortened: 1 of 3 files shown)


playground/nextjs-app-router/onchainkit/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@coinbase/onchainkit",
-  "version": "0.35.5",
+  "version": "0.35.7",
```
cpcramer, line 3: nit but we can bump to 0.35.8

cpcramer (approved):

brendan-defi (approved):

PR #1667, opened 2024-11-26
fix: fix `TokenDropdown ` when button parent container is larger than button

dschlabach:
**What changed? Why?**
TokenDropdown popover appears to the right when the parent is larger than the button.

**Notes to reviewers**

**How has it been tested?**


cpcramer:
I think this is worthy of adding a changeset!

cpcramer (approved):

brendan-defi (approved):

PR #1688, opened 2024-12-03
chore: Improve <Socials /> Farcaster URL processing 

cpcramer:
**What changed? Why?**
This bug comes from when setting your Social URL in [base.org](http://base.org/), there is no enforcement on how to enter your profile. One user could have their entire Farcaster url hardcoded where another user could only have his Farcaster name.

I've updated how we process the Farcaster URLs to be more flexible. 

image

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/identity/utils/getSocialPlatformDetails.tsx
+++ b/src/identity/utils/getSocialPlatformDetails.tsx
@@ -19,7 +19,10 @@ export const PLATFORM_CONFIG: Record<
     icon: githubSvg,
   },
   farcaster: {
-    href: (value) => value,
+    href: (value) => {
+      const username = value.split('/').pop();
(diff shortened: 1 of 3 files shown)


alessey (approved):

src/identity/utils/getSocialPlatformDetails.tsx
```diff
@@ -19,7 +19,10 @@ export const PLATFORM_CONFIG: Record<
     icon: githubSvg,
   },
   farcaster: {
-    href: (value) => value,
+    href: (value) => {
+      const username = value.split('/').pop();
```
0xAlec, line 23: is `/` a valid character for a Farcaster handle?

0xAlec (approved):

PR #1697, opened 2024-12-05
chore(Onramp-4734): Implement utility functions to call onramp API from OCK

rustam-cb:
**What changed? Why?**
Implemented util functions that will be used to call onramp APIs using client API keys 

Example of usage: 

```
setOnchainKitConfig({apiKey: 'YOUR_CLIENT_API_KEY_HERE'})
 fetchOnrampConfig()
```

```
setOnchainKitConfig({apiKey: 'YOUR_CLIENT_API_KEY_HERE'})
 fetchOnrampOptions({ country: 'US', subdivision: 'CA' })
```

**Notes to reviewers**

**How has it been tested?**


src/internal/utils/convertSnakeToCamelCase.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  if (Array.isArray(obj)) {
+    return obj.map((item) => convertSnakeToCamelCase(item)) as T;
+  }
+
+  if (obj && obj.constructor === Object) {
+    return Object.keys(obj).reduce((acc, key) => {
+      const camelCaseKey = toCamelCase(key);
+      // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+      (acc as any)[camelCaseKey] = convertSnakeToCamelCase<T>(
```
alessey, line 19: do you even need to cast these?  can't ts infer that acc is still T and obj is a property of T?

src/internal/utils/convertSnakeToCamelCase.ts
```diff
@@ -0,0 +1,31 @@
...
+      );
+      return acc;
+    }, {} as T);
+  }
+  return obj;
+}
+
+function toCamelCase(str: string): string {
+  return str.replace(/_([a-z])/g, (_, letter) => letter.toUpperCase());
+}
```
dschlabach, line 31: Nice! Only comment is that if we do it this way, we can remove the `any` and also have nice types outputted from the result!

So you'll get this:
const input = { snake_case: 1 }
const output = convertSnakeToCamelCase(input)

`output` will be of type: `{ snakeCase: 1 }`

```suggestion
/**
 * Converts snake_case keys to camelCase in an object or array of objects.
 * @param {T} obj - The object, array, or string to convert. (required)
 * @returns {T} The converted object, array, or string.
 */
export function convertSnakeToCamelCase<T>(obj: T): T {
  if (typeof obj === 'string') {
    return obj as T;
  }

  if (Array.isArray(obj)) {
    return obj.map((item) => convertSnakeToCamelCase(item)) as T;
  }

  if (obj && obj.constructor === Object) {
    return Object.keys(obj).reduce((acc, key) => {
      const camelCaseKey = toCamelCase(key);
      // biome-ignore lint/suspicious/noExplicitAny: <explanation>
      (acc as any)[camelCaseKey] = convertSnakeToCamelCase<T>(
        // biome-ignore lint/suspicious/noExplicitAny: <explanation>
        (obj as any)[key],
      );
      return acc;
    }, {} as T);
  }
  return obj;
}

function toCamelCase(str: string): string {
  return str.replace(/_([a-z])/g, (_, letter) => letter.toUpperCase());
}
```

src/internal/utils/convertSnakeToCamelCase.ts
rustam-cb: Done

src/internal/utils/convertSnakeToCamelCase.ts
rustam-cb: Done

src/fund/utils/fetchOnrampQuote.ts
```diff
@@ -0,0 +1,90 @@
...
+import { getApiKey } from '../../internal/utils/getApiKey';
+import { ONRAMP_API_BASE_URL } from '../constants';
+import type { OnrampAmout } from '../types';
+
+type OnrampQuoteResponseData = {
+  /**
+   * Object with amount and currency of the total fiat payment required to complete the purchase, inclusive of any fees.
+   * The currency will match the `paymentCurrency` in the request if it is supported, otherwise it falls back to USD.
+   */
+  paymenTotal: OnrampAmout;
```
abcrane123, line 11: typo ? `paymentTotal`

abcrane123: also `OnrampAmount`

src/fund/utils/fetchOnrampQuote.ts
```diff
@@ -0,0 +1,90 @@
+import { convertSnakeToCamelCase } from '../../internal/utils/convertSnakeToCamelCase';
+import { getApiKey } from '../../internal/utils/getApiKey';
+import { ONRAMP_API_BASE_URL } from '../constants';
+import type { OnrampAmout } from '../types';
```
abcrane123, line 4: typo ? `OnrampAmount`

abcrane123 (approved):

dschlabach (approved):

PR #1718, opened 2024-12-11
Implement Fund Card component

rustam-cb:
**What changed? Why?**
Implemented a new component `FundCard`
P/PS: https://docs.google.com/document/d/1y57fxsga2ZLF8-ieH2OY7jvqx5468txQuek1WTYPBxE/edit?tab=t.0

Usage: 
```
FundCard assetSymbol="ETH" /> 
```

Component is customizable and there are few components that can be passed into it.

```
  AmountInputComponent 
  HeaderComponent
  AmountInputTypeSwithComponent
  PaymentMethodSelectorDropdownComponent
  SubmitButtonComponent
```

|Fiat input|Crypto input|
|----------|------------|
|Screenshot 2024-12-17 at 9 38 52 PM|Screenshot 2024-12-17 at 9 39 06 PM|

|No input (default) | payment method dropdown|
|-------------------|--------------------------|
|Screenshot 2024-12-17 at 9 38 41 PM|Screenshot 2024-12-17 at 9 39 20 PM|

**Notes to reviewers**
Overall component structure:
- FundCard
- - HeaderComponent
- - AmountInputComponent
- - PaymentMethodDropdownComponent
- - SubmitButtonComponent

**How has it been tested?**
- Unit tests
- Manual testing


playground/nextjs-app-router/components/demo/Swap.tsx
```diff
@@ -45,9 +45,9 @@ function SwapComponent() {
   };
 
   const usdcToken: Token = {
-    name: 'USDC',
+    name: 'USD',
     address: '0x833589fcd6edb6e08f4c7c32d4f71b54bda02913',
-    symbol: 'USDC',
+    symbol: 'USD',
     decimals: 6,
     image:
```
dschlabach, line 52: Should revert this or else our Swap component won't show USDC as needed:

rustam-cb: Been testing it and forgot to revert. Thank you for catching!

vitest.config.ts
```diff
@@ -24,6 +24,7 @@ export default defineConfig({
         'playground/**',
         'site/**',
         'create-onchain/**',
+        '**/**.test.tsx',
```
dschlabach, line 27: Intentional?

rustam-cb: Yes. The code coverage was listing test files and counting them towards coverage %.  I had couple of my test files noted as covered 99.97% which prevented me from creating PR. (Its likely due to helper objects, functions that you create in a test files).

src/fund/components/FundButton.tsx
```diff
@@ -19,34 +21,65 @@ export function FundButton({
...
+      default:
+        return pressable.primary;
+    }
+  }, [buttonState]);
+
   const classNames = cn(
     componentTheme,
-    pressable.primary,
+    buttonColorClass,
     'px-4 py-3 inline-flex items-center justify-center space-x-2 disabled',
```
dschlabach, line 83: Probably want to add hover:cursor-disabled here

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,197 @@
...
+            currencySign={currencySign}
+          />
+        )}
+
+        <input
+          className={cn(
+            componentTheme,
+            text.body,
+            'border-[none] bg-transparent',
+            'text-[60px] leading-none outline-none',
```
dschlabach, line 113: Can we make this rem based? Otherwise won't scale properly

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,197 @@
...
+      <div className="flex" style={{ height: '78px' }}>
+        {/* Display the fiat currency sign before the input*/}
+        {inputType === 'fiat' && currencySign && (
+          <FundCardCurrencyLabel
+            ref={currencySpanRef}
+            currencySign={currencySign}
+          />
+        )}
+
+        <input
```
dschlabach, line 108: It would be nice if the input spanned the entire width, otherwise if a user clicks to the right, they won't focus the input
CleanShot 2024-12-18 at 13 31 46@2x

src/fund/components/FundCardPaymentMethodDropdown.tsx
```diff
@@ -0,0 +1,95 @@
...
+import { useTheme } from '../../core-react/internal/hooks/useTheme';
+import type {
+  FundCardPaymentMethodDropdownPropsReact,
+  PaymentMethodReact,
+} from '../types';
+import { FundCardPaymentMethodSelectRow } from './FundCardPaymentMethodSelectRow';
+import { FundCardPaymentMethodSelectorToggle } from './FundCardPaymentMethodSelectorToggle';
+import { useFundContext } from './FundCardProvider';
+
+export function FundCardPaymentMethodDropdown({
```
dschlabach, line 13: Should be able to close on Esc and either close when user tabs out of the dropdown (or trap focus).

Sometimes I also get this flash

https://github.com/user-attachments/assets/5e3bb4a1-f542-4cff-a079-f23cd3b34746

src/fund/components/FundCardPaymentMethodDropdown.tsx
```diff
@@ -0,0 +1,95 @@
...
+          ref={dropdownRef}
+          data-testid="ockFundCardPaymentMethodDropdown"
+          className={cn(
+            componentTheme,
+            border.radius,
+            'absolute right-0 z-10 mt-1 flex max-h-80 w-full flex-col overflow-y-hidden',
+            'ock-scrollbar',
+          )}
+        >
+          <div className={cn(background.inverse, 'p-2', 'overflow-y-auto')}>
```
dschlabach, line 81:
```suggestion
          <div className={cn(background.inverse, 'p-2 overflow-y-auto')}>
```

src/fund/components/FundButton.tsx
dschlabach, line 83: is `disabled` a class in our codebase? I can't find it

src/fund/components/FundCardAmountInput.tsx
rustam-cb: Unfortunately currently it is working this way because if you switch to Crypto it will have a space [1 - - - - - - -ETH] but we want the asset name next to the digits [1 ETH].  
But you are right, we should make it better for users. I'll take a look into focusing on the input field when that area is clicked. (I still have a few features to add to this component, so I'll make this change in a follow up PR )

cpcramer:
Seeing a strange flicker when using the dropdown

Clipboard-20241218-192142-422

src/core-react/internal/hooks/useIcon.tsx
```diff
@@ -1,26 +1,38 @@
...
+  className,
+}: { icon?: React.ReactNode; className?: string }) => {
   return useMemo(() => {
     if (icon === undefined) {
       return null;
     }
     switch (icon) {
       case 'coinbasePay':
-        return coinbasePaySvg;
+        return <CoinbasePaySvg className={className} />;
```
cpcramer, line 20: nit: can we return the svg similar to the other cases? 
```suggestion
        return coinbasePaySvg;
```

src/fund/components/FundCard.tsx
```diff
@@ -0,0 +1,154 @@
...
+import type { FundCardPropsReact, PaymentMethodReact } from '../types';
+import { FundButton } from './FundButton';
+import FundCardAmountInput from './FundCardAmountInput';
+import FundCardAmountInputTypeSwitch from './FundCardAmountInputTypeSwitch';
+import { FundCardHeader } from './FundCardHeader';
+import { FundCardPaymentMethodDropdown } from './FundCardPaymentMethodDropdown';
+import { FundCardProvider } from './FundCardProvider';
+import { useFundContext } from './FundCardProvider';
+
+const defaultPaymentMethods: PaymentMethodReact[] = [
```
cpcramer, line 15:
```suggestion
const DEFAULT_PAYMENT_METHODS: PaymentMethodReact[] = [
```

cpcramer: Can we add this to a constants file?

src/fund/components/FundCard.tsx
```diff
@@ -0,0 +1,154 @@
...
+          paymentMethodDropdownComponent={paymentMethodDropdownComponent}
+          paymentMethods={paymentMethods}
+          submitButtonComponent={submitButtonComponent}
+        />
+      </div>
+    </FundCardProvider>
+  );
+}
+
+export function FundCardContent({
```
cpcramer, line 78: Does this need to be exported?

src/core-react/internal/hooks/useIcon.tsx
rustam-cb: Is it possible to use className on svg? The reason is that I need to pass a color to the svg.

src/fund/components/FundCard.tsx
rustam-cb: nope, it does not.

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,197 @@
...
+    </div>
+  );
+};
+
+export default FundCardAmountInput;
+
+/**
+ * Ensure the decimal value is formatted correctly (i.e. "0.1" instead of ".1" and "0.1" instead of "01")
+ */
+const formatDecimalInputValue = (value: string) => {
```
cpcramer, line 167: This component is getting pretty complex - I feel like we should try to export some functionality into utility functions.

src/fund/components/FundCardAmountInputTypeSwitch.tsx
```diff
@@ -0,0 +1,99 @@
...
+export const FundCardAmountInputTypeSwitch = ({
+  selectedInputType,
+  setSelectedInputType,
+  selectedAsset,
+  fundAmountFiat,
+  fundAmountCrypto,
+  exchangeRate,
+  isLoading,
+}: FundCardAmountInputTypeSwitchPropsReact) => {
+  const componentTheme = useTheme();
```
cpcramer, line 18: `componentTheme` should only have to be applied once to the top level component

src/fund/components/FundCardAmountInputTypeSwitch.tsx
```diff
@@ -0,0 +1,99 @@
...
+    setSelectedInputType(selectedInputType === 'fiat' ? 'crypto' : 'fiat');
+  };
+
+  const formatUSD = useCallback((amount: string) => {
+    if (!amount || amount === '0') {
+      return null;
+    }
+    const roundedAmount = Number(getRoundedAmount(amount, 2));
+    return `$${roundedAmount.toFixed(2)}`;
+  }, []);
```
cpcramer, line 32: I've seen this function a few times, we should a reusable one cc @dschlabach @alissacrane-cb

src/fund/components/FundCardAmountInputTypeSwitch.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  const exchangeRateLine = useMemo(() => {
+    return (
+      <span
+        className={cn(
+          text.label2,
+          color.foregroundMuted,
+          'font-normal',
```
cpcramer, line 40: Is this necessary? I think it defaults to font-normal?

src/fund/components/FundCardAmountInputTypeSwitch.tsx
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  return (
+    <div className="flex items-center">
+      <button
+        type="button"
+        aria-label="amount type switch"
+        className={cn(
+          pressable.default,
+          'mr-1 rounded-full p-1 opacity-50 transition-opacity hover:opacity-100',
```
cpcramer, line 76: Can use the `border.radius` class here

src/fund/components/FundCardAmountInputTypeSwitch.tsx
```diff
@@ -0,0 +1,99 @@
...
+    </div>
+  );
+};
+
+const textStyle = {
+  textOverflow: 'ellipsis',
+  width: '390px',
+  overflow: 'hidden',
+  whiteSpace: 'nowrap',
+};
```
cpcramer, line 97: Let's try to avoid inline styling in favor of Tailwind class based styling,

src/fund/components/FundCardCurrencyLabel.tsx
```diff
@@ -0,0 +1,26 @@
+import { forwardRef } from 'react';
+import { useTheme } from '../../core-react/internal/hooks/useTheme';
+import { cn, text } from '../../styles/theme';
+import type { FundCardCurrencyLabelPropsReact } from '../types';
+
+export const FundCardCurrencyLabel = forwardRef<
+  HTMLSpanElement,
+  FundCardCurrencyLabelPropsReact
+>(({ currencySign }, ref) => {
+  const componentTheme = useTheme();
```
cpcramer, line 10: Same as above but you should only need to pass this componentTheme down in the top level component

src/fund/components/FundCardPaymentMethodDropdown.tsx
```diff
@@ -0,0 +1,95 @@
...
+  }, []);
+
+  useEffect(() => {
+    // Add event listener for outside clicks
+    document.addEventListener('click', handleBlur);
+    return () => {
+      // Clean up the event listener
+      document.removeEventListener('click', handleBlur);
+    };
+  }, [handleBlur]);
```
cpcramer, line 60: We have `useOutsideClick` hook that should help with this

src/fund/components/FundCardPaymentMethodDropdown.tsx
```diff
@@ -0,0 +1,95 @@
...
+      />
+      {isOpen && (
+        <div
+          ref={dropdownRef}
+          data-testid="ockFundCardPaymentMethodDropdown"
+          className={cn(
+            componentTheme,
+            border.radius,
+            'absolute right-0 z-10 mt-1 flex max-h-80 w-full flex-col overflow-y-hidden',
+            'ock-scrollbar',
```
cpcramer, line 78: What's this?

src/fund/components/FundButton.tsx
rustam-cb: not sure. I did not add it but I'll remove it. Does not seem to have any effect/

src/fund/components/FundCardPaymentMethodImage.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }, [size]);
+
+  if (!iconSvg) {
+    return (
+      <div
+        className={cn('overflow-hidden rounded-full', className)}
+        data-testid="fundCardPaymentMethodImage__noImage"
+        style={styles.image}
+      >
+        <div style={styles.placeholderImage} />
```
cpcramer, line 42: I'd prefer to avoid inline styling but curious what the team thinks

src/fund/components/FundCardAmountInputTypeSwitch.tsx
rustam-cb: Removing it makes the text `bold`

src/fund/components/FundCardPaymentMethodSelectRow.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const componentTheme = useTheme();
+
+    return (
+      <button
+        data-testid="ockFundCardPaymentMethodSelectRow__button"
+        type="button"
+        className={cn(
+          componentTheme,
+          pressable.default,
+          'ock-border-radius flex w-full items-center justify-between px-2 py-1',
```
cpcramer, line 24: Use `border.radius` from theme.ts instead

src/internal/components/Skeleton.tsx
```diff
@@ -0,0 +1,19 @@
+import { background, cn } from '../../styles/theme';
+
+type SkeletonReact = {
+  className?: string;
+};
+
+export function Skeleton({ className }: SkeletonReact) {
```
cpcramer, line 7: What is this `Skeleton` component for?

src/internal/components/Skeleton.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <div
+      className={cn(
+        'animate-pulse bg-opacity-50',
+        background.alternate,
+        'rounded',
```
cpcramer, line 13: Would` border.*` work from theme.ts?

src/core-react/internal/hooks/useIcon.tsx
dschlabach: Btw you can use `fill="currentColor"` on the SVG and then set `text-blue-400` or whatever you want with Tailwind

src/fund/components/FundButton.tsx
dschlabach: Sorry, meant cursor-not-allowed https://tailwindcss.com/docs/cursor#setting-the-cursor-style

src/fund/components/FundButton.tsx
```diff
@@ -19,47 +21,114 @@ export function FundButton({
...
           height,
           width,
           target,
         });
+
+        if (!popupWindow) {
+          return null;
+        }
+
+        const interval = setInterval(() => {
```
alessey, line 57: possible to use this?  https://github.com/coinbase/onchainkit/blob/main/src/buy/hooks/usePopupMonitor.ts

src/fund/components/FundCard.tsx
```diff
@@ -0,0 +1,134 @@
...
+          headerText={headerText}
+          amountInputComponent={amountInputComponent}
+          headerComponent={headerComponent}
+          amountInputTypeSwithComponent={amountInputTypeSwithComponent}
+          paymentMethodDropdownComponent={paymentMethodDropdownComponent}
+          paymentMethods={paymentMethods}
+          submitButtonComponent={submitButtonComponent}
+        />
+      </div>
+    </FundCardProvider>
```
alessey, line 54: the rest of our components are set up to accept child components vs props

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+        // Calculate the crypto value based on the exchange rate
+        const calculatedCryptoValue = String(
+          Number(value) * Number(exchangeRate),
+        );
+
+        const resultCryptoValue = truncateDecimalPlaces(
+          calculatedCryptoValue,
+          8,
+        );
```
alessey, line 42: nit: you could just use toFixed on your calculatedCryptoValue before casting it to a string, below as well for fiat

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+            -webkit-appearance: none;
+            appearance: none;
+            margin: 0;
+          }
+
+          input[type="number"] {
+            -moz-appearance: textfield;
+          }
+        `}
+      </style>
```
alessey, line 104: could you move this to a tailwind utility instead of an inline style

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+  useEffect(() => {
+    if (hiddenSpanRef.current) {
+      const width = Math.max(42, hiddenSpanRef.current.offsetWidth);
+      const currencyWidth =
+        currencySpanRef.current?.getBoundingClientRect().width || 0;
+
+      // Set the input width based on the span width
+      if (inputRef.current) {
+        inputRef.current.style.width = `${width}px`;
+        inputRef.current.style.maxWidth = `${390 - currencyWidth}px`;
```
alessey, line 68: what are you trying to do here?    Something like this may be more useful: https://github.com/coinbase/onchainkit/pull/1224

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+        )}
+      </div>
+
+      {/* Hidden span for measuring text width 
+          Without this span the input field would not adjust its width based on the text width and would look like this:
+          [0.12--------Empty Space-------][ETH] - As you can see the currency symbol is far away from the inputed value
+
+          With this span we can measure the width of the text in the input field and set the width of the input field to match the text width
+          [0.12][ETH] - Now the currency symbol is displayed next to the input field
+      */}
```
alessey, line 145: can't the input just scale as needed with a max-width?

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+      <span
+        data-testid="ockHiddenSpan"
+        ref={hiddenSpanRef}
+        className={cn(
+          text.body,
+          'border-[none] bg-transparent',
```
alessey, line 151: border-none

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,167 @@
...
+          <FundCardCurrencyLabel
+            ref={currencySpanRef}
+            currencySign={currencySign}
+          />
+        )}
+
+        <input
+          className={cn(
+            text.body,
+            'border-[none] bg-transparent',
```
alessey, line 118: border-none

src/fund/components/FundCardAmountInputTypeSwitch.tsx
```diff
@@ -0,0 +1,84 @@
...
+    return (
+      <span className={cn(text.label1)}>
+        {Number(fundAmountCrypto).toFixed(8)} {selectedAsset}
+      </span>
+    );
+  }, [fundAmountCrypto, selectedAsset]);
+
+  const fiatAmountLine = useMemo(() => {
+    return <span className={cn(text.label1)}>{formatUSD(fundAmountFiat)}</span>;
+  }, [formatUSD, fundAmountFiat]);
```
alessey, line 56: nit: this would be better as a single useMemo that returned the correctAmount line based on type

src/fund/components/FundCardPaymentMethodDropdown.tsx
```diff
@@ -0,0 +1,92 @@
...
+      className="relative py-4"
+      ref={dropdownContainerRef}
+      data-testid="ockFundCardPaymentMethodDropdownContainer"
+      onKeyUp={handleEscKeyPress}
+    >
+      <FundCardPaymentMethodSelectorToggle
+        ref={buttonRef}
+        onClick={handleToggle}
+        isOpen={isOpen}
+        paymentMethod={selectedPaymentMethod || paymentMethods[0]}
```
alessey, line 67: why is this needed if you are doing this in the useEffect, or can you remove the useEffect and do this instead?

src/fund/components/FundCardPaymentMethodImage.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  return (
+    <div
+      data-testid="fundCardPaymentMethodImage__iconContainer"
+      className={cn(
+        'flex items-center justify-center overflow-hidden rounded-[50%]',
+        className,
+      )}
+      style={styles.image}
```
alessey, line 54: use tailwind

src/fund/components/FundCardPaymentMethodImage.tsx
```diff
@@ -0,0 +1,59 @@
...
+        height: `${size}px`,
+        minWidth: `${size}px`,
+        minHeight: `${size}px`,
+      },
+      placeholderImage: {
+        width: `${size}px`,
+        height: `${size}px`,
+        minWidth: `${size}px`,
+        minHeight: `${size}px`,
+      },
```
alessey, line 31: anything defined in px won't scale properly with browser zoom levels

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,171 @@
...
+            margin: 0;
+          }
+
+          input[type="number"] {
+            -moz-appearance: textfield;
+          }
+        `}
+      </style>
+
+      <div className="flex" style={{ height: '78px' }}>
```
dschlabach, line 110: Stick to tailwind classes for consistency

dschlabach: preferably `h-20` here

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,171 @@
...
+            ref={currencySpanRef}
+            currencySign={currencySign}
+          />
+        )}
+
+        <input
+          className={cn(
+            text.body,
+            'border-none bg-transparent',
+            'text-[3.75rem] leading-none outline-none',
```
dschlabach, line 123: should be `text-6xl` instead of arbitrary

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,171 @@
...
+          With this span we can measure the width of the text in the input field and set the width of the input field to match the text width
+          [0.12][ETH] - Now the currency symbol is displayed next to the input field
+      */}
+      <span
+        data-testid="ockHiddenSpan"
+        ref={hiddenSpanRef}
+        className={cn(
+          text.body,
+          'border-none bg-transparent',
+          'text-[3.75rem] leading-none outline-none',
```
dschlabach, line 156: `text-6xl`

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,171 @@
...
+        )}
+        style={{
+          position: 'absolute',
+          opacity: 0,
+          whiteSpace: 'nowrap',
+          pointerEvents: 'none',
+        }}
```
dschlabach, line 163: move into Tailwind utilities for consistency

src/internal/svg/successSvg.tsx
```diff
@@ -11,7 +13,7 @@ export const successSvg = (
     <title>Success SVG</title>
     <path
       d="M8 0C3.58 0 0 3.58 0 8C0 12.42 3.58 16 8 16C12.42 16 16 12.42 16 8C16 3.58 12.42 0 8 0ZM6.72667 11.5333L3.73333 8.54L4.67333 7.6L6.72667 9.65333L11.44 4.94L12.38 5.88L6.72667 11.5333Z"
-      fill="#65A30D"
+      className={className}
```
dschlabach, line 16: This will break color styling

src/internal/svg/errorSvg.tsx
```diff
@@ -11,7 +13,7 @@ export const errorSvg = (
     <title>Error SVG</title>
     <path
       d="M8 16C12.4183 16 16 12.4183 16 8C16 3.58171 12.4183 0 8 0C3.58172 0 0 3.58171 0 8C0 12.4183 3.58172 16 8 16ZM11.7576 5.0909L8.84853 8L11.7576 10.9091L10.9091 11.7576L8 8.84851L5.09093 11.7576L4.2424 10.9091L7.15147 8L4.2424 5.0909L5.09093 4.24239L8 7.15145L10.9091 4.24239L11.7576 5.0909Z"
-      fill="#E11D48"
+      className={className}
```
dschlabach, line 16: This will break color styling

src/internal/svg/coinbasePaySvg.tsx
```diff
@@ -12,7 +14,7 @@ export const coinbasePaySvg = (
       fillRule="evenodd"
       clipRule="evenodd"
       d="M10.0145 14.1666C7.82346 14.1666 6.04878 12.302 6.04878 9.99996C6.04878 7.69788 7.82346 5.83329 10.0145 5.83329C11.9776 5.83329 13.6069 7.33677 13.9208 9.30552H17.9163C17.5793 5.02774 14.172 1.66663 10.0145 1.66663C5.63568 1.66663 2.08301 5.39926 2.08301 9.99996C2.08301 14.6007 5.63568 18.3333 10.0145 18.3333C14.172 18.3333 17.5793 14.9722 17.9163 10.6944H13.9208C13.6069 12.6632 11.9776 14.1666 10.0145 14.1666Z"
-      fill="#f9fafb"
+      className={className}
```
dschlabach, line 17: Doesn't this break default styling in things like `CheckoutButton`?

src/internal/components/TextInput.tsx
```diff
@@ -49,6 +51,7 @@ export function TextInput({
 
   return (
     <input
+      style={style}
```
dschlabach, line 54: Why do we need to add an inline style attribute here? We should try to stick to `className` for consistency

src/fund/components/FundCardAmountInput.tsx
rustam-cb: @dschlabach  Where is the Tailwing utilities? I can't find any in the codebase. (The css I see is autogenerated and I can't add anything there)

src/core-react/internal/hooks/useIcon.tsx
dschlabach: I would like to have this standard so that when you call `useIcon` it behaves the same regardless of any icon that you pass in. 

I understand you want to change the color of this CoinbasePay icon, but we should have one standard interface for all the SVGs or else there are going to be unintentional cases of icons breaking

src/internal/svg/successSvg.tsx
rustam-cb: @dschlabach  It used to be hardcoded `65A30D`, and now I'm using `icon.success`. Do you mean that `icon.success` is a different tone than `65A30D` ?

src/fund/components/FundCard.tsx
```diff
@@ -0,0 +1,164 @@
...
+  headerComponent,
+  amountInputTypeSwithComponent,
+  paymentMethodDropdownComponent,
+  paymentMethods = DEFAULT_PAYMENT_METHODS,
+  submitButtonComponent,
+}: FundCardContentPropsReact) {
+  /**
+   * Fetches and sets the exchange rate for the asset
+   */
+  useExchangeRate(assetSymbol);
```
dschlabach, line 94: Should this be moved into the provider or is there a reason it's here? A bit hard to tell what side effect it's having

src/internal/svg/coinbasePaySvg.tsx
rustam-cb: I have set color `icon.foreground` by default

src/fund/components/FundCardAmountInput.tsx
cpcramer: All Tailwind utilities are accessible anywhere in /src. You can apply them directly as classes. https://tailwindcss.com/docs/utility-first

src/internal/svg/coinbasePaySvg.tsx
cpcramer: This will break the default styling for this SVG when themes are applied. If you need a CoinbasePaySVG to update then we might have to create a new SVG for that

src/internal/svg/successSvg.tsx
cpcramer: This is likely hardcoded because the icon fill color should not update based off of themes / modes

cpcramer:
I recommend splitting this into smaller PRs, it will make reviewing and updating a lot easier and faster!

rustam-cb:
I have addressed all the comments. I'm not moving forward with splitting the PR up since its already had a few iterations of the review and splitting would mean each PR would need to be reviewed again. Moving forward I'll make sure to not have such big PRs.

src/fund/components/FundCardAmountInput.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const defaultProps = {
+    fiatValue: '100',
+    setFiatValue: vi.fn(),
+    cryptoValue: '0.05',
+    setCryptoValue: vi.fn(),
+    currencySign: '$',
+    assetSymbol: 'ETH',
+    inputType: 'fiat',
+    exchangeRate: '2',
+  } as unknown as FundCardAmountInputPropsReact;
```
dschlabach, line 17: Is it possible to use type annotation instead of type assertion here? That will prevent us from forgetting to update this object in case `FundCardAmountInputPropsReact` changes in the future

e.g. `const defaultProps: FundCardAmountInputPropsReact`, or does that break because of the `vi.fn()`?

src/fund/components/FundCardHeader.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export function FundCardHeader({
+  headerText,
+  assetSymbol,
+}: FundCardHeaderPropsReact) {
+  const defaultHeaderText = `Buy ${assetSymbol.toUpperCase()}`;
+
+  return (
+    <div
+      className={cn('font-display text-[16px]', 'leading-none outline-none')}
```
dschlabach, line 12: `text-[16px]` should be `text-base` and no need to call `cn()` here, can be a single string

src/fund/components/FundCardCurrencyLabel.tsx
```diff
@@ -0,0 +1,22 @@
...
+  HTMLSpanElement,
+  FundCardCurrencyLabelPropsReact
+>(({ currencySign }, ref) => {
+  return (
+    <span
+      ref={ref}
+      className={cn(
+        text.body,
+        'flex items-center justify-center bg-transparent',
+        'text-[60px] leading-none outline-none',
```
dschlabach, line 15: `text-6xl` is 60px by default, but it's in rem so scales better if the user's default font size is different 

https://tailwindcss.com/docs/font-size

src/fund/components/FundCardAmountInput.test.tsx
dschlabach: @rustam-cb - does that not work or is there a reason you resolved this comment? We just made many of our tests typesafe and would like to keep that pattern going forward

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,171 @@
...
+      const currencyWidth =
+        currencySpanRef.current?.getBoundingClientRect().width || 0;
+
+      // Set the input width based on the span width
+      if (inputRef.current) {
+        inputRef.current.style.width = `${width}px`;
+        inputRef.current.style.maxWidth = `${390 - currencyWidth}px`;
+      }
+    }
+  }, [value]);
```
alessey, line 75: can you just use two inline elements here and let the browser determine the width?

cpcramer:
Buy ETH text is not updating on theme change: 
Screenshot 2025-01-06 at 2 06 38 PM

src/fund/components/FundButton.tsx
```diff
@@ -19,49 +22,121 @@ export function FundButton({
...
+
+  const buttonContent = useMemo(() => {
+    if (buttonState === 'loading') {
+      return <Spinner />;
+    }
+
+    return (
+      <>
+        {buttonIcon && (
+          // h-6 is to match the icon height to the line-height set by text.headline
```
cpcramer, line 124: Nit: can probably remove this comment

rustam-cb: @cpcramer  Done

src/fund/components/FundCardAmountInput.tsx
rustam-cb: Can't do this. 
The reason I have this is because I have 2 cases: 
First:
[$ 50            ]

Second: 
[50 ETH        ]

For the second case I need the ETH to be next to the text. All other solutions I tried makes the ETH to be at the end of the input like so: 
[50         ETH]

src/styles/theme.ts
```diff
@@ -26,6 +26,7 @@ export const pressable = {
     'cursor-pointer ock-bg-inverse active:bg-[var(--ock-bg-inverse-active)] hover:bg-[var(--ock-bg-inverse-hover)]',
   primary:
     'cursor-pointer ock-bg-primary active:bg-[var(--ock-bg-primary-active)] hover:bg-[var(--ock-bg-primary-hover)]',
+  error: 'cursor-pointer ock-bg-error',
```
cpcramer, line 29: Curious why the error needs cursor-pointer? 

Why not just use the existing background.error?

rustam-cb: Good point. Missed `background.error`. Updated now.

src/fund/components/FundCardAmountInput.tsx
alessey: inline elements have a width equal to their contents.  instead of using a flex box to align block level elements, just use inline and let the browser do the work, https://www.w3schools.com/html/html_blocks.asp

rustam-cb: @alessey 
- I have an inline element which is hidden and I have input element. 
- Input does not resize based on its content.  
- When you enter text into the input it also gets added to the inline element and then I take the inline element width and set it to the input. 

Could you please help me understand the solution with two inline elements that you are suggesting?

alessey: since there are no borders or anything, you can  have your input and "labels" on different layers, grid, etc.  you can use your hidden span to be the spacer for the currency without actually resizing the input using js, but that may be less flexible if we want to reuse since it depends on styling.  This solution is fine, ideally we'd extract to an `<AutoresizeInput/>` component.

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,152 @@
...
+      */}
+      <span
+        data-testid="ockHiddenSpan"
+        ref={hiddenSpanRef}
+        className={cn(
+          text.body,
+          'border-none bg-transparent',
+          'text-6xl leading-none outline-none',
+          'pointer-events-none absolute whitespace-nowrap opacity-0',
+        )}
```
alessey, line 144: with a long enough input, this span will trigger scrollbars on the page

src/fund/components/FundButton.tsx
```diff
@@ -19,49 +29,120 @@ export function FundButton({
...
+    if (hideIcon) {
+      return null;
+    }
+    switch (buttonState) {
+      case 'loading':
+        return '';
+      case 'success':
+        return <SuccessSvg fill="#F9FAFB" />;
+      case 'error':
+        return <ErrorSvg fill="#F9FAFB" />;
```
alessey, line 104: should these use a theme variable?

src/fund/components/FundCard.tsx
alessey: any reason not to do the same here?

src/fund/components/FundCard.tsx
```diff
@@ -0,0 +1,157 @@
...
+          hideIcon={submitButtonState === 'default'}
+          text={buttonText}
+          className="w-full"
+          fundingUrl={fundingUrl}
+          state={submitButtonState}
+          onClick={() => setSubmitButtonState('loading')}
+          onPopupClose={() => setSubmitButtonState('default')}
+        />
+      )}
+    </form>
```
alessey, line 155: I'm confused by this component design.  Its set up so you can pass in subcomponents, i.e. FundCardAmountInput, etc. but you can't really because of all the provider props they require.   How is this supposed to work?

rustam-cb: This does not work. It did initially work the way I implemented it but I changed it to follow the other pattern in onchainkit and now it does not. I'll change it back. 
https://github.com/coinbase/onchainkit/pull/1718#discussion_r1892923451

src/fund/components/FundButton.tsx
rustam-cb: No, we don't need the color to change based on the theme.

src/fund/components/FundCardAmountInput.tsx
rustam-cb: Fixed

src/fund/components/FundCard.tsx
rustam-cb: Changed the customizable components to be passed as props

src/fund/components/FundCard.tsx
```diff
@@ -0,0 +1,149 @@
...
+          buttonText={buttonText}
+          headerText={headerText}
+          paymentMethods={paymentMethods}
+          AmountInputComponent={AmountInputComponent}
+          HeaderComponent={HeaderComponent}
+          AmountInputTypeSwitchComponent={AmountInputTypeSwitchComponent}
+          PaymentMethodDropdownComponent={PaymentMethodDropdownComponent}
+          ButtonComponent={ButtonComponent}
+        />
+      </div>
```
alessey, line 75: if you pass in these components as children (and have a default children vs. default props), and rewrite the components to use the values from the providers instead of passing them from the provider as props, you give users the ability to customize layout and with className props more granularity over styling.  You do not really need the children loop, should be able to just pass them through.

src/fund/components/FundCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { FundCardProvider, useFundContext } from './FundCardProvider';
+
+export function FundCard({
+  assetSymbol,
+  buttonText = 'Buy',
+  headerText,
+  currencySign = '$',
+  paymentMethods = DEFAULT_PAYMENT_METHODS,
+  children,
+}: FundCardPropsReact) {
```
alessey, line 21: can we add a className prop to this and the sub components to allow consumers to customize?

src/fund/components/FundCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+      {children || (
+        <>
+          <FundCardHeader />
+
+          <FundCardAmountInput />
+
+          <FundCardAmountInputTypeSwitch />
+
+          <FundCardPaymentMethodDropdown />
+        </>
```
alessey, line 74: this could be a prop default

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,160 @@
...
+          Number(truncatedValue) / Number(exchangeRate),
+        );
+        const resultFiatValue = truncateDecimalPlaces(calculatedFiatValue, 2);
+        setFundAmountFiat(resultFiatValue === '0' ? '' : resultFiatValue);
+      }
+    },
+    [exchangeRate, setFundAmountFiat, setFundAmountCrypto, selectedInputType],
+  );
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: When value changes, we want to update the input width
```
alessey, line 68: nit: you could explicitly resize when value changes vs. listening for value updates

src/internal/svg/applePaySvg.tsx
```diff
@@ -0,0 +1,23 @@
...
+      className={icon.foreground}
+    />
+    <path
+      d="M13.5228 11.7446C13.5228 11.7558 13.5228 11.7657 13.5228 11.7762C13.2031 12.7444 12.7471 13.5741 12.1907 14.3442C11.6827 15.0432 11.0602 15.984 9.9487 15.984C8.98826 15.984 8.35031 15.3664 7.36598 15.3496C6.32475 15.3327 5.75214 15.866 4.80013 16.0002C4.69123 16.0002 4.58233 16.0002 4.47554 16.0002C3.77646 15.899 3.21228 15.3454 2.80127 14.8465C1.58931 13.3725 0.652757 11.4685 0.478516 9.0319C0.478516 8.79302 0.478516 8.55485 0.478516 8.31597C0.552287 6.57215 1.39961 5.15432 2.52586 4.46719C3.12025 4.10185 3.93735 3.7906 4.84721 3.92971C5.23714 3.99014 5.63551 4.12363 5.98469 4.25571C6.31561 4.38288 6.72944 4.60841 7.12148 4.59647C7.38706 4.58874 7.65123 4.45033 7.91892 4.35267C8.70301 4.06953 9.47164 3.74493 10.4848 3.89739C11.7024 4.08147 12.5665 4.62246 13.1005 5.45714C12.0705 6.11265 11.2562 7.10049 11.3953 8.7874C11.519 10.3197 12.4099 11.2162 13.5228 11.7446Z"
+      className={icon.foreground}
+    />
+  </svg>
+);
```
alessey, line 23: is this different then the appleSvg in the codebase?

src/fund/utils/formatDecimalInputValue.ts
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  // Add a leading zero if the value starts with a zero and is not a decimal. (i.e. "01" -> "0.1")
+  if (
+    resultValue.length === 2 &&
+    resultValue[0] === '0' &&
+    resultValue[1] !== '.'
+  ) {
+    resultValue = `${resultValue[0]}.${resultValue[1]}`;
+  }
```
alessey, line 18: this seems like slightly unintuitive behavior, might be better to replace 01 with 1 and just not allow 00

src/fund/components/FundCard.tsx
rustam-cb: Added

src/fund/components/FundCard.tsx
rustam-cb: Done

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,160 @@
...
+  useEffect(() => {
+    if (hiddenSpanRef.current) {
+      const width = Math.max(42, hiddenSpanRef.current.offsetWidth);
+      const currencyWidth =
+        currencySpanRef.current?.getBoundingClientRect().width || 0;
+
+      // Set the input width based on the span width
+      if (inputRef.current) {
+        inputRef.current.style.width = `${width}px`;
+        inputRef.current.style.maxWidth = `${390 - currencyWidth}px`;
```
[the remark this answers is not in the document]

src/internal/svg/applePaySvg.tsx
rustam-cb: Yes, those are different icons. apple icon vs applePay icon

src/fund/utils/formatDecimalInputValue.ts
rustam-cb: I feel like when user enters 01 they meant to have 0.1  and when they enter 00 they meant to enter 0.0 

I'll have this tested by the UX and will make any followup adjustments

src/fund/components/FundCardPaymentMethodDropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+          data-testid="ockFundCardPaymentMethodDropdown"
+          className={cn(
+            border.radius,
+            'ock-scrollbar absolute z-10 flex w-full flex-col overflow-y-hidden',
+          )}
+        >
+          <div className={cn(background.inverse, 'overflow-y-auto p-2')}>
+            {paymentMethods.map((paymentMethod) => (
+              <FundCardPaymentMethodSelectRow
+                className={cn(background.inverse, 'px-4 py-2')}
```
abcrane123, line 77: I might've missed it but are you disabling debit card and apple pay if the amount is less than $5? the onramp team recently added this check so the onramp flow will error out if the amount is less than $5.

src/fund/components/FundCardPaymentMethodDropdown.tsx
```diff
@@ -0,0 +1,88 @@
...
+        <div
+          ref={dropdownRef}
+          data-testid="ockFundCardPaymentMethodDropdown"
+          className={cn(
+            border.radius,
+            'ock-scrollbar absolute z-10 flex w-full flex-col overflow-y-hidden',
+          )}
+        >
+          <div className={cn(background.inverse, 'overflow-y-auto p-2')}>
+            {paymentMethods.map((paymentMethod) => (
```
abcrane123, line 75: are you hiding apple pay on non safari / mobile devices?

src/fund/components/FundCardPaymentMethodDropdown.tsx
rustam-cb: Addressed

src/fund/components/FundCardPaymentMethodDropdown.tsx
rustam-cb: Addressed

abcrane123 (dismissed):

abcrane123:
one last thing i forgot to ask about is the disabling of fields if < $5. You might want to check with design but for the Buy component we display a message that states minimum amount is $5 which i think is better than just hiding the row which could be confusing.

Screenshot 2025-01-09 at 9 50 57 AM

abcrane123:
i think the buy button is disabled here because i haven't connected wallet. you can probably do the same thing I did in the [Buy component ](https://github.com/coinbase/onchainkit/pull/1729/files#diff-25e2adaaccba43de700647b3d299088473ebba994e441797f2f4a8245c09415fR78)(replace buy button with connect wallet button if not connected) 

Screenshot 2025-01-09 at 9 55 46 AM

alessey:
can we add limits that correspond to the onramp limits?  i.e. 
image

alessey:
is this shadow under the dropdown correct?  
image

alessey:
debit card says up to 500/week (but doesn't disable when i input > 500), but when it goes to onramp, it tells me up to 750.

alessey:
spacing between value and currency changes when you max out the input, may need to adjust the max width
image

abcrane123:
To follow the same pattern as our other components, can you use the `text` prop on `ConnectWallet` button to replace text with `Buy`

Screenshot 2025-01-10 at 7 16 33 AM

src/fund/components/FundButton.tsx
```diff
@@ -19,49 +30,121 @@ export function FundButton({
...
+  onPopupClose,
+  onClick,
 }: FundButtonReact) {
   const componentTheme = useTheme();
   // If the fundingUrl prop is undefined, fallback to our recommended funding URL based on the wallet type
-  const fundingUrlToRender = fundingUrl ?? useGetFundingUrl();
+  const fallbackFundingUrl = useGetFundingUrl();
+  const fundingUrlToRender = fundingUrl ?? fallbackFundingUrl;
   const isDisabled = disabled || !fundingUrlToRender;
+  const shouldShowConnectWallet = !fundingUrlToRender;
```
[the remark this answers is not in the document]
abcrane123: seems like this also will render it if `projectId` is undefined

abcrane123:
i would also recommend posting the playground link in #proj-onchainkit-code-review so that the team / Tina / Mind can QA it for you before merging

src/fund/components/FundButton.tsx
```diff
@@ -19,49 +30,121 @@ export function FundButton({
...
+          </span>
+        )}
+        {hideText || (
+          <span data-testid="ockFundButtonTextContent">
+            {buttonTextContent}
+          </span>
+        )}
+      </>
+    );
+  }, [buttonState, buttonIcon, buttonTextContent, hideText]);
```
alessey, line 147: nit: this might be more readable extracting one layer further so you can just have a single buttonState switch statement to return spinner/success/error/default

src/fund/components/FundCard.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const defaultChildren = useMemo(
+    () => (
+      <>
+        <FundCardHeader />
+        <FundCardAmountInput />
+        <FundCardAmountInputTypeSwitch />
+        <FundCardPaymentMethodDropdown />
+      </>
+    ),
```
alessey, line 34: nit: could make these an actual default prop

src/fund/hooks/useFundCardSetupOnrampEventListeners.ts
```diff
@@ -0,0 +1,38 @@
...
+          setSubmitButtonState('default');
+        }, FUND_BUTTON_RESET_TIMEOUT);
+      },
+    });
+
+    return () => {
+      unsubscribe();
+    };
+  }, []);
+};
```
alessey, line 38: nit: could make this more loosely coupled by calling a callback instead of tying directly to the fund context

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,156 @@
...
+    [exchangeRate, setFundAmountFiat, setFundAmountCrypto, selectedInputType],
+  );
+
+  // Update width when value changes
+  // biome-ignore lint/correctness/useExhaustiveDependencies: <explanation>
+  useEffect(() => {
+    updateInputWidth();
+  }, [value, updateInputWidth]);
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: We only want to focus the input when the input type changes
```
alessey, line 78: would adding handleFocusInput to the deps and wrapping that in useCallback change anything?

src/fund/components/FundCardProvider.tsx
```diff
@@ -0,0 +1,118 @@
...
+export function FundCardProvider({
+  children,
+  asset,
+  paymentMethods,
+  headerText,
+  buttonText,
+  currencySign,
+  inputType,
+}: FundCardProviderReact) {
+  const [selectedAsset, setSelectedAsset] = useState<string>(asset);
```
alessey, line 47: does this need to be in state?  it only changes is solely based off the prop right?

src/fund/components/FundCardPaymentMethodDropdown.tsx
```diff
@@ -0,0 +1,140 @@
...
+      return (
+        (method.id === 'APPLE_PAY' || method.id === 'ACH_BANK_ACCOUNT') &&
+        amount < MIN_AMOUNT_FOR_CARD_PAYMENTS
+      );
+    },
+    [fundAmountFiat],
+  );
+
+  // If current selected method becomes disabled, switch to Coinbase
+  // biome-ignore lint/correctness/useExhaustiveDependencies: <explanation>
```
alessey, line 45: why is this ignore needed?

src/fund/components/FundCardPaymentMethodDropdown.tsx
```diff
@@ -0,0 +1,140 @@
...
+        >
+          <div className={cn(background.inverse, 'overflow-y-auto p-2')}>
+            {filteredPaymentMethods.map((paymentMethod) => (
+              <FundCardPaymentMethodSelectRow
+                className={cn(background.inverse, 'px-4 py-2')}
+                key={paymentMethod.name}
+                testId={`ockFundCardPaymentMethodSelectRow__${paymentMethod.id}`}
+                paymentMethod={paymentMethod}
+                onClick={handlePaymentMethodSelect}
+                disabled={isPaymentMethodDisabled(paymentMethod)}
```
alessey, line 127: nit: since you are already calling isPaymentMethodDisabled you could conditionally apply the onClick using that as well

src/fund/components/FundCardPaymentMethodSelectorToggle.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <span
+          className={cn(text.headline, color.foreground, 'flex w-full')}
+          data-testid="ockFundCardPaymentMethodSelectorToggle__paymentMethodName"
+        >
+          {paymentMethod.name}
+        </span>
+
+        <div className="relative flex items-center justify-center">
+          <div className="absolute top-0 left-0 h-4 w-4" />
+          {isOpen ? caretUpSvg : caretDownSvg}
```
alessey, line 48: nit: would be a little nicer to use the same svg and rotate it with an animation

src/fund/components/FundCardProvider.tsx
```diff
@@ -0,0 +1,118 @@
...
+    });
+
+    setExchangeRateLoading(false);
+
+    setExchangeRate(
+      Number(quote.purchaseAmount.value) / Number(quote.paymentSubtotal.value),
+    );
+  }, 1000);
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: One time effect
```
alessey, line 80: if you only call fetchExchangeRate once, why is it debounced?

src/fund/components/FundCardProvider.tsx
```diff
@@ -0,0 +1,118 @@
...
+    setFundAmountCrypto,
+    selectedInputType,
+    setSelectedInputType,
+    exchangeRate,
+    setExchangeRate,
+    exchangeRateLoading,
+    setExchangeRateLoading,
+    submitButtonState,
+    setSubmitButtonState,
+    paymentMethods: paymentMethods || DEFAULT_PAYMENT_METHODS,
```
alessey, line 102: how would this be used?  A user could restrict the payment methods based on DEFAULT_PAYMENT_METHODS?  this code seems pretty tied to the defaults, wouldn't other functionality like the payment select break if there was no Coinbase method?

src/fund/utils/truncateDecimalPlaces.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Limit the value to N decimal places
+ */
+export const truncateDecimalPlaces = (value: string, decimalPlaces: number) => {
+  const decimalIndex = value.indexOf('.');
```
alessey, line 5: nit: could easily make this more flexible by allowing value to be string | number

src/internal/svg/successSvg.tsx
```diff
@@ -11,7 +11,7 @@ export const successSvg = (
     <title>Success SVG</title>
     <path
       d="M8 0C3.58 0 0 3.58 0 8C0 12.42 3.58 16 8 16C12.42 16 16 12.42 16 8C16 3.58 12.42 0 8 0ZM6.72667 11.5333L3.73333 8.54L4.67333 7.6L6.72667 9.65333L11.44 4.94L12.38 5.88L6.72667 11.5333Z"
-      fill="#65A30D"
+      fill={fill}
```
alessey, line 14: nit: do the success and error icons really need to be different colors between ock components?

src/fund/components/FundCard.tsx
```diff
@@ -0,0 +1,94 @@
...
+import FundCardAmountInputTypeSwitch from './FundCardAmountInputTypeSwitch';
+import { FundCardHeader } from './FundCardHeader';
+import { FundCardPaymentMethodDropdown } from './FundCardPaymentMethodDropdown';
+import { FundCardProvider, useFundContext } from './FundCardProvider';
+
+export function FundCard({
+  assetSymbol,
+  buttonText = 'Buy',
+  headerText,
+  currencySign = '$',
```
alessey, line 19: do we support any other currencies?

src/fund/components/FundCard.tsx
```diff
@@ -0,0 +1,94 @@
...
+      <FundButton
+        disabled={!fundAmountFiat && !fundAmountCrypto}
+        hideIcon={submitButtonState === 'default'}
+        text={buttonText}
+        className="w-full"
+        fundingUrl={fundingUrl}
+        state={submitButtonState}
+        onClick={() => setSubmitButtonState('loading')}
+        onPopupClose={() => setSubmitButtonState('default')}
+      />
```
alessey, line 91: shouldn't the FundButton be set up the same as the other children?

src/fund/components/FundCard.tsx
```diff
@@ -0,0 +1,94 @@
...
+      headerText={headerText}
+      buttonText={buttonText}
+      currencySign={currencySign}
+    >
+      <div
+        className={cn(
+          componentTheme,
+          background.default,
+          color.foreground,
+          'flex w-[440px] flex-col p-6',
```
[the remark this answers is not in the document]
alessey: isn't 440px to large for most mobile devices?

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,156 @@
...
+            '[appearance:textfield]',
+            '[&::-webkit-inner-spin-button]:m-0 [&::-webkit-inner-spin-button]:appearance-none',
+            '[&::-webkit-outer-spin-button]:m-0 [&::-webkit-outer-spin-button]:appearance-none',
+          )}
+          type="number"
+          value={value}
+          onChange={handleChange}
+          ref={inputRef}
+          inputMode="decimal"
+          minLength={1}
```
alessey, line 120: minLength is not a valid attribute for type="number"

src/fund/components/FundCardPaymentMethodDropdown.tsx
```diff
@@ -0,0 +1,140 @@
...
+                disabledReason={
+                  isPaymentMethodDisabled(paymentMethod)
```
alessey, line 129: extract isPaymentMethodDisabled to a const in this map

src/fund/components/FundCardPaymentMethodDropdown.tsx
```diff
@@ -0,0 +1,140 @@
...
+          data-testid="ockFundCardPaymentMethodDropdown"
+          className={cn(
+            border.radius,
+            'ock-scrollbar absolute z-10 flex w-full flex-col overflow-y-hidden',
+          )}
+        >
+          <div className={cn(background.inverse, 'overflow-y-auto p-2')}>
+            {filteredPaymentMethods.map((paymentMethod) => (
+              <FundCardPaymentMethodSelectRow
+                className={cn(background.inverse, 'px-4 py-2')}
```
alessey, line 122: any reason not to just apply these styles in the FundCardPaymentMethodSelectRow so they are all in once place?  it appears you are overriding styles set in that component even though its only used here

alessey:
this component should implement a lifecycle status to provide onStatus, onSuccess, onError callbacks: https://onchainkit.xyz/guides/lifecycle-status

src/internal/svg/successSvg.tsx
rustam-cb: These icons are used in other components with the color that is different than what I need.

src/fund/hooks/useFundCardSetupOnrampEventListeners.ts
rustam-cb: There is already loosely coupled hook `setupOnrampEventListeners`. And this is just a wrapper to split out the code

src/fund/components/FundCardPaymentMethodDropdown.tsx
rustam-cb: not needed

src/fund/components/FundCardPaymentMethodDropdown.tsx
rustam-cb: `FundCardPaymentMethodSelectRow` already does that

src/fund/components/FundCardProvider.tsx
rustam-cb: It was for future implementation, but removed for now.

src/fund/components/FundCard.tsx
rustam-cb: We do. I'll have supporting different currencies in a followup separate PR

src/fund/components/FundCardPaymentMethodDropdown.tsx
rustam-cb: Moved.

src/fund/components/FundCardAmountInput.tsx
rustam-cb: Not sure what that would change but the current implementation works fine.

src/fund/components/FundButton.tsx
rustam-cb: Will have that in a separate PR

src/fund/components/FundCardAmountInputTypeSwitch.tsx
```diff
@@ -0,0 +1,99 @@
...
+    );
+  }, [
+    fundAmountCrypto,
+    fundAmountFiat,
+    selectedInputType,
+    formatUSD,
+    formatCrypto,
+  ]);
+
+  if (exchangeRateLoading || !exchangeRate) {
```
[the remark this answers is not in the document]
rustam-cb: This is going to be handled in a follow up PRs

src/fund/components/FundCardCurrencyLabel.tsx
```diff
@@ -0,0 +1,23 @@
...
+  return (
+    <span
+      ref={ref}
+      className={cn(
+        text.body,
+        color.foregroundMuted,
+        'flex items-center justify-center bg-transparent',
+        'text-6xl leading-none outline-none',
+      )}
+      data-testid="currencySpan"
```
alessey, line 18: nit: ockCurrencySpan

src/fund/components/FundCardHeader.tsx
```diff
@@ -0,0 +1,13 @@
+import { cn, text } from '@/styles/theme';
+import type { FundCardHeaderPropsReact } from '../types';
+import { useFundContext } from './FundCardProvider';
+
+export function FundCardHeader({ className }: FundCardHeaderPropsReact) {
+  const { headerText } = useFundContext();
+
+  return (
+    <div className={cn(text.headline, className)} data-testid="fundCardHeader">
```
alessey, line 9: nit: testIds are generally prefixed with ock, i.e. ockFundCardHeader

src/fund/components/FundCardCurrencyLabel.tsx
rustam-cb: updated

src/fund/components/FundCardHeader.tsx
rustam-cb: updated

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -0,0 +1,148 @@
...
+import { formatDecimalInputValue } from '../utils/formatDecimalInputValue';
+import { truncateDecimalPlaces } from '../utils/truncateDecimalPlaces';
+import { FundCardCurrencyLabel } from './FundCardCurrencyLabel';
+import { useFundContext } from './FundCardProvider';
+
+export const FundCardAmountInput = ({
+  className,
+}: FundCardAmountInputPropsReact) => {
+  // TODO: Get currency label from country
+  const currencyLabel = 'USD';
```
abcrane123, line 14: is this TODO a follow up ?

src/internal/svg/creditCardSvg.tsx
```diff
@@ -0,0 +1,23 @@
...
+    data-testid="ock-creditCardSvg"
+    role="img"
+    aria-label="ock-creditCardSvg"
+    width="14"
+    height="16"
+    viewBox="0 0 14 16"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
```
abcrane123, line 14: title here as well

abcrane123: is this the same as the buy component svg ? if so we already have one `cardSvg`. same for apple pay

src/internal/svg/addSvg.tsx
```diff
@@ -13,7 +13,7 @@ export const addSvg = (
   >
     <path
       d="M7.125 0H5.875V5.37501L0.5 5.37501L0.5 6.62501H5.875V12H7.125V6.62501H12.5V5.37501H7.125V0Z"
-      className={icon.inverse}
+      className={className}
```
abcrane123, line 16: can you add a title here plz

src/fund/components/FundCardAmountInput.tsx
rustam-cb: Yes, its going to be in the follow up PRs.

src/fund/components/FundButton.tsx
```diff
@@ -19,49 +31,122 @@ export function FundButton({
...
+  const buttonColorClass = useMemo(() => {
+    switch (buttonState) {
+      case 'error':
+        return background.error;
+      case 'loading':
+      case 'success':
+        return pressable.primary;
+      default:
+        return pressable.primary;
+    }
```
brendan-defi, line 84: are there cases where you can envision returning anything other than background.error or pressable.primary?

unless you envision a whole bunch of different styles, it would match our style better to do something like:
``` 
if (error) {
  return background.error;
}
return pressable.primary;
```

src/fund/components/FundButton.tsx
```diff
@@ -19,49 +31,122 @@ export function FundButton({
...
+  }, [buttonState]);
+
   const classNames = cn(
     componentTheme,
-    pressable.primary,
-    'px-4 py-3 inline-flex items-center justify-center space-x-2 disabled',
+    buttonColorClass,
+    'px-4 py-3 inline-flex items-center justify-center space-x-2',
     isDisabled && pressable.disabled,
```
brendan-defi, line 91: the idiomatic way to write this using the `cn` utility is with object syntax:
{
  pressable.disabled: isDisabled,
}

src/fund/components/FundButton.tsx
```diff
@@ -19,49 +31,122 @@ export function FundButton({
...
     text.headline,
     border.radius,
     color.inverse,
     className,
   );
```
brendan-defi, line 96: the idiomatic way to write conditional values using the cn utility is with object syntax:
{
pressable.disabled: isDisabled,
}

src/fund/components/FundCardAmountInput.tsx
brendan-defi: i use this in the new Wallet component we are releasing this week, can you use this?

```
new Intl.NumberFormat('en-US', {
    minimumFractionDigits: 2,
    maximumFractionDigits: 5,
  })
```

src/fund/components/FundCardAmountInputTypeSwitch.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const roundedAmount = Number(getRoundedAmount(amount || '0', 2));
+    return `$${roundedAmount}`;
+  }, []);
+
+  const formatCrypto = useCallback(
+    (amount: string) => {
+      return `${truncateDecimalPlaces(amount || '0', 8)} ${asset}`;
+    },
+    [asset],
+  );
```
brendan-defi, line 39: mentioned above a method that I think is probably better for crypto. here's a fiat version:
```
new Intl.NumberFormat('en-US', {
    style: 'currency',
    currency: 'USD',
  })
```

src/fund/components/FundCardSubmitButton.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <FundButton
+      disabled={!fundAmountFiat && !fundAmountCrypto}
+      hideIcon={submitButtonState === 'default'}
+      text={buttonText}
+      className="w-full"
+      fundingUrl={fundingUrl}
+      state={submitButtonState}
+      onClick={() => setSubmitButtonState('loading')}
+      onPopupClose={() => setSubmitButtonState('default')}
```
brendan-defi, line 25: even for simple functions, i believe we prefer callbacks here

src/internal/svg/addSvg.tsx
```diff
@@ -11,9 +11,10 @@ export const addSvg = (
     fill="none"
     xmlns="http://www.w3.org/2000/svg"
   >
+    <title>Add SVG</title>
```
brendan-defi, line 14: this is what is read when accessibility settings are on, so we prefer to make the title "nicer". Recommend keeping this as "add"

src/internal/svg/appleSvg.tsx
```diff
@@ -1,29 +1,24 @@
...
     xmlns="http://www.w3.org/2000/svg"
-    viewBox="0 -29.75 165.5 165.5"
-    preserveAspectRatio="xMidYMid meet"
-    id="Artwork"
-    data-testid="appleSvg"
-    width="24"
-    height="24"
   >
-    <title>Apple Pay Onramp</title>
+    <title>Apple SVG</title>
```
brendan-defi, line 14: same comment about titles. prefer to remove SVG and make as intuitive as possible when it's read out loud

src/internal/svg/creditCardSvg.tsx
```diff
@@ -0,0 +1,24 @@
...
+    data-testid="ock-creditCardSvg"
+    role="img"
+    aria-label="ock-creditCardSvg"
+    width="14"
+    height="16"
+    viewBox="0 0 14 16"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <title>Credit Card SVG</title>
```
brendan-defi, line 14: same comment about titles

src/internal/svg/errorSvg.tsx
```diff
@@ -11,7 +11,7 @@ export const errorSvg = (
     <title>Error SVG</title>
     <path
       d="M8 16C12.4183 16 16 12.4183 16 8C16 3.58171 12.4183 0 8 0C3.58172 0 0 3.58171 0 8C0 12.4183 3.58172 16 8 16ZM11.7576 5.0909L8.84853 8L11.7576 10.9091L10.9091 11.7576L8 8.84851L5.09093 11.7576L4.2424 10.9091L7.15147 8L4.2424 5.0909L5.09093 4.24239L8 7.15145L10.9091 4.24239L11.7576 5.0909Z"
-      fill="#E11D48"
+      fill={fill}
```
brendan-defi, line 14: is it expected that this would take values that are not in our style system? 

would prefer to do this via `background.error` or `background.warning` by importing background from `src/styles`. 

additional benefit of not needing to make this a function

src/styles/theme.ts
```diff
@@ -1,5 +1,5 @@
-import { clsx } from 'clsx';
 import type { ClassValue } from 'clsx';
+import { clsx } from 'clsx';
```
brendan-defi, line 2: nit: can combine these into one import line

src/internal/svg/errorSvg.tsx
rustam-cb: This icon is being used by other component with the specific color `#E11D48`. So I'm just making this to work for me at this point. Somebody else did not like that approach  because it was not supposed to change colors with theme.

src/fund/components/FundCardSubmitButton.tsx
rustam-cb: using `useCallback` should not be preferred in this case as its more cleaner this way and using useCallback in this case would not provide any benefits.

src/fund/components/FundCardAmountInputTypeSwitch.tsx
rustam-cb: I'll keep the currently implemented solution for this iteration

rustam-cb: Will have this updated in follow up PRs

abcrane123:
could you add the lifecycle callbacks to the demo in playground so we can check those are working as expected? i.e onStatus, onSuccess, onError

alessey:
inputs kinda wonky

https://github.com/user-attachments/assets/3e430f7d-f6b7-4677-9a37-0391012af603

alessey:
possible to follow the pattern defined in our docs for lifecycle status?
https://onchainkit.xyz/guides/lifecycle-status

cpcramer:
Not a blocker but currently the scroll wheel updates the input price - this feels odd

alessey (dismissed):

abcrane123 (approved):

alessey (approved):

brendan-defi (approved):

PR #1720, opened 2024-12-11
dev: fix test file type errors and set dev aliases

dschlabach:
**What changed? Why?**
As the repo structure changes, it will be nice to have aliases set up to avoid long import names.

In order to do so in test files, we also need to fix the type errors. I think it's desirable to type check test files because:
1. Tests are code - if we update parameter types, we should also be notified that we need to update tests
2. In a couple cases, tests were passing that should have been failing because no type checks were applied (e.g. calling the wrong vitest method to assert equality)

This will allow us to import like:
`import type { PayTransaction } from '@/core/api/types';`

**Notes to reviewers**

**How has it been tested?**


cpcramer:
LGTM - pulled down the branch as well and did not notice any issues 👍

cpcramer (approved):

alessey (approved):

PR #1727, opened 2024-12-12
chore: Update WalletModal UI

cpcramer:
**What changed? Why?**

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
Screenshot 2024-12-12 at 12 20 17 PM

</td>
    <td>

Screenshot 2024-12-12 at 12 20 59 PM

   </td>
  </tr>
</table>

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/wallet/components/WalletModal.tsx
+++ b/src/wallet/components/WalletModal.tsx
@@ -159,8 +159,7 @@ export function WalletModal({
           border.radius,
           background.default,
           'w-[22rem] p-6 pb-4',
-          'flex flex-col items-center gap-4',
-          'relative',
+          'flex flex-col items-center gap-4 relative',
           '-translate-x-1/2 -translate-y-1/2 fixed top-1/2 left-1/2',
           'transition-opacity duration-200',
           isOpen ? 'opacity-100' : 'opacity-0',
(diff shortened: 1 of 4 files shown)


src/wallet/components/WalletModal.tsx
```diff
@@ -158,9 +158,8 @@ export function WalletModal({
           border.lineDefault,
           border.radius,
           background.default,
-          'w-[323px] p-6 pb-4',
-          'flex flex-col items-center gap-4',
-          'relative',
+          'w-[22rem] p-6 pb-4',
```
dschlabach, line 161: Does this work at either `w-80` or `w-96` ?

Otherwise, we could add an `88` as `22rem` to the `tailwind.config.ts`

cc @brendan-defi @alessey - thoughts?

cpcramer: I suppose` w-80 `or `w-96` could work but I was pairing with Mind and we landed in between at `22rem`.

cpcramer: I'm fine hardcoding properties to exact measurements if tailwind doesn't have a class for it

brendan-defi: i'm creating a size 88 tailwind extension for 22rem for walletisland

cpcramer: Sounds good, I'll update this style when that's in

brendan-defi (approved):

0xAlec (approved):

PR #1730, opened 2024-12-13
Feat: Draggable

brendan-defi:
**What changed? Why?**
* created internal Draggable component
  * configurable grid size enables snapping behavior
  * handles touch and mouse events

**Notes to reviewers**

**How has it been tested?**
vitest, and locally

### DIFF
--- a/src/internal/components/Draggable.tsx
+++ b/src/internal/components/Draggable.tsx
@@ -4,13 +4,15 @@ import { cn } from '../../styles/theme';
 type DraggableProps = {
   children: React.ReactNode;
   gridSize?: number;
-  startingPosition?: { x: number; y: number }; // TODO [BOE-886]: make this based on the parent component's position
+  startingPosition?: { x: number; y: number };
+  enableSnapToGrid?: boolean;
 };
 
 export default function Draggable({
   children,
   gridSize = 1,
   startingPosition = { x: 20, y: 20 },
+  enableSnapToGrid = true,
 }: DraggableProps) {
   const [position, setPosition] = useState(startingPosition);
   const [dragOffset, setDragOffset] = useState({ x: 0, y: 0 });
@@ -23,15 +25,12 @@ export default function Draggable({
     [gridSize],
   );
 
-  const handleDragStart = (e: React.MouseEvent | React.TouchEvent) => {
+  const handleDragStart = (e: React.PointerEvent) => {
     setIsDragging(true);
 
-    const clientX = 'touches' in e ? e.touches[0].clientX : e.clientX;
-    const clientY = 'touches' in e ? e.touches[0].clientY : e.clientY;
-
     setDragOffset({
-      x: clientX - position.x,
-      y: clientY - position.y,
+      x: e.clientX - position.x,
+      y: e.clientY - position.y,
     });
   };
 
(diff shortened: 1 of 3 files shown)


src/internal/components/Draggable.tsx
```diff
@@ -0,0 +1,93 @@
...
+        x: snapToGrid(prev.x),
+        y: snapToGrid(prev.y),
+      }));
+      setIsDragging(false);
+    };
+
+    document.addEventListener('mousemove', handleGlobalMove);
+    document.addEventListener('touchmove', handleGlobalMove);
+    document.addEventListener('mouseup', handleGlobalEnd);
+    document.addEventListener('touchend', handleGlobalEnd);
```
alessey, line 64: should be able to use [pointer events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events) here to support both touch and mouse

src/internal/components/Draggable.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [position, setPosition] = useState(startingPosition);
+  const [dragOffset, setDragOffset] = useState({ x: 0, y: 0 });
+  const [isDragging, setIsDragging] = useState(false);
+
+  const snapToGrid = useCallback(
+    (positionValue: number) => {
+      return Math.round(positionValue / gridSize) * gridSize;
+    },
+    [gridSize],
+  );
```
alessey, line 24: can this be disabled?   I believe touch or pointer events on mobile could return decimals.

cpcramer: + How does a developer disable this draggable functionality from a component?

brendan-defi: > How does a developer disable this draggable functionality from a component?

developer does not disable draggability on this component. this component is specifically for enabling draggability. if they don't want draggability, they should not use the component called `Draggable`.

> can this be disabled? I believe touch or pointer events on mobile could return decimals.

IIUC, you're referring to the snap-to-grid right? I didn't realize that decimals were possible so i didn't see a need to disable. i'll add that in

src/internal/components/Draggable.tsx
brendan-defi: ack, will refactor, thank you for the pointer (no pun intended :P)

src/internal/components/Draggable.tsx
alessey: on mobile where screens have a higher actual resolution then the pixel count they display this is possible.

src/internal/components/Draggable.tsx
```diff
@@ -0,0 +1,84 @@
...
+      data-testid="ockDraggable"
+      className={cn(
+        'fixed select-none',
+        isDragging ? 'cursor-grabbing' : 'cursor-grab',
+      )}
+      style={{
+        left: `${position.x}px`,
+        top: `${position.y}px`,
+        zIndex: 1000,
+        touchAction: 'none',
```
alessey, line 77: these two styles can be classes.  may want to move zIndex to a const file as const zIndex = { modal: 1000 } so we can use element type zindexes across components.

src/internal/components/Draggable.tsx
```diff
@@ -0,0 +1,84 @@
...
+  }, [isDragging, dragOffset, snapToGrid, enableSnapToGrid]);
+
+  return (
+    <div
+      data-testid="ockDraggable"
+      className={cn(
+        'fixed select-none',
+        isDragging ? 'cursor-grabbing' : 'cursor-grab',
+      )}
+      style={{
```
cpcramer, line 73: Is inline styling unavoidable here? We usually try to avoid it

src/internal/components/Draggable.tsx
brendan-defi: TIL, thank you! 

i've made the update so that snapToGrid is optional

src/internal/components/Draggable.tsx
brendan-defi: with regard to `left` and `top` — these cannot be classes. class-based was my first approach, but I found that since tailwind classes are applied at build-time, they do not recalculate. and since these properties need to be dynamic (they govern in real-time where the draggable component is on the page), inline styles is required here.

to double check, i just tried this, and can confirm that this does not work:
image

src/internal/components/Draggable.tsx
brendan-defi: see response to similar point from Adam: https://github.com/coinbase/onchainkit/pull/1730#discussion_r1888380288

src/internal/components/Draggable.tsx
brendan-defi: I like the zIndex const idea though, i've implemented that with placeholder values

src/internal/components/Draggable.tsx
```diff
@@ -0,0 +1,85 @@
...
+  gridSize?: number;
+  startingPosition?: { x: number; y: number };
+  enableSnapToGrid?: boolean;
+};
+
+export default function Draggable({
+  children,
+  gridSize = 1,
+  startingPosition = { x: 20, y: 20 },
+  enableSnapToGrid = true,
```
alessey, line 16: nit: since this is default true, may be an easier developer implementation to add disableSnapToGrid to turn it off vs. enableSnapToGrid={false}

src/internal/components/Draggable.tsx
alessey: i meant touchAction: 'none' -> touch-none above

alessey (approved):

src/internal/components/Draggable.tsx
brendan-defi: ah, sorry, yeah that can be, i will move

src/internal/components/Draggable.tsx
brendan-defi: ah, sorry i meant to default this to false.

alessey: in that case you could probably just call it `snapToGrid`

dschlabach (approved):

PR #1731, opened 2024-12-13
Feat: QrCode Component

brendan-defi:
**What changed? Why?**
* Implemented components and utilities for creating beautiful QR codes

Default
image

Base
image

Cyberpunk
image

Hacker
image

**Notes to reviewers**

**How has it been tested?**
locally, and with test coverage


src/internal/components/QrCode/QrCodeSvg.tsx
```diff
@@ -0,0 +1,183 @@
...
+  quietZone?: number;
+  quietZoneBorderRadius?: number;
+  ecl?: 'L' | 'M' | 'Q' | 'H';
+  gradientType?: 'radial' | 'linear';
+};
+
+export function QrCodeSvg({
+  value,
+  size = QR_CODE_SIZE,
+  backgroundColor = '#ffffff',
```
alessey, line 42: should these default to theme colors to support dark mode?

src/internal/components/QrCode/gradientConstants.ts
```diff
@@ -0,0 +1,65 @@
...
+  },
+  black: {
+    startColor: '#a1a1aa',
+    endColor: '#27272a',
+  },
+  baseBlue: {
+    startColor: '#0052ff',
+    endColor: '#b2cbff',
+  },
+};
```
alessey, line 42: is it possible to make all of these pull from theme colors so when a user customizes a theme (or adds their own) the qr code adapts?

src/internal/components/QrCode/QrCodeSvg.tsx
```diff
@@ -0,0 +1,183 @@
...
+                key={`${gradientColor}${offset}`}
+                offset={offset}
+                stopColor={gradientColor}
+                stopOpacity={1}
+              />
+            ))}
+          </radialGradient>
+        ) : (
+          <linearGradient
+            id="linearGrad"
```
alessey, line 143: concat with [useId](https://react.dev/reference/react/useId) to ensure no conflicts

src/internal/components/QrCode/QrCodeSvg.tsx
```diff
@@ -0,0 +1,183 @@
...
+  logoBackgroundColor?: string;
+  logoMargin?: number;
+  logoBorderRadius?: number;
+  quietZone?: number;
+  quietZoneBorderRadius?: number;
+  ecl?: 'L' | 'M' | 'Q' | 'H';
+  gradientType?: 'radial' | 'linear';
+};
+
+export function QrCodeSvg({
```
alessey, line 39: may want to wrap in [memo](https://react.dev/reference/react/memo)

src/internal/components/QrCode/useCorners.tsx
```diff
@@ -0,0 +1,83 @@
...
+      <g>
+        <rect
+          x={0}
+          y={0}
+          rx={9.5}
+          ry={9.5}
+          width={rectSize}
+          height={rectSize}
+          fill={fillColor}
+          id="Corner"
```
alessey, line 25: ids should be unique

src/internal/components/QrCode/useDotsPath.test.ts
```diff
@@ -0,0 +1,97 @@
...
+    const { result: roundLogoResult } = renderHook(() =>
+      useDotsPath(withRoundLogo),
+    );
+    const { result: squareLogoResult } = renderHook(() =>
+      useDotsPath({ ...withRoundLogo, logoBorderRadius: 0 }),
+    );
+
+    expect(roundLogoResult.current).not.toBe(squareLogoResult.current);
+    // expect(roundLogoResult.current).not.toBe('');
+    // expect(squareLogoResult.current).not.toBe('');
```
alessey, line 70: nit: delete

src/internal/components/QrCode/useLogo.tsx
```diff
@@ -0,0 +1,71 @@
...
+  logoSize,
+  logoBackgroundColor,
+  logoMargin,
+  logoBorderRadius,
+}: RenderLogoProps) {
+  const svgLogo = useMemo(() => {
+    if (React.isValidElement(logo)) {
+      logo = `data:image/svg+xml;charset=utf-8,${encodeURIComponent(
+        ReactDOMServer.renderToString(logo),
+      )}`;
```
alessey, line 31: can't we just display the svg?

src/internal/components/QrCode/useMatrix.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export function useMatrix(
+  value: string,
+  errorCorrectionLevel: 'L' | 'M' | 'Q' | 'H',
+) {
+  if (!value) {
+    return [];
+  }
+
+  const matrix = useMemo(() => {
```
alessey, line 12: can't conditionally call useMemo https://react.dev/reference/rules/rules-of-hooks

src/internal/components/QrCode/useMatrix.ts
```diff
@@ -0,0 +1,27 @@
...
+    const arr = Array.prototype.slice.call(
+      QRCode.create(value, { errorCorrectionLevel }).modules.data,
+      0,
+    );
```
alessey, line 16: why do you need to clone this?

dschlabach (approved):

src/internal/components/QrCode/QrCodeSvg.tsx
```diff
@@ -0,0 +1,183 @@
+import { useMemo } from 'react';
```
cpcramer, line 1: This should live in `src/internal/svg`

src/internal/components/QrCode/QrCodeSvg.tsx
```diff
@@ -0,0 +1,183 @@
...
+  backgroundColor?: string;
+  logo?: React.ReactNode;
+  logoSize?: number;
+  logoBackgroundColor?: string;
+  logoMargin?: number;
+  logoBorderRadius?: number;
+  quietZone?: number;
+  quietZoneBorderRadius?: number;
+  ecl?: 'L' | 'M' | 'Q' | 'H';
+  gradientType?: 'radial' | 'linear';
```
cpcramer, line 36: I know many of these are optional but curious if we even need all of these props - do we want this low level of customization for a QR code? I imagine that a few of these params will always stay the same and therefore we can just default in the func.

src/internal/components/QrCode/gradientConstants.ts
cpcramer: Agreed - we should avoid having to update this file when adding new themes. 

Has the QR component been tested with custom themes? Curious if it's updating as expected.

src/internal/components/QrCode/QrCodeSvg.tsx
brendan-defi: eventually yes. unfortunately, enough of the code here assumes lightmode that getting to dark mode compatible was out of scope. this is what it currently looks like if we try to use the darkmode backgrounds:

image

dark-mode compatibility is going to be a fast-follow

alessey (approved):
approved with themes/dark mode as fast follows

PR #1733, opened 2024-12-13
Chore: Swap Refactors

brendan-defi:
**What changed? Why?**
* refactored swap to prep for wallet island

**Notes to reviewers**

**How has it been tested?**
locally


src/swap/components/Swap.tsx
```diff
@@ -75,6 +76,7 @@ export function Swap({
         data-testid="ockSwap_Container"
       >
         <div className="mb-4 flex items-center justify-between">
+          {backButton}
```
dschlabach, line 79: What's the `backButton` for again?

brendan-defi: I need a backbutton in walletisland's implementation of swap
image

dschlabach: To me this feels like something that should live outside of Swap. I don't think that we want to ship that option to third party consumers of the <Swap /> component.

Could it instead be a button that's positioned absolutely to the parent? I think that'd make more sense than adding a `backButton` prop

brendan-defi: the issue is how to center the title between the backbutton and the settings button. 

the parent container for these is set to `justify-between`. this is the correct default behavior. i have not figured out a way to override this from "within" the container — that is, no class directive on the swap-title makes it center-aligned. 

AFAICT, i need a third element so that the title will move to the center.

brendan-defi: The best that I can come up with is to make the topbar relative, wrap swapSettings in a div, and make that div absolutely-positioned with `right-0`.

brendan-defi:
closing in favor of a cleaner branch

PR #1738, opened 2024-12-16
typo fix Update types.ts

mdqst:
**What changed? Why?**

A typo in the word "available," which was previously written as "avaialable".

Fixed.


dschlabach:
Fixed in #1735

PR #1752, opened 2024-12-17
headerLeftContent in Swap

brendan-defi:
**What changed? Why?**
* Added a prop to accommodate content on the left-side of the header (for example, to have a back-button)
* change the title prop from a string to ReactNode to allow for restyling

**Notes to reviewers**

**How has it been tested?**
locally, in playground


playground/nextjs-app-router/onchainkit/package.json
```diff
@@ -59,6 +60,7 @@
     "@storybook/test-runner": "^0.19.1",
     "@testing-library/jest-dom": "^6.4.6",
     "@testing-library/react": "^14.2.0",
+    "@types/qrcode": "^1",
```
alessey, line 63: nit: these don't seem related to this PR

fakepixels:
Thank you!

playground/nextjs-app-router/onchainkit/package.json
brendan-defi: yeah, they're not. they're from the QR code component we just merged. I guess when I tested things in playground they got added

alessey (approved):

src/swap/types.ts
```diff
@@ -298,7 +298,8 @@ export type SwapReact = {
   onError?: (error: SwapError) => void; // An optional callback function that handles errors within the provider.
   onStatus?: (lifecycleStatus: LifecycleStatus) => void; // An optional callback function that exposes the component lifecycle state
   onSuccess?: (transactionReceipt: TransactionReceipt) => void; // An optional callback function that exposes the transaction receipt
-  title?: string; // Title for the Swap component. (default: "Swap")
+  title?: ReactNode; // Title for the Swap component. (default: "Swap")
+  headerLeftContent?: ReactNode; // Header left content for the Swap component (eg. back button).
```
cpcramer, line 302: Why not just call this the backButton? Or make it into a back button sub component?

cpcramer: what other elements are we expecting to be passed through here?

brendan-defi: that was the original change, but we made it more generic to be able to handle things like logos, tray-open/close-buttons, etc.

cpcramer (approved):

PR #1754, opened 2024-12-17
docs: Enable auto logged in components on landing page

cpcramer:
**What changed? Why?**
Create new demo app provider which includes our auto logged in demo wallet. This is to achieve a "logged in" state for our components on the landing page, allowing users to see the intended components rather than a bunch of Connect Wallet buttons. 

For now, we are only auto logging in for the Wallet and Swap component. The Fund and Transaction component would have to be set to disabled. We've decided that it's better to leave these and allow the user to connect their own wallet to test out the functionality. 

Screenshot 2024-12-17 at 1 19 56 PM

Screenshot 2024-12-17 at 1 20 04 PM

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/site/docs/components/AppDemo.tsx
+++ b/site/docs/components/AppDemo.tsx
@@ -16,10 +16,10 @@ const queryClient = new QueryClient();
 // WARNING: This is a publicly exposed private key used only for documentation demos.
 // Do not use this key for any real transactions or store any assets in this account.
 const DUMMY_PK =
-  '0xa67bdd8a49324aa36cb3f7f7064b9b560e3aa653b774be9793415c0a6fc62cf8';
+  '0xa67bdd8a49324aa36cb3f7f7064b9b560e3aa653b774be9793415c0a6fc62cf8' as const;
 
 const demoWalletConnector = (_config: Config) => {
-  const account = privateKeyToAccount(DUMMY_PK as `0x${string}`);
+  const account = privateKeyToAccount(DUMMY_PK);
 
   const client = createWalletClient({
     account,
--- a/site/docs/components/AppDemo.tsx
+++ b/site/docs/components/AppDemo.tsx
@@ -8,6 +8,7 @@ import { coinbaseWallet } from 'wagmi/connectors';
 import '@coinbase/onchainkit/styles.css';
 import { createWalletClient } from 'viem';
 import { privateKeyToAccount } from 'viem/accounts';
+import type { Config } from 'wagmi';
 import { useTheme } from '../contexts/Theme.tsx';
 
 const queryClient = new QueryClient();
@@ -17,7 +18,7 @@ const queryClient = new QueryClient();
 const DUMMY_PK =
   '0xa67bdd8a49324aa36cb3f7f7064b9b560e3aa653b774be9793415c0a6fc62cf8';
 
-const demoWalletConnector = (_config: any) => {
+const demoWalletConnector = (_config: Config) => {
   const account = privateKeyToAccount(DUMMY_PK as `0x${string}`);
 
   const client = createWalletClient({
--- a/site/docs/components/AppDemo.tsx
+++ b/site/docs/components/AppDemo.tsx
@@ -1,0 +1,118 @@
+'use client';
+import { OnchainKitProvider } from '@coinbase/onchainkit';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import type { ReactNode } from 'react';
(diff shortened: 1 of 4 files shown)


site/docs/components/AppDemo.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { coinbaseWallet } from 'wagmi/connectors';
+import '@coinbase/onchainkit/styles.css';
+import { createWalletClient } from 'viem';
+import { privateKeyToAccount } from 'viem/accounts';
+import { useTheme } from '../contexts/Theme.tsx';
+
+const queryClient = new QueryClient();
+
+const demoWalletConnector = () => () => {
+  const demoWalletPK = import.meta.env.VITE_DEMO_WALLET_PK;
```
cpcramer, line 16: This demo wallet will never have a value associated with it therefore I believe it's fine to leak the PK in the client.

site/docs/components/AppDemo.tsx
```diff
@@ -0,0 +1,120 @@
...
+
+const queryClient = new QueryClient();
+
+// Default value to avoid build errors when running locally.
+const DEFAULT_DEMO_PK =
+  '0x1234567890123456789012345678901234567890123456789012345678901234';
+
+const demoWalletConnector = () => () => {
+  // Demo account with no real assets
+  const demoWalletPK = import.meta.env.VITE_DEMO_WALLET_PK || DEFAULT_DEMO_PK;
```
cpcramer, line 21: Demo accounts with no value attached. We could just hard code VITE_DEMO_WALLET_PK

site/docs/components/AppDemo.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { useTheme } from '../contexts/Theme.tsx';
+
+const queryClient = new QueryClient();
+
+// WARNING: This is a publicly visible demo/test account with absolutely no real assets or value.
+// It is used ONLY for documentation examples and should NEVER be used in production or to store any assets.
+const DUMMY_PK =
+  '0xa67bdd8a49324aa36cb3f7f7064b9b560e3aa653b774be9793415c0a6fc62cf8';
+
+const demoWalletConnector = () => () => {
```
brendan-defi, line 20: nit: i find chained arrow functions pretty hard to reason about. consider making the outer function a named function, and return the inner arrow function

site/docs/components/AppDemo.tsx
```diff
@@ -0,0 +1,119 @@
...
+const queryClient = new QueryClient();
+
+// WARNING: This is a publicly exposed private key used only for documentation demos.
+// Do not use this key for any real transactions or store any assets in this account.
+const DUMMY_PK =
+  '0xa67bdd8a49324aa36cb3f7f7064b9b560e3aa653b774be9793415c0a6fc62cf8';
+
+const demoWalletConnector = (_config: Config) => {
+  const account = privateKeyToAccount(DUMMY_PK as `0x${string}`);
+
```
dschlabach, line 23: not strictly necessary, but can do this which is kinda nice :)

```suggestion
const DUMMY_PK =
  '0xa67bdd8a49324aa36cb3f7f7064b9b560e3aa653b774be9793415c0a6fc62cf8' as const;

const demoWalletConnector = (_config: Config) => {
  const account = privateKeyToAccount(DUMMY_PK);

```

dschlabach (approved):

site/docs/components/AppDemo.tsx
cpcramer: Very nice

site/docs/components/landing/WalletDemo.tsx
```diff
@@ -61,10 +59,9 @@ function WalletDemo() {
             Wallet
           </WalletDropdownLink>
           <WalletDropdownBasename />
-          <WalletDropdownDisconnect />
```
brendan-defi, line 64: given that we are removing disconnect option, consider adding a warning like, "wallet address for demonstration purposes only. do not send funds to this address"

site/docs/components/AppDemo.tsx
dschlabach: BTW the reason this is slightly better is because you're allowing Typescript to check that the const adheres to `0x${string}` rather than you telling it that it adheres to `0x${string}`

In this case there'd almost never be any issues with it as we won't be updating it, but it's nice to do it the more robust way anyway

brendan-defi (approved):

PR #1771, opened 2024-12-19
fix: improve support for OnchainKit in Next.js 14+ apps

dschlabach:
**What changed? Why?**
Next.js uses server components by default. When users attempt to bring OnchainKit components in a server component, they get an obscure error because the client boundary is not set properly at our component level.

**Notes to reviewers**
I've started with just the Identity component, we can add support for other components in future PRs

**How has it been tested?**
- Built and packed package locally using `yarn build && yarn pack`
- Create and test in various environments: Next.js 14 app (in server component), Vite app, Node script.


alessey (approved):

abcrane123 (approved):
nice

PR #1791, opened 2024-12-25
fix: typos in documentation files

zeevick10:
The utility name `isWalletASmartCoinbaseWallet` should be changed to `isWalletACoinbaseSmartWallet`, which matches the correct function name specified in the import and code example.


cpcramer:
Thanks for the contribution!

PR #1793, opened 2025-01-03
feat: wallet island

brendan-defi:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
```diff
@@ -0,0 +1,51 @@
...
+  PortfolioAPIResponse,
+  PortfolioTokenBalanceAPIResponse,
+  PortfolioTokenBalances,
+} from '@/core/api/types';
+import { isApiError } from '@/core/utils/isApiResponseError';
+import { type UseQueryResult, useQuery } from '@tanstack/react-query';
+
+export function usePortfolioTokenBalances({
+  addresses,
+}: GetPortfolioTokenBalancesParams): UseQueryResult<PortfolioTokenBalances[]> {
```
cpcramer, line 13: Rename to `usePortfolioTokenBalancesOptions`

cpcramer: We should standardize on how we name this, I see the Get*Params version in some places and the *Options version in others

src/wallet/components/WalletIslandWalletActions.tsx
```diff
@@ -0,0 +1,104 @@
...
+  const handleQr = useCallback(() => {
+    setShowQr(true);
+  }, [setShowQr]);
+
+  const handleRefreshPortfolioData = useCallback(async () => {
+    if (
+      portfolioDataUpdatedAt &&
+      Date.now() - portfolioDataUpdatedAt < 1000 * 15
+    ) {
+      return; // TODO: Add toast
```
cpcramer, line 33: Add Toast todo

cpcramer:
We should center the ConnectWallet button and WalletIsland demo in the playground:
Screenshot 2025-01-06 at 1 42 51 PM

src/wallet/components/WalletIslandSwap.tsx
```diff
@@ -0,0 +1,120 @@
...
+      onClick={handleCloseSwap}
+      className={cn(
+        pressable.default,
+        border.radius,
+        border.default,
+        'flex items-center justify-center p-4',
+      )}
+      aria-label="Back button"
+    >
+      <div>{backArrowSvg}</div>
```
cpcramer, line 66: Remove outer div element since it's not providing any styling or layout benefits

src/wallet/components/WalletIslandQrReceive.tsx
```diff
@@ -0,0 +1,142 @@
...
+          onClick={handleCloseQr}
+          className={cn(
+            pressable.default,
+            border.radius,
+            border.default,
+            'flex items-center justify-center p-4',
+          )}
+          aria-label="Back"
+        >
+          <div>{backArrowSvg}</div>
```
cpcramer, line 94: Can remove outer div

src/wallet/components/WalletIslandQrReceive.tsx
```diff
@@ -0,0 +1,142 @@
...
+            onClick={() => handleCopyAddress('icon')}
+            className={cn(
+              pressable.default,
+              border.radius,
+              border.default,
+              'flex items-center justify-center p-3',
+            )}
+            aria-label="Copy icon"
+          >
+            <div>{copySvg}</div>
```
cpcramer, line 109: Can remove outer div

src/wallet/components/WalletIslandQrReceive.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+      <QrCodeSvg value={address ? `ethereum:${address}` : ''} />
+
+      <button
+        type="button"
+        className={cn(border.radius, pressable.alternate, 'w-full p-3')}
+        onClick={() => handleCopyAddress('button')}
+        aria-label="Copy button"
+      >
+        <span>{copyButtonText}</span>
```
cpcramer, line 138: Can remove outer span I believe

src/wallet/components/WalletIslandWalletActions.tsx
brendan-defi: thanks. going to remove without adding toast for now. will add if the non-update-within-15-seconds causes user confusion

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
brendan-defi: i don't have a strong opinion, but at least in src/core/api/types it seems like we use params a lot more than options

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
```diff
@@ -0,0 +1,65 @@
...
+    queryFn: async () => {
+      const response = await getPortfolioTokenBalances({
+        addresses,
+      });
+
+      if (isApiError(response)) {
+        throw new Error(response.message);
+      }
+
+      const userPortfolio = response.tokens[0];
```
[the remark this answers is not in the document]
brendan-defi: few things going on here:

1. `response.tokens` is a bad label. it should be `response.portfolios` (i've merged a change into rpc-proxy, will deploy and make the change here today). 

2. `response.portfolios` should never be empty, because it returns the user's address and their portfolioFiatValue (even if 0) and their tokenHoldings. 

3. all that said, it still seems better to add an empty check, so i'll do that

playground/nextjs-app-router/components/Demo.tsx
```diff
@@ -140,12 +147,24 @@ function Demo() {
...
+) {
+  if (activeComponent === OnchainKitComponent.WalletIslandDefault) {
+    if (anchorPosition?.includes('top')) {
+      return 'justify-start';
+    }
+    return 'justify-end';
+  }
+
+  return 'items-center justify-center';
+}
```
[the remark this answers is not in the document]
brendan-defi: gonna end up deleting this because walletisland no longer has an anchor

src/wallet/components/WalletIslandQrReceive.tsx
```diff
@@ -0,0 +1,133 @@
...
+          <button
+            type="button"
+            onClick={() => handleCopyAddress('icon')}
+            className={cn(
+              pressable.alternate,
+              text.legal,
+              color.foreground,
+              border.default,
+              border.radius,
+              'absolute top-full right-[0%] z-10 mt-0.5 px-1.5 py-0.5 opacity-0 transition-opacity group-hover:opacity-100',
```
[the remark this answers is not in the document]
brendan-defi: good catch, right-0 is fine

src/wallet/components/WalletIslandAddressDetails.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <button
+          type="button"
+          onClick={handleCopyAddress}
+          className={cn(
+            pressable.alternate,
+            text.legal,
+            color.foreground,
+            border.default,
+            border.radius,
+            'absolute top-full right-[0%] z-10 mt-0.5 px-1.5 py-0.5 opacity-0 transition-opacity group-hover:opacity-100',
```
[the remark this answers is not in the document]
brendan-defi: good catch, right-0 is fine

src/wallet/components/WalletIslandProvider.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { usePortfolioTokenBalances } from '@/core-react/wallet/hooks/usePortfolioTokenBalances';
+import { type ReactNode, createContext, useContext, useState } from 'react';
+import type { WalletIslandContextType } from '../types';
+import { useWalletContext } from './WalletProvider';
+
+type WalletIslandProviderReact = {
+  children: ReactNode;
+};
+
+const emptyContext = {} as WalletIslandContextType;
```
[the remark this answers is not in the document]
brendan-defi: unfortunately this doesn't work. 

type annotations - `variable: Type` - say what to expect, and error if the expectation is not met. 

type assertions - `variable = value as Type` assert that the value _will be_ of the type, so don't error even if the value isn't currently meeting the type. 

image

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
```diff
@@ -0,0 +1,65 @@
...
+  PortfolioTokenBalances,
+  PortfolioTokenWithFiatValue,
+} from '@/core/api/types';
+import { isApiError } from '@/core/utils/isApiResponseError';
+import { type UseQueryResult, useQuery } from '@tanstack/react-query';
+
+export function usePortfolioTokenBalances({
+  addresses,
+}: GetPortfolioTokenBalancesParams): UseQueryResult<PortfolioTokenBalances> {
+  const actionKey = `usePortfolioTokenBalances-${addresses}`;
```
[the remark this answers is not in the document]
brendan-defi: `queryKey` is a required parameter in useQuery. It's what useQuery uses to cache the query, and needs to be unique: https://tanstack.com/query/latest/docs/framework/react/guides/query-keys

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
```diff
@@ -0,0 +1,73 @@
...
+        addresses,
+      });
+
+      if (isApiError(response)) {
+        throw new Error(response.message);
+      }
+
+      if (response.portfolios.length === 0) {
+        return {
+          address: '',
```
alessey, line 28: in this case would address be the address they passed in?

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
```diff
@@ -0,0 +1,73 @@
...
+            }) as PortfolioTokenWithFiatValue,
+        ),
+      };
+
+      const filteredPortfolio = {
+        ...transformedPortfolio,
+        tokenBalances: transformedPortfolio.tokenBalances.filter(
+          (tokenBalance: PortfolioTokenWithFiatValue) =>
+            tokenBalance.cryptoBalance > 0,
+        ),
```
alessey, line 60: this endpoint includes tokens with a 0 balance?  things you used to own?  or just dust?

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
```diff
@@ -0,0 +1,73 @@
...
+
+      if (response.portfolios.length === 0) {
+        return {
+          address: '',
+          portfolioBalanceUsd: 0,
+          tokenBalances: [],
+        };
+      }
+
+      const userPortfolio = response.portfolios[0];
```
alessey, line 34: how does this work if you pass in an array of addresses?

src/core/api/getPortfolioTokenBalances.ts
```diff
@@ -0,0 +1,31 @@
...
+    >(CDP_GET_PORTFOLIO_TOKEN_BALANCES, [{ addresses }]);
+    if (res.error) {
+      return {
+        code: `${res.error.code}`,
+        error: 'Error fetching portfolio token balances',
+        message: res.error.message,
+      };
+    }
+    return res.result;
+  } catch (_error) {
```
alessey, line 24: _error?

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
```diff
@@ -0,0 +1,73 @@
...
+
+      const transformedPortfolio: PortfolioTokenBalances = {
+        address: userPortfolio.address,
+        portfolioBalanceUsd: userPortfolio.portfolio_balance_usd,
+        tokenBalances: userPortfolio.token_balances.map(
+          (tokenBalance: PortfolioTokenBalanceAPIResponse) =>
+            ({
+              address:
+                tokenBalance.symbol === 'ETH' ? '' : tokenBalance.address,
+              chainId: tokenBalance.chain_id,
```
alessey, line 44: were these returned snake case from the original endpoint?

src/internal/components/PressableIcon.tsx
```diff
@@ -0,0 +1,24 @@
...
+        border.radius,
+        border.default,
+        'flex items-center justify-center',
+        className,
+      )}
+    >
+      {children}
+    </div>
+  );
+}
```
alessey, line 24: what makes this a pressableIcon?  I would expect something with icon and onClick props

src/internal/svg/arrowUpRightSvg.tsx
```diff
@@ -0,0 +1,17 @@
+import { icon } from '../../styles/theme';
+
+export const arrowUpRightSvg = (
+  <svg
+    width="12"
+    height="12"
```
alessey, line 6: width/height=100% may be more flexible for these

src/internal/svg/clockSvg.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export const clockSvg = (
+  <svg
+    width="28"
+    height="28"
+    viewBox="0 0 28 28"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <title>Clock Icon</title>
```
alessey, line 11: just "Clock" would match the other titles

src/internal/svg/refreshSvg.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const refreshSvg = (
+  <svg
+    width="100%"
+    height="100%"
+    viewBox="0 0 32 32"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <title>Refresh SVG</title>
```
alessey, line 11: iirc, title is what will be read by screen readers, so we should make this something that makes sense to hear

src/wallet/components/Wallet.tsx
```diff
@@ -1,53 +1,173 @@
...
 
-function WalletContent({ children, className }: WalletReact) {
+export const Wallet = ({
+  children,
+  className,
+  draggable = false,
+  startingPosition = {
+    x: window.innerWidth - 250,
+    y: window.innerHeight - 100,
+  },
```
alessey, line 25: the initial position of the connect wallet button won't be determined by the consumer?

src/wallet/components/Wallet.tsx
```diff
@@ -1,53 +1,173 @@
...
+          <WalletSubComponent
+            connect={connect}
+            connectRef={connectRef}
+            dropdown={dropdown}
+            island={island}
+            isOpen={isOpen}
+            alignSubComponentRight={alignSubComponentRight}
+            showSubComponentAbove={showSubComponentAbove}
+          />
+        </Draggable>
```
alessey, line 107: whats draggable if this is dropdown?

src/wallet/components/WalletIslandDraggable.tsx
```diff
@@ -0,0 +1,27 @@
...
+      </ConnectWallet>
+      <WalletIsland>
+        <WalletIslandWalletActions />
+        <AddressDetails />
+        <WalletIslandTransactionActions />
+        <WalletIslandTokenHoldings />
+      </WalletIsland>
+    </Wallet>
+  );
+}
```
alessey, line 27: seems like this would just be a standard WalletIslandDefault with a draggable prop?

src/wallet/components/WalletIslandProvider.tsx
```diff
@@ -0,0 +1,63 @@
...
+  } = usePortfolioTokenBalances({ addresses: [address ?? '0x000'] });
+
+  const portfolioFiatValue = portfolioData?.portfolioBalanceUsd;
+  const tokenBalances = portfolioData?.tokenBalances;
+
+  const value = useValue({
+    showSwap,
+    setShowSwap,
+    isSwapClosing,
+    setIsSwapClosing,
```
alessey, line 46: I think it'd be great to be able to customize the island to add my own buttons/panes using this provider, i.e. have a status button for my app instead of swap and use the show/etc. from the provider to show them on click.  just a thought for the future.

src/wallet/components/WalletIslandQrReceive.tsx
```diff
@@ -0,0 +1,121 @@
...
+              'absolute top-full right-0 z-10 mt-0.5 px-1.5 py-0.5 opacity-0 transition-opacity group-hover:opacity-100',
+            )}
+            aria-live="polite"
+            aria-label="Copy tooltip"
+          >
+            {copyText}
+          </button>
+        </div>
+      </div>
+      <QrCodeSvg value={address ? `ethereum:${address}` : ''} />
```
alessey, line 110: what does the qrcode do if no address?

src/wallet/components/WalletIslandQrReceive.tsx
```diff
@@ -0,0 +1,121 @@
...
+      <button
+        type="button"
+        className={cn(border.radius, pressable.alternate, 'w-full p-3')}
+        onClick={() => handleCopyAddress('button')}
+        aria-label="Copy button"
```
alessey, line 115: may be more useful to have the label be an action, so you'd hear "Copy your address" etc.

src/wallet/components/WalletIslandSwap.tsx
```diff
@@ -0,0 +1,95 @@
...
+        onStatus={onStatus}
+        onSuccess={onSuccess}
+        onError={onError}
+        config={config}
+        isSponsored={isSponsored}
+        title={title}
+        experimental={experimental}
+        headerLeftContent={backButton}
+      >
+        <SwapSettings className="w-auto">
```
alessey, line 78: is this class the only thing preventing you from using swap default here?

src/wallet/components/WalletIslandTokenHoldings.tsx
```diff
@@ -0,0 +1,82 @@
...
+          }
+          valueInFiat={Number(tokenBalance.fiatBalance)}
+        />
+      ))}
+    </div>
+  );
+}
+
+function TokenDetails({ token, balance, valueInFiat }: TokenDetailsProps) {
+  const currencySymbol = '$'; // TODO: get from user settings
```
alessey, line 62: we only support USD here right?

src/wallet/components/WalletIslandTransactionActions.tsx
```diff
@@ -0,0 +1,81 @@
...
+        action={() => {
+          window.open('https://wallet.coinbase.com', '_blank');
+        }}
+      />
+      <WalletIslandTransactionAction
+        icon={toggleSvg}
+        label="Swap"
+        action={() => {
+          setShowSwap(true);
+        }}
```
alessey, line 44: nit: extract callbacks to useCallback to prevent creating a new function on every render

src/wallet/components/WalletIslandTransactionActions.tsx
```diff
@@ -0,0 +1,81 @@
...
+      />
+    </div>
+  );
+}
+
+function WalletIslandTransactionAction({
```
alessey, line 50: this would be a cool export if you wanted to build your own action

src/wallet/components/WalletIslandWalletActions.tsx
```diff
@@ -0,0 +1,89 @@
...
+    setShowQr(true);
+  }, [setShowQr]);
+
+  const handleRefreshPortfolioData = useCallback(async () => {
+    if (
+      portfolioDataUpdatedAt &&
+      Date.now() - portfolioDataUpdatedAt < 1000 * 15
+    ) {
+      return;
+    }
```
alessey, line 35: seems weird to throttle an explicit user action to refresh

playground/nextjs-app-router/components/AppProvider.tsx
```diff
@@ -159,6 +168,8 @@ export const AppProvider = ({ children }: { children: React.ReactNode }) => {
         setNFTToken,
         setIsSponsored,
         isSponsored,
+        anchorPosition,
+        setAnchorPosition,
```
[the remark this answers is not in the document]
brendan-defi: thanks, missed this one. deleting

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
brendan-defi: What do you think is best? If we get an empty response from the API, should I put in the address anyway?

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
brendan-defi: per slack convo, these are tokens you used to own. adding BE filter as well

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
alessey: ok, i think you are accepting an array of addresses here though

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
alessey: yea, i'd try to return the most correct response possible

src/internal/svg/arrowUpRightSvg.tsx
brendan-defi: yeah, i am going to spend some time soon standardizing all the SVGs

src/internal/svg/clockSvg.tsx
brendan-defi: okay, i'll remove "icon" from the titles

src/internal/svg/refreshSvg.tsx
brendan-defi: makes sense, i'll remove the "SVG"

src/wallet/components/Wallet.tsx
brendan-defi: this is just the default. if dev enables draggability and passes a value for the `startingPosition` prop, it'll default to that location

src/wallet/components/Wallet.tsx
brendan-defi: we don't have a default for this, but if the dev does this then the wallet-with-normal-dropdown component becomes draggable

src/wallet/components/WalletIslandDraggable.tsx
brendan-defi: we've gotten rid of WalletIslandDefault in favor of `WalletIslandDraggable` (which, as you said, is the-component-formerly-known-as-WalletIslandDefault with draggable prop), and `WalletIslandFixed`

src/wallet/components/WalletIslandProvider.tsx
brendan-defi: yeah agreed. Tina, Mind, and I discussed, and decided to keep the actions standard for now, but would be a nice feature

src/wallet/components/WalletIslandQrReceive.tsx
brendan-defi: should never happen, but this is what it looks like if address comes back empty:

image

src/wallet/components/WalletIslandQrReceive.tsx
brendan-defi: good call, changing all of the aria-labels here

src/wallet/components/WalletIslandTokenHoldings.tsx
brendan-defi: yes, for now only USD

src/wallet/components/WalletIslandTransactionActions.tsx
brendan-defi: makes sense, will do

src/wallet/components/Wallet.tsx
alessey: cool, that doesn't seem like a bad option to have

src/wallet/components/WalletIslandWalletActions.tsx
brendan-defi: okay, i was worried about user spamming refresh, but i will remove

brendan-defi:
> I think with some modifications we could make this more flexible for consumers, allowing them to use the wallet island structure but intermix custom components with ock components

totally agree, i think we'll move to allow more customization in the near future.

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
brendan-defi: okay, will do

src/internal/components/PressableIcon.tsx
brendan-defi: hmm, I use this to standardize the display for all of my icons, including some that are just links. 

i guess i could put a button in here, add onClick props, and then use window.open instead of using anchor tags. 

what do you think?

src/wallet/components/Wallet.tsx
brendan-defi: yeah i like leaving it available to devs who want it

src/internal/components/PressableIcon.tsx
alessey: if you just want a style wrapper for this component, this should probably be located with your component, but a common Pressable icon should probably be something like `<PressableIcon className="" icon="" onPress|onClick=""/>` and possibly an accessibility or arial label as well, and then internally could call useIcon internally

src/wallet/components/WalletIslandQrReceive.tsx
alessey: seems like you'd either leave address required and not render if '', or if you are counting on qrcode handling no address you could make it optional

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
brendan-defi: i did that to match `getPortfolioTokenBalances`, which requires an array. 

but you're right, that's bad, i'll pass in a single address here and transform to what `getPortfolioTokenBalances` expects

src/wallet/components/WalletIslandQrReceive.tsx
brendan-defi: makes sense, will do

src/internal/components/PressableIcon.tsx
brendan-defi: makes sense, i've refactored. for now i'm not using useIcon, but i'll look into that

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
brendan-defi: my understanding is that GO's standard JSON libraries all assume snake case, so I wrote the JSON to use snake case. 

I see that we use camelcase though, so i'll convert to camelcase

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
```diff
@@ -0,0 +1,42 @@
...
+        addresses: [address as Address], // Safe to coerce to Address because useQuery's enabled flag will prevent the query from running if address is undefined
+      });
+
+      if (isApiError(response)) {
+        throw new Error(response.message);
+      }
+
+      if (response.portfolios.length === 0) {
+        return {
+          address: address,
```
alessey, line 26: nit: can use object initialization shorthand for address

src/internal/components/PressableIcon.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <div
+      className={cn(
+        pressable.default,
+        border.radiusInner,
+        border.default,
+        'flex items-center justify-center',
+        className,
+      )}
+    >
```
alessey, line 28: possible to make this whole thing the button?

src/internal/svg/qrIconSvg.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const qrIconSvg = (
+  <svg
+    width="28"
+    height="28"
+    viewBox="0 0 28 28"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <title>QR Code Icon</title>
```
alessey, line 11: nit: QR Code

src/wallet/components/WalletIslandDraggable.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { AddressDetails } from './WalletIslandAddressDetails';
+import { WalletIslandTokenHoldings } from './WalletIslandTokenHoldings';
+import { WalletIslandTransactionActions } from './WalletIslandTransactionActions';
+import { WalletIslandWalletActions } from './WalletIslandWalletActions';
+
+export function WalletIslandDraggable({
+  startingPosition = {
+    x: window.innerWidth - 250,
+    y: window.innerHeight - 100,
+  },
```
alessey, line 15: nit: bottom right seems like an odd default

src/wallet/components/WalletIslandProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const [showSwap, setShowSwap] = useState(false);
+  const [isSwapClosing, setIsSwapClosing] = useState(false);
+  const [showQr, setShowQr] = useState(false);
+  const [isQrClosing, setIsQrClosing] = useState(false);
+  const {
+    data: portfolioData,
+    refetch: refetchPortfolioData,
+    isFetching: isFetchingPortfolioData,
+    dataUpdatedAt: portfolioDataUpdatedAt,
+  } = usePortfolioTokenBalances({ address: address ?? '0x000' });
```
alessey, line 37: should this have a default if no address?

src/wallet/components/WalletIslandQrReceive.tsx
```diff
@@ -0,0 +1,123 @@
...
+    setIsQrClosing(true);
+
+    setTimeout(() => {
+      setShowQr(false);
+    }, 200);
+
+    setTimeout(() => {
+      setIsQrClosing(false);
+    }, 400);
+  }, [setShowQr, setIsQrClosing]);
```
alessey, line 26: does setIsQrClosing trigger the animation, then 200ms later animation is complete and you remove it?  why wait an additional 200ms to set isQrClosing?

src/wallet/components/WalletIslandQrReceive.tsx
```diff
@@ -0,0 +1,123 @@
...
+          setCopyText('Failed to copy');
+        }
+        setTimeout(() => {
+          setCopyText('Copy');
+          setCopyButtonText('Copy address');
+        }, 2000);
+      }
+    },
+    [address],
+  );
```
alessey, line 55: nit: we may want to extract this to a utility in the future to reuse between this and the Address component

src/wallet/components/WalletIslandTokenHoldings.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <span className={cn(text.label1, color.foreground)}>
+            {token.name}
+          </span>
+          <span className={cn(text.legal, color.foregroundMuted)}>
+            {`${balance.toFixed(5)} ${token.symbol}`}
+          </span>
+        </div>
+      </div>
+      <span className={cn(text.label2, color.foregroundMuted)}>
+        {`${currencySymbol}${valueInFiat.toFixed(2)}`}
```
alessey, line 73: nit: unnecessary template strings

alessey:
nit: when you tab to the wallet and open, should focus in the island and trap focus there

src/wallet/components/WalletIslandQrReceive.tsx
abcrane123: +1 can you add a comment for this?

src/wallet/components/WalletIslandSwap.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const handleCloseSwap = useCallback(() => {
+    setIsSwapClosing(true);
+
+    setTimeout(() => {
+      setShowSwap(false);
+    }, 200);
+
+    setTimeout(() => {
+      setIsSwapClosing(false);
+    }, 400);
```
abcrane123, line 43: can you add a comment about what is happening here with the staggered set timeouts ?

src/wallet/constants.ts
```diff
@@ -10,3 +13,25 @@ export const ERC_1967_PROXY_IMPLEMENTATION_SLOT =
...
+  },
+  {
+    name: 'USDC',
+    address: '0x833589fcd6edb6e08f4c7c32d4f71b54bda02913',
+    symbol: 'USDC',
+    decimals: 6,
+    image:
+      'https://d3r81g40ycuhqg.cloudfront.net/wallet/wais/44/2b/442b80bd16af0c0d9b22e03a16753823fe826e5bfd457292b55fa0ba8c1ba213-ZWUzYjJmZGUtMDYxNy00NDcyLTg0NjQtMWI4OGEwYjBiODE2',
+    chainId: base.id,
+  },
```
abcrane123, line 36: nit but I think these tokens are already defined in `token/constants.ts` file so you can probably just import here

src/internal/components/PressableIcon.tsx
brendan-defi: ah, sure, i'll scoot the classnames down to the button and nix the div

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
brendan-defi: good call, will make the change

src/internal/svg/qrIconSvg.tsx
brendan-defi: thanks, updating

src/wallet/components/WalletIslandDraggable.tsx
brendan-defi: will consult her again on it, but this was Tina's instruction

alessey: if that's the spec, then it works for me

src/wallet/components/WalletIslandAddressDetails.tsx
```diff
@@ -0,0 +1,92 @@
...
+    useWalletIslandContext();
+
+  if (isFetchingPortfolioData) {
+    return <Spinner />;
+  }
+
+  return (
+    <span data-testid="ockWalletIsland_AddressBalance">
+      {portfolioFiatValue && `$${Number(portfolioFiatValue)?.toFixed(2)}`}
+    </span>
```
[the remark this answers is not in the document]
brendan-defi: per convo, going to add nullish check above, remove the nullish coalescence here

src/wallet/constants.ts
brendan-defi: oh nice, thank you!

src/wallet/components/WalletIslandQrReceive.tsx
brendan-defi: hmm, good call, let me try to refactor this to use `onAnimationEnd` instead

src/wallet/components/WalletIslandQrReceive.tsx
brendan-defi: good call, will add to my fast-follows

src/wallet/components/WalletIslandSwap.tsx
brendan-defi: good call, this is icky, i'm going to refactor to be based on onAnimationEnd instead

brendan-defi:
> nit: when you tab to the wallet and open, should focus in the island and trap focus there

I actually had that, and was asked to remove. I'll start a convo about aligning on an approach for focus / trapping focus.

src/wallet/components/WalletIslandAddressDetails.tsx
```diff
@@ -0,0 +1,97 @@
...
+  if (portfolioFiatValue === null) {
+    return null;
+  }
+
+  return (
+    <div
+      className={cn(text.title1, 'mt-1 font-normal')}
+      data-testid="ockWalletIsland_AddressBalance"
+    >
+      {`$${Number(portfolioFiatValue)?.toFixed(2)}`}
```
alessey, line 94: nit: unnecessary template string

src/wallet/components/WalletIslandTokenHoldings.tsx
```diff
@@ -0,0 +1,75 @@
...
+          <span className={cn(text.label1, color.foreground)}>
+            {token.name}
+          </span>
+          <span className={cn(text.legal, color.foregroundMuted)}>
+            {`${balance.toFixed(5)} ${token.symbol}`}
+          </span>
+        </div>
+      </div>
+      <span className={cn(text.label2, color.foregroundMuted)}>
+        {`$${valueInFiat.toFixed(2)}`}
```
alessey, line 71: nit: unnecessary template strings

alessey (approved):

src/wallet/components/WalletIslandAddressDetails.tsx
brendan-defi: yeah, i wouldn't have done it this way, but our linter says so:

image

src/wallet/components/WalletIslandTokenHoldings.tsx
brendan-defi: see above, our linter says to use template strings over string concatenation

src/wallet/components/WalletIslandAddressDetails.tsx
alessey: 🤷

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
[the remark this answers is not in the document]
dschlabach: You shouldn't need to handle that part yourself! What you have here is enough! Behind the scenes, Tanstack Query will hash the two items in this array and come up with your key

```
    queryKey: ['usePortfolioTokenBalances', address],

```

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export function usePortfolioTokenBalances({
+  address,
+}: {
+  address: Address | undefined | null;
+}): UseQueryResult<PortfolioTokenBalances> {
+  const actionKey = `usePortfolioTokenBalances-${address}`;
+  return useQuery({
+    queryKey: ['usePortfolioTokenBalances', actionKey],
+    queryFn: async () => {
```
dschlabach, line 15: The new way of doing this in v5 is by using `skipToken` to keep typesafety

So something like !!address ? (// do your fetch) : skipToken

https://tanstack.com/query/latest/docs/framework/react/guides/disabling-queries#typesafe-disabling-of-queries-using-skiptoken

Would be nice to keep it full typesafe, it should be fine how you have it here as well

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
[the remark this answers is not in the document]
brendan-defi: TIL, thank you! will make the change

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
brendan-defi: per convo, it seems like `skipToken` doesn't support `refetch`, which i leverage, so unfortunately can't use :(

cpcramer:
This is a little strange - for `WalletIslandDraggable` the `ConnectWallet` button is draggable 

https://screen.studio/share/Cd1O7ycK

src/internal/svg/cbwSvg.tsx
```diff
@@ -5,7 +5,7 @@ export const cbwSvg = (
     viewBox="0 0 1024 1024"
     xmlns="http://www.w3.org/2000/svg"
   >
-    <title>Wallet Icon</title>
+    <title>Wallet</title>
```
cpcramer, line 8: nit: update filename and const to `coinbaseWalletSvg`

src/internal/svg/clockSvg.tsx
```diff
@@ -0,0 +1,18 @@
+import { background, icon } from '../../styles/theme';
+
+export const clockSvg = (
+  <svg
+    width="28"
+    height="28"
```
cpcramer, line 6: Saw Adam had a similar comment but should default towards 100% width and height

src/wallet/components/WalletIslandProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+  });
+
+  return (
+    <WalletIslandContext.Provider value={value}>
+      {children}
+    </WalletIslandContext.Provider>
+  );
+}
+
+function getAnimations(isClosing: boolean, showSubComponentAbove: boolean) {
```
cpcramer, line 68: Hmm this function seems out of place

cpcramer: Might be better in `utilities/`

src/internal/svg/clockSvg.tsx
brendan-defi: yeah, i'm going to punt that to the SVG standardization that I'll do next week

src/internal/components/Draggable.tsx
```diff
@@ -9,15 +10,27 @@ type DraggableProps = {
...
+  const [cursorDisplay, setCursorDisplay] = useState('default');
+  const isModalOpen = useIsModalOpen();
+
+  useEffect(() => {
+    if (isModalOpen) {
+      setCursorDisplay('default');
+      return;
+    }
+
+    setCursorDisplay(isDragging ? 'cursor-grabbing' : 'cursor-grab');
```
alessey, line 32: do you need cursor state?  could you just base cursor off isDragging?

src/internal/hooks/useIsModalOpen.ts
```diff
@@ -0,0 +1,62 @@
...
+    subscriberCount: subscribers.size,
+    hasObserver: !!observer,
+  }),
+  reset: () => {
+    subscribers.clear();
+    if (observer) {
+      observer.disconnect();
+      observer = null;
+    }
+  },
```
alessey, line 61: since this just relies on the exported subscribers and observer would it be possible to move to your test file?

alessey:
lgtm, as follow-ups, focus trap and escape to close would be nice adds.

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
```diff
@@ -0,0 +1,41 @@
...
+          portfolioBalanceUsd: 0,
+          tokenBalances: [],
+        };
+      }
+
+      return response.portfolios[0];
+    },
+    retry: false,
+    enabled: !!address,
+    refetchOnWindowFocus: false,
```
dschlabach, line 35: How come we don't want to `refetchOnWindowFocus`?

src/internal/components/QrCode/useMatrix.ts
```diff
@@ -2,16 +2,18 @@ import QRCode from 'qrcode';
 import { useMemo } from 'react';
 
 export function useMatrix(
```
dschlabach, line 4: Could you add some comments above this hook explaining what it's doing?

src/internal/svg/baseScanSvg.tsx
```diff
@@ -0,0 +1,46 @@
...
+    viewBox="0 0 155 155"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+    className={icon.foreground}
+  >
+    <title>BaseScan</title>
+    <g clipPath="url(#clip0_13027_2497)">
+      <mask
+        id="mask0_13027_2497"
+        // style="mask-type:luminance"
```
dschlabach, line 16:
```suggestion

```

src/wallet/components/WalletIslandTransactionActions.tsx
```diff
@@ -0,0 +1,113 @@
...
+      'defaultPaymentMethod=CRYPTO_ACCOUNT',
+      'presetFiatAmount=25',
+    ];
+
+    window.open(
+      `${baseUrl}?${params.join('&')}`,
+      'popup',
+      'width=400,height=600,scrollbars=yes',
+    );
+  }, [address, chain.name, projectId]);
```
dschlabach, line 37: I'd suggest doing it like this - it makes it easier to read because it's more declarative, plus it handles the URL encoding for you.

```suggestion
 const handleBuy = useCallback(() => {
  const url = new URL('https://pay.coinbase.com/buy/select-asset');
  const params = new URLSearchParams({
    appId: projectId,
    destinationWallets: JSON.stringify([{
      address,
      blockchains: [chain.name.toLowerCase()]
    }]),
    defaultAsset: 'USDC',
    defaultPaymentMethod: 'CRYPTO_ACCOUNT',
    presetFiatAmount: '25'
  });

  window.open(`${url}?${params}`, 'popup', 'width=400,height=600,scrollbars=yes');
}, [address, chain.name, projectId]);
```

src/internal/components/Draggable.tsx
brendan-defi: Also need to know if the modal is open or not, otherwise when the modal is open everywhere on the screen is cursor-grab which looks weird

src/internal/hooks/useIsModalOpen.ts
brendan-defi: oh sorry, i actually meant to stop exporting observer and subscribers. felt weird to export them.

what do you think, better to export and have the test fn in the test file, or not export and have the test fn in the hook file?

src/core-react/wallet/hooks/usePortfolioTokenBalances.ts
brendan-defi: i don't feel strongly about it, but felt redundant with refetch interval

src/internal/components/QrCode/QrCodeSvg.tsx
```diff
@@ -66,20 +66,20 @@ export function QrCodeSvg({
...
+    ] ?? 'blue';
   const linearColors = [
     linearGradientStops[linearGradientColor].startColor,
     linearGradientStops[linearGradientColor].endColor,
   ];
 
+  const radialGradientColor =
+    ockThemeToRadialGradientColorMap[
+      themeName as keyof typeof ockThemeToLinearGradientColorMap
+    ] ?? 'default';
```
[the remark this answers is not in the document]
brendan-defi: good call, thanks

src/internal/components/QrCode/QrCodeSvg.tsx
```diff
@@ -66,20 +66,20 @@ export function QrCodeSvg({
   const linearGradientColor =
     ockThemeToLinearGradientColorMap[
       themeName as keyof typeof ockThemeToLinearGradientColorMap
-    ];
+    ] ?? 'blue';
```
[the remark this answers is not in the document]
brendan-defi: updating, thanks

src/internal/hooks/useIsModalOpen.ts
alessey: i think exporting to use in a test file is preferable to having test code in your component.  you wouldn't be exporting those externally anyways.

src/internal/svg/baseScanSvg.tsx
brendan-defi: deleting

src/wallet/components/WalletIslandTransactionActions.tsx
brendan-defi: nice, i like that, thank you

playground/nextjs-app-router/components/demo/WalletAdvancedDefault.tsx
```diff
@@ -0,0 +1,11 @@
+import { WalletAdvancedDefault } from '@coinbase/onchainkit/wallet';
+
+export default function WalletAdvancedDefaultDemo() {
+  return (
+    <div className="mx-auto">
+      <div className="flex justify-end">
```
dschlabach, line 6: nit but do you need this extra div?

src/internal/components/Draggable.tsx
```diff
@@ -1,23 +1,38 @@
+'use client';
```
dschlabach, line 1: Will be testing our updated build system this week some more, but might want to make sure this doesn't break anything when shipped as is

src/internal/components/Draggable.tsx
```diff
@@ -1,23 +1,38 @@
...
+  draggingDisabled?: boolean;
 };
 
-export default function Draggable({
+export function Draggable({
   children,
   gridSize = 1,
   startingPosition = { x: 20, y: 20 },
   snapToGrid = false,
+  draggingDisabled = false,
```
dschlabach, line 20: Could see the argument for just calling it `disabled` since it's less verbose

src/internal/components/Draggable.tsx
dschlabach: Shouldn't that be solvable in CSS? i.e. with the`hover:` pseudo class you can limit the cursor change to just the Draggable, not the entire screen?

src/wallet/utils/getDraggableWalletDefaultStartingPosition.ts
```diff
@@ -0,0 +1,12 @@
+'use client';
+
+export function getDraggableWalletDefaultStartingPosition() {
```
dschlabach, line 3: I think you could just call this something like `getDefaultStartingPosition()` to be honest, it's a bit too hard to read as is

src/wallet/utils/getDraggableWalletDefaultStartingPosition.ts
```diff
@@ -0,0 +1,12 @@
+'use client';
```
dschlabach, line 1:
```suggestion
```

dschlabach (dismissed):

playground/nextjs-app-router/components/demo/WalletAdvancedDefault.tsx
brendan-defi: no, removing and combining the classes

src/internal/components/Draggable.tsx
brendan-defi: will touch base with you tomorrow to double-check, but i've used a whole bunch of the other playground components and they seem unaffected

src/internal/components/Draggable.tsx
brendan-defi: makes sense, changing

src/wallet/utils/getDraggableWalletDefaultStartingPosition.ts
brendan-defi: so i know it's a little hard to read, and it's a little verbose. but the reason i have all of those is:

- it's not just the draggable starting position — other draggable components may make sense to have other defaults
- it's not the draggable wallet starting position — if devs pass in a different value, that is the starting position
- so it's specifically the `DraggableWalletDefault` 

I think we should keep this one as is

src/wallet/utils/getDraggableWalletDefaultStartingPosition.ts
brendan-defi: ack, changing

src/internal/components/Draggable/useRespositionOnResize.ts
```diff
@@ -0,0 +1,76 @@
...
+    repositionDraggable,
+    resetPosition,
+    isElementInViewport,
+  ]);
+
+  useEffect(() => {
+    window.addEventListener('resize', handleWindowResize);
+    return () => window.removeEventListener('resize', handleWindowResize);
+  }, [handleWindowResize]);
+}
```
alessey, line 76: can you repurpose this hook to prevent the draggableRef from being dragged outside of the screen bounds at all and then just listen to resize as part of that?

src/internal/svg/backArrowSvg.tsx
```diff
@@ -0,0 +1,17 @@
...
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <title>Back Arrow</title>
+    <path
+      d="M5.54479 12.077L6.46623 11.2323L2.42062 6.81894H11.9999V5.56894H2.50874L6.88374 1.19394L5.99985 0.310059L0.134766 6.17515L5.54479 12.077Z"
+      className={icon.foreground}
+    />
+  </svg>
+);
```
alessey, line 17: nit: once we have a single arrow we should favor rotating it with css rather than including additional arrows

src/internal/svg/clockSvg.tsx
```diff
@@ -0,0 +1,18 @@
...
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <title>Clock</title>
+    <rect width="28" height="28" rx="8" className={background.default} />
+    <path
+      d="M14 20C17.3137 20 20 17.3137 20 14C20 10.6863 17.3137 8 14 8C10.6863 8 8 10.6863 8 14C8 17.3137 10.6863 20 14 20ZM13 10H14.5V13.5H17V15H13V10Z"
+      className={icon.foreground}
+    />
+  </svg>
+);
```
alessey, line 18: nit: is this still needed?

src/ui/react/identity/components/Avatar.tsx
```diff
@@ -87,7 +87,9 @@ export function Avatar({
             {...props}
           />
         ) : (
-          defaultAvatar
+          <div className={cn(border.default, 'h-full w-full border')}>
+            {defaultAvatar}
+          </div>
```
alessey, line 92: adding a 1px border around defaultAvatar means that avatars display at 40x40 and default avatars render at 38x38 (i.e. 40 - 1px top + 1px left)

src/wallet/components/Wallet.tsx
```diff
@@ -1,53 +1,158 @@
...
+      <WalletContent
+        className={cn(componentTheme, className)}
+        {...(draggable
+          ? {
+              draggable,
+              draggableStartingPosition:
+                draggableStartingPosition ??
+                getDraggableWalletDefaultStartingPosition(),
+            }
+          : { draggable })}
```
alessey, line 40: nit: this may be more readable extracted to a function and then spreading the result here

src/wallet/utils/getDraggableWalletDefaultStartingPosition.ts
alessey: its in the wallet directory which implies its for wallet.  you could easily get away with a const where this is consumed
`
const defaultPosition = {
  x: window.innerWidth ? window.innerWidth - 125 : 100,
  y: window.innerHeight ? window.innerHeight - 125 : 100,
}`

might make sense to use percents to deal with differences in screen sizes.  this default position on an iphone pro puts the island modal off the screen.

src/internal/components/Draggable/Draggable.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const [cursorDisplay, setCursorDisplay] = useState('default');
+  const draggableRef = useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    if (disabled) {
+      setCursorDisplay('default');
+      return;
+    }
+
+    setCursorDisplay(isDragging ? 'cursor-grabbing' : 'cursor-grab');
```
dschlabach, line 36: Are you sure you need this state? Are you able to use pseudo classes (`hover:`, `active:`) to only show the grabbing on the draggable element, rather than window-wide?

e.g.: `active:cursor-grabbing hover:cursor-grab`

cpcramer:
Can we center `WalletAdvancedDefault` and `WalletIsland` (vertically and horizontally) in the playground?

cpcramer:
I feel like we should hide tokens with a balance of $0.00 - I have a ton of them in my wallet and it feels like spam 

Screenshot 2025-01-14 at 8 58 14 AM

src/ui/react/identity/components/Avatar.tsx
brendan-defi: Curious if you have a better way to do this. Without the border, I get a 1px outline coming from the background:
image

 I tried (and paired with some others to try) to fix without a border, but this was the best that we came up with

alessey:
> I feel like we should hide tokens with a balance of $0.00 - I have a ton of them in my wallet and it feels like spam
> 
> Screenshot 2025-01-14 at 8 58 14 AM

just the fiat balance is < 0, in this case I think it'd make sense to show < $0.01 or something similar

src/internal/components/Draggable/Draggable.tsx
alessey: if pseudo elements don't work you still shouldn't need state as this is just a derived property

cpcramer:
> > I feel like we should hide tokens with a balance of $0.00 - I have a ton of them in my wallet and it feels like spam
> > Screenshot 2025-01-14 at 8 58 14 AM
> 
> just the fiat balance is < 0, in this case I think it'd make sense to show < $0.01 or something similar

Ok and looks like this is what CB Wallet defaults to

brendan-defi:
> I'm a little confused by these changes. Seems like we are just creating a bunch of *Default components. I thought the goal with this component was to be able to compose the version you wanted. 

We ended up creating some defaults, but all of the underlying components are composable, so someone could definitely do 
```
<Wallet draggable={true/false}> // whole thing can be draggable or not, works nicely either way
  <ConnectWallet>
    ...custom ConnectWallet children
  </ConnectWallet>
  <WalletAdvanced>
    ...custom WalletIsland children
  </WalletAdvanced>
</Wallet>
```
 
> Additionally, drag seems like a much more useful feature (with a more straightforward implementation) if it only applied to the WalletIsland or WalletDropdown child vs. the entire component.

This is an implementation decision. A dev _could_ do this:
```
<Wallet draggable={false}>
  <ConnectWallet>
    ...custom ConnectWallet children
  </ConnectWallet>
  <Draggable>
    <WalletAdvanced>
      ...custom WalletAdvanced children
    </WalletAdvanced>
  </Draggable>
</Wallet>
```

This would leave a ConnectWallet "anchor", with a draggable dropdown/advanced box.

brendan-defi:
> Can we center `WalletAdvancedDefault` and `WalletIsland` (vertically and horizontally) in the playground?

No. WalletAdvancedDefault is tall enough that on normal laptop screens it overflows the bottom of the screen when vertically centered. 

And WalletIsland is meant to be defaulted to bottom-left, so that's the playground implementation.

brendan-defi:
> I feel like we should hide tokens with a balance of $0.00 - I have a ton of them in my wallet and it feels like spam

We are using the Coinbase Wallet spam filter already, so these tokens are below the "likely spam" threshold. And we hide tokens with a zero-crypto balance. But we chose to keep tokens with a crypto balance but no fiat balance because some tokens aren't super liquid on dexes and don't have fiat prices, but would still be valuable to be seen.

src/internal/components/Draggable/Draggable.tsx
brendan-defi: I will investigate this as a fast-follow, but I think the answer is no. 

The issue is that Draggable considers everything in the wallet DOM tree to be part of it's system. Modal is part of the wallet DOM tree, and has a full-screen overlay. So draggable then considers the entire screen to be "part of the draggable component".

src/ui/react/identity/components/Avatar.tsx
alessey: this is because of the way the svg path is drawn, you are seeing the bleed from the edge of the path (which is styled to be white), you may be able to fix this by separating into a path for the white part and a circle behind it in the inverse.  you could apply the border hack with a transparent border to the users img but you'd still end up changing the size from 40 to 38 unless you adjust the sizes to a non-standard tailwind size.

alessey (approved):

src/ui/react/identity/components/Avatar.tsx
brendan-defi: gotcha, okay, i will try to fix this up as a fast-follow. thanks!

src/internal/components/Draggable/Draggable.tsx
brendan-defi: with Paul's refactor of WalletModal, I can take this on now.

src/internal/svg/clockSvg.tsx
brendan-defi: nope, deleting

src/internal/components/Draggable/useRespositionOnResize.ts
brendan-defi: refactored a few things here

src/wallet/utils/getDraggableWalletDefaultStartingPosition.ts
brendan-defi: renamed, moved to Draggable directory so it can be reused by anyone using draggable, and refactored to offset a pct not fixed pixels

PR #1797, opened 2025-01-08
chore: `Buy` component touch ups

abcrane123:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/internal/svg/closeSvg.tsx
+++ b/src/internal/svg/closeSvg.tsx
@@ -1,18 +1,24 @@
 import { icon } from '../../styles/theme';
 
-export const closeSvg = (
-  <svg
-    aria-label="ock-closeSvg"
-    width="12"
-    height="12"
-    viewBox="0 0 16 16"
-    fill="none"
-    xmlns="http://www.w3.org/2000/svg"
-  >
-    <title>Close SVG</title>
-    <path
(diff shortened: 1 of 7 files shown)


src/internal/svg/closeSvg.tsx
```diff
@@ -1,18 +1,24 @@
...
+      height="12"
+      viewBox="0 0 16 16"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <title>Close</title>
+      <path
+        d="M2.14921 1L1 2.1492L6.8508 8L1 13.8508L2.1492 15L8 9.1492L13.8508 15L15 13.8508L9.14921 8L15 2.1492L13.8508 1L8 6.8508L2.14921 1Z"
+        className={className}
+      />
```
dschlabach, line 21: Hmmm this seems a little counterintuitive that it'd be passed to the path instead of the svg.

@brendan-defi what is your progress on standardizing our treatment of SVGs? My concern here is if we end up standardizing around className on <svg /> to control things like size, color, etc., then this won't work

brendan-defi: still WIP, but yeah i'm thinking classname on svg, which probably won't play nicely here.

dschlabach (approved):

brendan-defi (approved):

PR #1799, opened 2025-01-09
fix: set client boundary in `Buy`, `Checkout`, `Fund`, and `OnchainkitProvider`

dschlabach:
**What changed? Why?**
Continuing #1771 with support for `Buy` and `OnchainkitProvider`

**Notes to reviewers**
- Consumers will still need to set a client boundary manually with `OnchainkitProvider` because of the way that the `chain` prop from `viem` is handled.
- Only setting directive on `Buy` since it's the only exported Buy component (no subcomponents)

**How has it been tested?**
- Locally with Next 15 and Vite

### DIFF
--- a/src/fund/index.ts
+++ b/src/fund/index.ts
@@ -1,4 +1,8 @@
+// 🌲☀🌲
+// Components
 export { FundButton } from './components/FundButton';
+
+// Utils
 export { getCoinbaseSmartWalletFundUrl } from './utils/getCoinbaseSmartWalletFundUrl';
 export { getOnrampBuyUrl } from './utils/getOnrampBuyUrl';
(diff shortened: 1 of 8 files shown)


src/fund/index.ts
```diff
@@ -1,4 +1,8 @@
+// 🌲☀🌲
+// Components
```
dschlabach, line 2: Happy to revert this organization structure if people aren't fans. Thought it improves readability when we have components, hooks, utils, and types

alessey (approved):

cpcramer (approved):

alessey (approved):

src/fund/index.ts
brendan-defi: yeah that's nice

PR #1800, opened 2025-01-09
fix: set client boundary for `Swap` and `Token` components

dschlabach:
**What changed? Why?**

**Notes to reviewers**
- `findComponent` has been updated to support RSC payloads from Next. This is only required if the user attempts to compose `Swap` with subcomponents (because the children the component receive look slightly different). This is a temporary patch until we migrate to default children.

**How has it been tested?**


src/token/components/TokenChip.tsx
```diff
@@ -1,10 +1,12 @@
...
 import { background, cn, pressable, text } from '../../styles/theme';
 import type { TokenChipReact } from '../types';
 import { TokenImage } from './TokenImage';
 
 /**
- * Small button that display a given token symbol and image.
+ * Small button that displays a given token symbol and image.
  *
  * WARNING: This component is under development and
  *          may change in the next few weeks.
```
alessey, line 12: 😆

alessey (dismissed):

brendan-defi (dismissed):

alessey (dismissed):

abcrane123 (approved):

brendan-defi (approved):

PR #1806, opened 2025-01-13
Amount Preset inputs

rustam-cb:
**What changed? Why?**
Added amount preset inputs for quick input options

PRODUCT REQUIREMENTS: 
- Devs can provide preset amounts (should be exactly 3)
- Preset amounts are only for fiat (similar to how it works on most other platforms)

https://github.com/user-attachments/assets/a35a178c-605b-4172-81f6-c61f8686f8bd

**Notes to reviewers**

**How has it been tested?**


src/fund/hooks/useFundCardSetupOnrampEventListeners.test.tsx
```diff
@@ -12,17 +12,29 @@ import {
...
+  coinbaseFee: { value: '2.00', currency: 'USD' },
+  networkFee: { value: '1.00', currency: 'USD' },
+  quoteId: 'quote-id-123',
+};
 
 describe('useFundCardSetupOnrampEventListeners', () => {
   beforeEach(() => {
     vi.clearAllMocks();
     vi.useFakeTimers();
+    (fetchOnrampQuote as Mock).mockResolvedValue(mockResponseData);
```
rustam-cb, line 37: Flaky tests fix

src/fund/components/AmountInputSnippet.tsx
```diff
@@ -0,0 +1,55 @@
+import { useCallback, useMemo } from 'react';
+import { border, cn, color, text } from '../../styles/theme';
```
brendan-defi, line 2: nit: we prefer path aliasing `@/styles` over relative paths. this is a new thing so not fully updated backwards-facing, but on new files we use aliases

src/fund/components/AmountInputSnippet.tsx
```diff
@@ -0,0 +1,55 @@
...
+}: AmountInputSnippetPropsReact) {
+  const snippetText = useMemo(() => {
+    return `${amountInputSnippet.value} ${currencyOrAsset}`;
+  }, [amountInputSnippet, currencyOrAsset]);
+
+  const handleClick = useCallback(() => {
+    onClick(amountInputSnippet);
+  }, [amountInputSnippet, onClick]);
+
+  const handleKeyPress = useCallback(
```
brendan-defi, line 18: nice

playground/nextjs-app-router/components/demo/FundCard.tsx
```diff
@@ -1,10 +1,39 @@
...
+      value: '50',
+      type: 'fiat' as AmountInputTypeReact,
+    },
+  ];
   return (
     <div className="mx-auto grid w-[500px] gap-8">
       <FundCard
         assetSymbol="ETH"
         country="US"
+        amountInputSnippets={amountInputSnippets}
```
alessey, line 36: can we call these presetInputs (i'm open to suggestions) snippet makes me think of small pieces of information, but these are like presetInputs or quickInputs or something

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -99,13 +114,27 @@ export const FundCardAmountInput = ({
...
+   * If the selected input type is 'crypto', we only want to display the snippets that have a type of 'crypto'.
+   * i.e [0.1 ETH] [0.2 ETH] [0.3 ETH]
+   */
+  const filteredAmountInputSnippets = useMemo(
+    () =>
+      amountInputSnippets?.filter(
+        (snippet) => snippet.type === selectedInputType,
+      ),
+    [amountInputSnippets, selectedInputType],
+  );
```
alessey, line 131: should we also filter out duplicates?

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -125,12 +154,22 @@ export const FundCardAmountInput = ({
...
-        <FundCardCurrencyLabel
-          ref={currencySpanRef}
-          label={selectedInputType === 'crypto' ? asset : currencyLabel}
-        />
+        <FundCardCurrencyLabel ref={currencySpanRef} label={currencyOrAsset} />
       </div>
 
+      {!value && (
+        <div className="flex w-[100%] flex-wrap items-center justify-end">
+          {filteredAmountInputSnippets?.map((snippet) => (
```
alessey, line 162: may want to remove this div if !filteredAmountInputSnippets

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -125,12 +154,22 @@ export const FundCardAmountInput = ({
...
         />
 
-        <FundCardCurrencyLabel
-          ref={currencySpanRef}
-          label={selectedInputType === 'crypto' ? asset : currencyLabel}
-        />
+        <FundCardCurrencyLabel ref={currencySpanRef} label={currencyOrAsset} />
       </div>
 
+      {!value && (
```
alessey, line 160: is this so the input has more room?  It might be nice to allow someone to click on one then another one without ever using the input, maybe they should go below the input or something?

src/fund/types.ts
```diff
@@ -359,8 +359,22 @@ export type LifecycleEvents = {
...
+  currencyOrAsset: string;
+  onClick: (snippet: AmountInputSnippetReact) => void;
+};
+
+export type AmountInputSnippetReact = {
+  value: string;
+  type: AmountInputTypeReact;
+};
+
+export type AmountInputTypeReact = 'fiat' | 'crypto';
```
alessey, line 376: i think the type naming convention that's been used is to only append React to types of top level components, but not necessarily their individual props

playground/nextjs-app-router/components/demo/FundCard.tsx
rustam-cb: how about `presetAmountInputs`. To be explicit about what those inputs are

src/fund/types.ts
```diff
@@ -359,8 +359,22 @@ export type LifecycleEvents = {
   onError?: (e: OnrampError | undefined) => void;
   onStatus?: (lifecycleStatus: LifecycleStatus) => void;
   onSuccess?: (result: SuccessEventData) => void;
+  amountInputSnippets?: AmountInputSnippetReact[];
```
alessey, line 362: shouldn't these live in the provider?

playground/nextjs-app-router/components/demo/FundCard.tsx
alessey: yea, that sounds good to me

src/fund/components/FundCardAmountInput.tsx
rustam-cb: I think we should leave this up to the consumer. 
Its kind of an edge case that is up to the developer to handle and on our side its not obvious we would filter out duplicates.

src/fund/components/FundCardAmountInput.tsx
rustam-cb: yes, that is done to provide more room for the input. I''ll check with the design.

src/fund/components/FundCardAmountInput.tsx
alessey: that works, but you'll need to update your key in the map below to ensure its unique with duplicates

src/fund/types.ts
rustam-cb: yes, good catch. Must be bad rebase.

src/fund/components/AmountInputSnippet.tsx
rustam-cb: Updated

playground/nextjs-app-router/components/demo/FundCard.tsx
rustam-cb: DOne

src/fund/components/FundCardAmountInput.tsx
rustam-cb: Set it to index (In this case using index as a key is safe)

src/fund/components/FundCardAmountInput.tsx

src/fund/components/FundCardAmountInput.tsx
rustam-cb: I have reached out to design. How about we make this change in a follow up PR in case we decide to make the change in the design?

src/fund/types.ts
rustam-cb: Updated

src/fund/types.ts
rustam-cb: Updated

src/fund/index.ts
```diff
@@ -20,5 +20,5 @@ export type {
   GetOnrampUrlWithProjectIdParams,
   GetOnrampUrlWithSessionTokenParams,
   OnrampError,
-  PaymentMethodReact,
+  PaymentMethod as PaymentMethodReact,
```
brendan-defi, line 23: do we need to preserve `PaymentMethodReact`?

rustam-cb: No seems like it got auto-updated with the rename. 
Removed it altogether since there is no use of it outside of the component.

brendan-defi: cool thanks

brendan-defi (approved):

rustam-cb (commented):
Not merging for now. Need to make some design adjustments

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -39,50 +32,20 @@ export const FundCardAmountInput = ({
...
-        Number(fiatValue) * Number(exchangeRate),
-      );
-      const resultCryptoValue = truncateDecimalPlaces(calculatedCryptoValue, 8);
-      setFundAmountCrypto(
-        calculatedCryptoValue === '0' ? '' : resultCryptoValue,
-      );
-    },
-    [exchangeRate, setFundAmountFiat, setFundAmountCrypto],
-  );
+  const { handleChange } = useAmountInput();
```
rustam-cb, line 35: Moved code into `useAmountInput` hook to reuse it from outside of the amount input component

src/fund/types.ts
```diff
@@ -382,3 +386,18 @@ export type LifecycleStatus =
...
+};
+
+/**
+ * To use this type, you must provide a tuple of strings with a length of 3.
+ *
+ * Example:
+ *
+ * const presetAmountInputs: PresetAmountInputs = ['100', '200', '300'];
+ */
+export type PresetAmountInputs = readonly [string, string, string];
```
alessey, line 403: why does this require 3?

src/fund/hooks/useAmountInput.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  return useMemo(
+    () => ({
+      handleChange,
+      handleFiatChange,
+      handleCryptoChange,
+    }),
+    [handleChange, handleFiatChange, handleCryptoChange],
+  );
+};
```
alessey, line 65: this seems strange to extract since its so tightly coupled to this implementation, how are you planning to reuse it?

src/fund/types.ts
rustam-cb: @alessey  We only want to have 3 items listed (so that it looks the way we want it on the UI). We don't want to have less than or more than 3 items on the Ui

src/fund/hooks/useAmountInput.ts
rustam-cb: Extracted it since I'm using it to control the amount input from outside of the amount input component (from the preset amounts component)

src/fund/types.ts
rustam-cb: @alessey  Wrapping is not hard to handle. It is a design decision to not wrap the preset input amounts and have 3 items or nothing. 
Though I have not tested that way but nothing stops it from working with only preset amounts.

src/fund/components/FundCardPresetAmountInputList.tsx
```diff
@@ -0,0 +1,32 @@
...
+  const { handleFiatChange } = useAmountInput();
+
+  if (!presetAmountInputs) {
+    return null;
+  }
+
+  return (
+    <div
+      data-testid="ockPresetAmountInputList"
+      className="flex w-[100%] flex-wrap items-center justify-between gap-2 pt-8"
```
dschlabach, line 19:
```suggestion
      className="flex w-full flex-wrap items-center justify-between gap-2 pt-8"
```

rustam-cb: Updated

alessey (approved):

dschlabach (approved):

PR #1814, opened 2025-01-14
fix: modal state local to connect wallet

brendan-defi:
**What changed? Why?**
* duplicated connect modal state in ConnectWallet
  * fixes issue where ConnectWallet not always wrapped in WalletProvider. in these cases, modal is unable to be displayed because the handlers come from context
  * by adding state local to this component, we can use it in a disconnected state without wrapping in a a WalletProvider

**Notes to reviewers**

**How has it been tested?**
locally

### DIFF
--- a/src/wallet/components/ConnectWallet.tsx
+++ b/src/wallet/components/ConnectWallet.tsx
@@ -41,7 +41,7 @@ export function ConnectWallet({
 
   // State
   const [hasClickedConnect, setHasClickedConnect] = useState(false);
-  const [isModalOpen, setIsModalOpen] = useState(false);
+  const [isModalOpen, setIsModalOpen] = useState(false); // duplicate modal state because ConnectWallet not always within WalletProvider
 
   // Get connectWalletText from children when present,
   // this is used to customize the connect wallet button text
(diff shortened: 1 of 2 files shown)


src/wallet/components/ConnectWallet.tsx
```diff
@@ -31,7 +31,6 @@ export function ConnectWallet({
 
   // Core Hooks
   const {
-    isConnectModalOpen,
     setIsConnectModalOpen,
     isSubComponentOpen,
```
rustam-cb, line 35: Won't this destructuring fail if there is no context? 
If `useWalletContext` returns undefined, then destructuring of `undefined` should fail

alessey (dismissed):

dschlabach (approved):

abcrane123 (approved):

src/wallet/components/ConnectWallet.tsx
```diff
@@ -42,6 +41,7 @@ export function ConnectWallet({
 
   // State
   const [hasClickedConnect, setHasClickedConnect] = useState(false);
+  const [isModalOpen, setIsModalOpen] = useState(false);
```
rustam-cb: If we want to make the provider optional, then would it be more cleaner to wrap this whole component in another container component and have provider initialized if there is no other provider. 

```
 function ConnectWallet() {
  const walletContext = useWalletContext();

    if(!walletContext) {
        return <Wallet> <ConnectWalletContent> <Wallet>;
    }

return <ConnectWaleetContent />;
}

function ConnetWalletContent() {

}
```

alessey (approved):

src/wallet/components/ConnectWallet.tsx
alessey: wallet context will be {} in this case

brendan-defi: all of the destructured values will be undefined, but if they are not accessed they don't throw. this is why all of the context-based-state-updaters are accessed optionally, eg. `setIsConnectModalOpen?.(false)` or `handleClose?.()`

src/wallet/components/ConnectWallet.tsx
brendan-defi: interesting idea, will think if something like this makes sense. in the interest of fixing the bug, going to ship this version

PR #1818, opened 2025-01-15
Fund Card Payment Methods

rustam-cb:
**What changed? Why?**
`US Should have 3 Payment methods`
- Coinbase
- Debit card
- Apple pay (If apple pay is supported on the device)

`Other countries`
- Coinbase

We are using `options` API to get data about `min` `max` amounts

**Notes to reviewers**

**How has it been tested?**

|Payment options Loading state|
|---------------|
|Screenshot 2025-01-21 at 9 59 27 AM|


src/fund/components/FundCardPaymentMethodDropdown.tsx
```diff
@@ -32,15 +32,35 @@ export function FundCardPaymentMethodDropdown({
...
+    (method: PaymentMethod) => {
+      const amount = Number(fundAmountFiat);
+
+      if (method.minAmount && amount < method.minAmount) {
+        return `Minimum amount of $${method.minAmount} required`;
+      }
+
+      if (method.maxAmount && amount > method.maxAmount) {
+        return `Maximum amount allowed is $${method.maxAmount}`;
+      }
```
alessey, line 59: could add a return undefined here and your function would have a consistent return statement and not need to be conditional below.  could also reference this function in isPaymentMethodDisabled to avoid duplicating logic

src/fund/components/FundCardPaymentMethodDropdown.tsx
```diff
@@ -22,6 +21,7 @@ export function FundCardPaymentMethodDropdown({
     setSelectedPaymentMethod,
     paymentMethods,
     fundAmountFiat,
+    paymentMethodsLoading,
```
alessey, line 24: nit: may be more readable to preface bools as is*, i.e. isPaymentMethodsLoading

src/fund/components/FundCardPresetAmountInputItem.tsx
```diff
@@ -8,8 +9,11 @@ export function FundCardPresetAmountInputItem({
   onClick,
 }: PresetAmountInputItemPropsReact) {
   const presetAmountInputText = useMemo(() => {
-    // TODO: This is going to be formatted in the next PR where we add the currency symbol so that instead of USD we have $ or other currency symbols
-    return `${presetAmountInput} ${currency}`;
+    return formatFiatAmount({
+      amount: presetAmountInput,
+      currency,
+      minimumFractionDigits: 0,
+    });
```
alessey, line 16: if you aren't passing in locale, it will always default to US?  won't that cause formatting issues showing other currencies?

src/fund/components/FundCardPaymentMethodDropdown.tsx
rustam-cb: Adding `return undefined` makes the coverage to drop and for some reason the coverage does not pick up that line as covered. (Spent too much time trying to have that one line covered and gave up on it)

src/fund/components/FundCardPaymentMethodDropdown.tsx
rustam-cb: Addressed

src/fund/components/FundCardPresetAmountInputItem.tsx
rustam-cb: You are right. It would default to `en`. Updated to remove default so that it uses browser's locale

src/fund/components/FundCardPaymentMethodDropdown.tsx
alessey: if you removed the conditional and referenced in isPaymentMethodDisabled, you should hit this code twice for any valid state

src/fund/components/FundCardPresetAmountInputItem.tsx
alessey: wouldn't that also cause problems if the user is not in the same country as the currency?

rustam-cb: It would format the number in a format that are using on their device/browser. What issues do you think there would be with this approach?

src/fund/components/FundCardPaymentMethodDropdown.tsx
rustam-cb: @alessey  I don't think I understood what you meant in the last comment. Could you please clarify?

src/fund/components/FundCardPresetAmountInputItem.tsx
alessey: maybe its not an issue, formatting 123456.789 if you were displaying USD and had an indian locale, you'd see $1,23,456.79, vs. being in the US $123,456.79, is this correct or would you want to always show dollars formatted as en-US?

src/fund/components/FundCardPaymentMethodDropdown.tsx
alessey: adding a return undefined alone wouldn't allow that to be testable as the conditional in render never calls the function if undefined.  Although if you removed the conditional in render and additionally, used this function to determine disabled state in the isPaymentMethodDisabled check, the return undefined should get hit when testing a valid value.

src/fund/components/FundCardPresetAmountInputItem.tsx
rustam-cb: @alessey  This is how it works currently.
Using currency  `USD`:
locale en-EN: ---> `$123456`
locale en-CA: ---> `US$123456`
locale en-IN: ----> '$123456'

This is correct.

src/fund/components/FundCardPaymentMethodDropdown.tsx
rustam-cb: @alessey  Updated.

src/fund/components/FundCardProvider.tsx
```diff
@@ -21,6 +21,7 @@ import { fetchOnrampQuote } from '../utils/fetchOnrampQuote';
 
 type FundCardContextType = {
   asset: string;
+  currency: string;
```
alessey, line 24: what currency strings are supported?  is there a way to validate? what happens with an invalid currency?

rustam-cb: @alessey any string is supported. If developer enters the non-existent or non-supported currency then they will get an error.

src/core/utils/formatFiatAmount.ts
```diff
@@ -0,0 +1,32 @@
...
+export const formatFiatAmount = ({
+  amount,
+  currency = 'USD',
+  locale,
+  minimumFractionDigits = 2,
+  maximumFractionDigits = 2,
+}: {
+  amount: number | string;
+  currency?: string;
+  locale?: string;
```
brendan-defi, line 17: seems like locale is required right? maybe put a default?

rustam-cb: @brendan-defi `locale` is optional here. It will use the browser default locale when nothing is provided.

src/fund/components/FundCard.tsx
```diff
@@ -18,6 +17,7 @@ export function FundCard({
   headerText,
   country = 'US',
   subdivision,
+  currency,
```
brendan-defi, line 20: i think you need a default here, since this type is optional but it's required in FundCardProvider

src/fund/components/FundCardProvider.tsx
alessey: on ramp will return an error which will be emitted?  works for me, would it be possible to link to a list of supported currencies from the docs?

src/core/utils/formatFiatAmount.ts
brendan-defi: gotcha, didn't realize that, thank you

src/core/utils/formatFiatAmount.test.ts
```diff
@@ -0,0 +1,100 @@
+import { describe, expect, it } from 'vitest';
+import { formatFiatAmount } from './formatFiatAmount';
+
+describe('formatFiatAmount', () => {
+  it('uses USD as default currency', () => {
+    expect(formatFiatAmount({ amount: 100 })).toBe('$100.00');
```
alessey, line 6: may want to pass in locale for all these tests or they'll fail for devs in different local

src/core/utils/formatFiatAmount.test.ts
```diff
@@ -0,0 +1,100 @@
...
+  it('respects different locales', () => {
+    expect(
+      formatFiatAmount({
+        amount: 1000.5,
+        currency: 'EUR',
+        locale: 'de-DE',
+      })
+        .replace(/\u00A0/g, ' ')
+        .trim(),
+    ).toBe('1.000,50 €'.replace(/\u00A0/g, ' ').trim());
```
alessey, line 40: why the replaces?

src/core/utils/formatFiatAmount.ts
alessey: could add this to the jsdoc param

src/fund/types.ts
```diff
@@ -94,7 +94,7 @@ type GetOnrampBuyUrlOptionalProps = {
   /**
    * The default payment method that will be selected for the user in the Onramp UI. Should be one of the payment methods
    */
-  defaultPaymentMethod?: PaymentAccountReact;
+  defaultPaymentMethod?: string;
```
alessey, line 97: do you know what the possibilities here are prior to calling onramp?

src/fund/utils/buildPaymentMethods.ts
```diff
@@ -0,0 +1,92 @@
...
+) => {
+  const availableMethods = [
+    // Check API-provided methods
+    paymentMethodLimits.some((limit) => limit.id === 'ACH_BANK_ACCOUNT') &&
+      'ACH',
+    paymentMethodLimits.some((limit) => limit.id === 'CARD') && 'debit',
+    // Always include these methods
+    'cash',
+    'crypto balance',
+  ].filter(Boolean); // Remove falsy values
```
alessey, line 25: nit: could use reduce and do this in 1 loop instead of 3

src/fund/components/FundCardProvider.tsx
rustam-cb: Updated. Added error emit.

src/core/utils/formatFiatAmount.ts
rustam-cb: updated

src/fund/components/FundCard.tsx
rustam-cb: Updated

src/core/utils/formatFiatAmount.test.ts
rustam-cb: Updated

src/core/utils/formatFiatAmount.test.ts
rustam-cb: Must be something with the locale (I think different ASCI for different locale)

src/fund/types.ts
rustam-cb: No, we don't know what payment methods user has in their Coinbase account at this time.

src/fund/utils/buildPaymentMethods.ts
rustam-cb: I need these to be in a certain order.

alessey (approved):

brendan-defi (approved):

PR #1822, opened 2025-01-16
feat: Add UI Primitives - FocusTrap, DismissableLayer, and Dialog. 

cpcramer:
**What changed? Why?**
Introduces new UI primitive components - `FocusTrap`, `DismissableLayer`, and `Dialog`.

`FocusTrap`:

- Ensures keyboard focus remains within a contained area for accessibility

`DismissableLayer`:

- Handles dismissal using outside clicks and escape key events

`Dialog`:
 * Portaling to document.body
 * Focus management (trapping focus within dialog)
 * Click outside and escape key dismissal
 * Proper ARIA attributes for accessibility

Design Principles: 
- Limited styling 
- ARIA Accessibility built in
- Optional props by default 
- Stand-alone primitive components

Updated `WalletModal` to use the primitives for:
- Proper modal dialog behavior with ARIA attributes
- Focus trapping for keyboard navigation
- Click outside and escape key dismissal
- Portal-based rendering

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/internal/primitives/Dialog.tsx
+++ b/src/internal/primitives/Dialog.tsx
@@ -1,0 +1,77 @@
+import type React from 'react';
+import { useRef } from 'react';
+import { createPortal } from 'react-dom';
+import { useTheme } from '../../core-react/internal/hooks/useTheme';
+import { cn } from '../../styles/theme';
+import { DismissableLayer } from './DismissableLayer';
+import { FocusTrap } from './FocusTrap';
+
+type DialogProps = {
+  children?: React.ReactNode;
+  isOpen?: boolean;
+  onClose?: () => void;
+  modal?: boolean;
+  ariaLabel?: string;
+  ariaLabelledby?: string;
+  ariaDescribedby?: string;
+};
+
+/**
+ * Dialog primitive that handles:
+ * Portaling to document.body
+ * Focus management (trapping focus within dialog)
+ * Click outside and escape key dismissal
+ * Proper ARIA attributes for accessibility
+ */
+export function Dialog({
+  children,
+  isOpen,
+  modal = true,
+  onClose,
+  ariaLabel,
+  ariaLabelledby,
+  ariaDescribedby,
+}: DialogProps) {
+  const componentTheme = useTheme();
+  const dialogRef = useRef<HTMLDivElement>(null);
+
+  if (!isOpen) {
+    return null;
+  }
+
+  const dialog = (
+    <div
+      className={cn(
+        componentTheme,
+        'fixed inset-0 z-50 flex items-center justify-center',
+        'bg-black/50 transition-opacity duration-200',
+        'fade-in animate-in duration-200',
+      )}
+    >
+      <FocusTrap active={isOpen}>
+        <DismissableLayer onDismiss={onClose}>
+          <div
+            aria-modal={modal}
+            aria-label={ariaLabel}
+            aria-labelledby={ariaLabelledby}
+            aria-describedby={ariaDescribedby}
+            data-testid="ockDialog"
+            onClick={(e) => e.stopPropagation()}
+            onKeyDown={(e: React.KeyboardEvent<HTMLDivElement>) => {
+              if (e.key === 'Enter' || e.key === ' ') {
+                e.stopPropagation();
+              }
+            }}
+            ref={dialogRef}
+            role="dialog"
+            className="zoom-in-95 animate-in duration-200"
+          >
+            {children}
+          </div>
+        </DismissableLayer>
+      </FocusTrap>
+    </div>
+  );
+
+  return createPortal(dialog, document.body);
+}
--- a/src/internal/primitives/Dialog.tsx
+++ b/src/internal/primitives/Dialog.tsx
@@ -11,9 +11,9 @@ type DialogProps = {
   isOpen?: boolean;
   onClose?: () => void;
   modal?: boolean;
-  ariaLabel?: string;
-  ariaLabelledby?: string;
-  ariaDescribedby?: string;
+  'aria-label'?: string;
+  'aria-labelledby'?: string;
+  'aria-describedby'?: string;
 };
 
 /**
@@ -28,9 +28,9 @@ export function Dialog({
   isOpen,
   modal = true,
   onClose,
-  ariaLabel,
-  ariaLabelledby,
-  ariaDescribedby,
+  'aria-label': ariaLabel,
+  'aria-labelledby': ariaLabelledby,
+  'aria-describedby': ariaDescribedby,
 }: DialogProps) {
   const componentTheme = useTheme();
   const dialogRef = useRef<HTMLDivElement>(null);
--- a/src/internal/primitives/DismissableLayer.tsx
+++ b/src/internal/primitives/DismissableLayer.tsx
@@ -1,0 +1,75 @@
+import type React from 'react';
+import { useEffect, useRef } from 'react';
+
+type DismissableLayerProps = {
+  children?: React.ReactNode;
+  disableEscapeKey?: boolean;
+  disableOutsideClick?: boolean;
+  onDismiss?: () => void;
+};
+
+// DismissableLayer handles dismissal using outside clicks and escape key events
+export function DismissableLayer({
+  children,
+  disableEscapeKey = false,
+  disableOutsideClick = false,
+  onDismiss,
+}: DismissableLayerProps) {
+  const layerRef = useRef<HTMLDivElement>(null);
+  // Tracks whether the pointer event originated inside the React component tree
+  const isPointerInsideReactTreeRef = useRef(false);
+
+  useEffect(() => {
+    if (disableOutsideClick && disableEscapeKey) {
+      return;
+    }
+
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if (!disableEscapeKey && event.key === 'Escape') {
+        onDismiss?.();
+      }
+    };
+
+    const shouldDismiss = (target: Node) => {
+      return layerRef.current && !layerRef.current.contains(target);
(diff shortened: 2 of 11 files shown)


src/wallet/components/ConnectWallet.test.tsx
```diff
@@ -341,32 +355,6 @@ describe('ConnectWallet', () => {
   });
 
   describe('modal handling', () => {
-    it('should close modal when clicking overlay', () => {
```
cpcramer, line 344: Redundant now since we have testing elsewhere.

src/primitives/Dialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { FocusTrap } from './FocusTrap';
+
+type DialogProps = {
+  children?: React.ReactNode;
+  isOpen?: boolean;
+  onClose?: () => void;
+  modal?: boolean;
+  'aria-label'?: string;
+  'aria-labelledby'?: string;
+  'aria-describedby'?: string;
```
brendan-defi, line 16: why not camel case the keys?

src/primitives/Dialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+            aria-label={ariaLabel}
+            aria-labelledby={ariaLabelledBy}
+            aria-describedby={ariaDescribedBy}
+            data-testid="ockDialog"
+            onClick={(e) => e.stopPropagation()}
+            onKeyDown={(e) => {
+              if (e.key === 'Enter' || e.key === ' ') {
+                e.stopPropagation();
+              }
+            }}
```
brendan-defi, line 64: should these events be typed?

src/primitives/Dialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+            className="zoom-in-95 animate-in duration-200"
+          >
+            {children}
+          </div>
+        </DismissableLayer>
+      </FocusTrap>
+    </div>
+  );
+
+  return createPortal(dialog, document.body);
```
brendan-defi, line 76: should we put the createPortal on `DismissableLayer`? under what circumstances would we want a `DismissableLayer` that's not portaled?

src/primitives/FocusTrap.tsx
```diff
@@ -0,0 +1,80 @@
...
+    return () => {
+      // Restore focus to previous element when trap is destroyed
+      previousFocusRef.current?.focus();
+    };
+  }, [active]);
+
+  const getFocusableElements = () =>
+    containerRef.current?.querySelectorAll<HTMLElement>(
+      'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])',
+    );
```
brendan-defi, line 38: nit: maybe store these values somewhere since you reuse them

cpcramer: Threw it in a const at the top of the page

src/primitives/Dialog.tsx
cpcramer: Some `DismissableLayer` use cases that should not be portaled are dropdowns and popovers!

cpcramer: But Dialogs / Modals should always be portaled

src/primitives/Dialog.tsx
cpcramer: I was matching the HTML syntax. But I'll go ahead and update to camel case - seems more appropriate

brendan-defi:
you've created a new directory `src/primitives`. Seems like this should either be exported or put into `src/internal`, no?

src/primitives/Dialog.tsx
cpcramer: yeah, I wonder why linter doesn't catch this hm

src/primitives/Dialog.tsx
brendan-defi: makes sense, thanks!

cpcramer:
> you've created a new directory `src/primitives`. Seems like this should either be exported or put into `src/internal`, no?

I was also considering `src/ui/react`

brendan-defi:
> > you've created a new directory `src/primitives`. Seems like this should either be exported or put into `src/internal`, no?
> 
> I was also considering `src/ui/react`

could also work, and that would make more sense if we are including them in the package.

alessey:
> > > you've created a new directory `src/primitives`. Seems like this should either be exported or put into `src/internal`, no?
> > 
> > 
> > I was also considering `src/ui/react`
> 
> could also work, and that would make more sense if we are including them in the package.

since we are reverting the codebase changes, i think these make sense in src/internal

dschlabach:
+1 @alessey, should be in `src/internal`

src/internal/primitives/FocusTrap.tsx
```diff
@@ -0,0 +1,83 @@
...
+    }
+    // Store currently focused element to restore later
+    previousFocusRef.current = document.activeElement as HTMLElement;
+
+    if (containerRef.current) {
+      // Query all interactive elements that can receive focus
+      const firstFocusable = containerRef.current.querySelector<HTMLElement>(
+        FOCUSABLE_ELEMENTS_SELECTOR,
+      );
+      firstFocusable?.focus();
```
dschlabach, line 29: I'm not sure if this is exactly where it's happening, but there's a slight bug when you _click_ on the trigger and the first element will be focused briefly, then unfocused (see the gray flash?)

https://github.com/user-attachments/assets/0803f1bf-1138-4a43-8213-c8cdf54ad94e

src/internal/primitives/Dialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+type DialogProps = {
+  children?: React.ReactNode;
+  isOpen?: boolean;
+  onClose?: () => void;
+  modal?: boolean;
+  ariaLabel?: string;
+  ariaLabelledby?: string;
+  ariaDescribedby?: string;
+};
```
alessey, line 17: I personally prefer matching the actual dom attributes when possible

src/internal/primitives/DismissableLayer.tsx
```diff
@@ -0,0 +1,75 @@
...
+      // Set flag when pointer event starts inside the component
+      // This prevents dismissal when dragging from inside to outside
+      onPointerDownCapture={() => {
+        isPointerInsideReactTreeRef.current = true;
+      }}
+      ref={layerRef}
+    >
+      {children}
+    </div>
+  );
```
alessey, line 74: is this basically useOutsideClick + escape key handling?

src/internal/primitives/FocusTrap.tsx
```diff
@@ -0,0 +1,83 @@
+import type React from 'react';
+import { useEffect, useRef } from 'react';
+
+const FOCUSABLE_ELEMENTS_SELECTOR =
+  'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])';
```
dschlabach, line 5: You probably want to add `:not([disabled])` pseudo classes to the button, input, select, and textarea elements. Like:

`button:not([disabled])`
```suggestion
  'button:not([disabled]), [href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), [tabindex]:not([tabindex="-1"])';
```

src/internal/primitives/Dialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+            ref={dialogRef}
+            role="dialog"
+            className="zoom-in-95 animate-in duration-200"
+          >
+            {children}
+          </div>
+        </DismissableLayer>
+      </FocusTrap>
+    </div>
+  );
```
alessey, line 74: should we implement a standard (overridable) mobile style?  i.e. modals are full screen by default on mobile?

src/internal/primitives/DismissableLayer.tsx
cpcramer: Yep!

src/internal/primitives/Dialog.tsx
cpcramer: Sounds good, will revert

src/internal/primitives/FocusTrap.tsx
cpcramer: Good call

src/internal/primitives/FocusTrap.tsx
cpcramer: fixed

src/internal/primitives/Dialog.tsx
cpcramer: Good idea, that would be similar to how Radix functions. I'll implement as a follow up task

cpcramer: tracking https://linear.app/coinbase/issue/TOOLS-1037/update-all-modals-to-be-full-screen-on-mobile

brendan-defi: should modals be trays by default on mobile?

src/internal/primitives/DismissableLayer.tsx
alessey: any reason to not just use the useOutsideClick hook and possibly create a separate useEscapeKey hook (or useKeyPress({key[], callback})?

src/internal/primitives/Dialog.tsx
brendan-defi: sorry for my bad suggestion

src/internal/primitives/DismissableLayer.tsx
alessey: or alternatively if its preferable to keep this ux, have it implement those hooks?

dschlabach (approved):

brendan-defi (approved):

PR #1842, opened 2025-01-19
Chore: WalletIsland Code Quality Fast Follows

brendan-defi:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/internal/components/Draggable/getDefaultStartingPosition.ts
```diff
@@ -0,0 +1,10 @@
+export function getDefaultStartingPosition() {
+  if (typeof window === 'undefined') {
+    return { x: 100, y: 100 };
+  }
+
+  return {
+    x: window.innerWidth * 0.95,
+    y: window.innerHeight * 0.95,
+  };
+}
```
alessey, line 10: this seems like a wallet util

src/internal/components/Draggable/useBoundedPosition.ts
```diff
@@ -0,0 +1,77 @@
...
+        Math.max(10, currentPosition.y),
+        viewportHeight - rect.height - 10,
+      );
+
+      return { x: newX, y: newY };
+    },
+    [],
+  );
+
+  const timeoutRef = useRef<NodeJS.Timeout>(undefined);
```
alessey, line 31: isn't this just type number?

src/internal/components/Draggable/useBoundedPosition.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+    const newPosition = repositionDraggable(rect, position);
+    resetPosition(newPosition);
+  }, [draggableRef, position, repositionDraggable, resetPosition]);
+
+  useEffect(() => {
+    window.addEventListener('resize', handleWindowResize);
+    return () => window.removeEventListener('resize', handleWindowResize);
+  }, [handleWindowResize]);
+}
```
alessey, line 77: can this hook just accept a position input and return a bounded position?

src/core/utils/copyToClipboard.ts
```diff
@@ -0,0 +1,20 @@
...
+  onSuccess,
+  onError,
+}: CopyToClipboardParams) {
+  try {
+    await navigator.clipboard.writeText(text);
+    onSuccess?.();
+  } catch (err) {
+    onError?.(err);
+  }
+}
```
alessey, line 20: a more useful component may be a button that can wrap elements adding this copy functionality

dschlabach: +1, Yeah I've used something like that in the past and it ends up being a really nice pattern

src/internal/components/Draggable/useBoundedPosition.ts
```diff
@@ -0,0 +1,77 @@
...
+    return;
+  }
+
+  const repositionDraggable = useCallback(
+    (rect: DOMRect, currentPosition: { x: number; y: number }) => {
+      const viewportWidth = window.innerWidth;
+      const viewportHeight = window.innerHeight;
+
+      const newX = Math.min(
+        Math.max(10, currentPosition.x),
```
dschlabach, line 18: Is `10` the minimum offset from the edge of the screen? It would be helpful for readability to extract that into a helper const

src/internal/components/Draggable/Draggable.tsx
```diff
@@ -110,16 +100,16 @@ export function Draggable({
     dragStartPosition,
   ]);
 
-  useRespositionOnWindowResize(draggableRef, position, setPosition);
+  useBoundedPosition(draggableRef, position, setPosition);
```
dschlabach, line 103: I think it'd be better to prevent the user from dragging the `Draggable` off the screen altogether rather than waiting to see if they move it off and repositioning it. It's a little odd to allow it and then reset it vs just prevent altogether

dschlabach: Just saw @alessey comment below, that'd be a nice way to do it as well

src/core/utils/copyToClipboard.ts
brendan-defi: i can make a button. would that be instead of, or in addition to this util?

src/internal/components/Draggable/Draggable.tsx
brendan-defi: interesting. i feel like all the islands i've played with allow off-screen dragging. will investigate @alessey's proposal

src/internal/components/Draggable/getDefaultStartingPosition.ts
brendan-defi: okay, i'll move to wallet. in that case will rename to `getDefaultDraggableStartingPosition`

src/internal/components/Draggable/useBoundedPosition.ts
brendan-defi: i'll parameterize with a default

src/internal/components/Draggable/useBoundedPosition.ts
brendan-defi: hmm, no. 
image

this also works fine, and is perhaps more readable?
```
const timeoutRef = useRef<ReturnType<typeof setTimeout>>(undefined);
```

brendan-defi: not relevant anymore since we're moving away from this structure

src/core/utils/copyToClipboard.ts
brendan-defi: see `CopyButton`

src/internal/components/Draggable/Draggable.tsx
alessey: if you drag offscreen how would you get it back?

src/internal/components/CopyButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+  className,
+  'aria-label': ariaLabel,
+}: CopyButtonProps) {
+  return (
+    <button
+      type="button"
+      data-testid="ockCopyButton"
+      className={className}
+      onClick={() => {
+        console.log('clicked copyButton');
```
alessey, line 27: intentional?

src/core/utils/copyToClipboard.ts
alessey: I'd prefer consumers use the button vs. the util

src/internal/components/CopyButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+        copyToClipboard({ copyValue, onSuccess, onError });
+      }}
+      aria-label={ariaLabel}
+    >
+      {label}
+    </button>
```
alessey, line 33: can this also handle the tooltip?

brendan-defi: yes, the tooltip is implemented with CopyButton
image

src/internal/components/CopyButton.tsx
brendan-defi: removing

dschlabach (approved):

cpcramer (approved):

PR #1852, opened 2025-01-22
feat: add transaction builder functions for interacting with Morpho vaults

dschlabach:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/earn/components/EarnProvider.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return <EarnContext.Provider value={value}>{children}</EarnContext.Provider>;
+}
+
+export function useEarnContext() {
+  const context = useContext(EarnContext);
+  if (!context) {
+    throw new Error('useEarnContext must be used within an EarnProvider');
+  }
+  return context;
+}
```
brendan-defi, line 30: our pattern has not been to throw if `useComponentContext`s have been used outside of their providers. Perhaps that's bad practice, but it's useful in certain cases (eg. if you want to attempt to use a context, and if you get back values use them, and if not source them from elsewhere).

src/earn/constants.ts
```diff
@@ -0,0 +1,26 @@
+export const METAMORPHO_ABI = [
```
brendan-defi, line 1: "_meta_morpho"?

src/earn/utils/deposit.ts
```diff
@@ -0,0 +1,42 @@
...
+import { type Address, encodeFunctionData, erc20Abi } from 'viem';
+
+type DepositToMorphoArgs = {
+  vaultAddress: Address;
+  tokenAddress: Address;
+  amount: bigint;
+  receiverAddress: Address;
+};
+
+export function buildDepositToMorphoTx({
```
brendan-defi, line 12: why not align the names between the funcs and the file? shouldn't the file be called depositToMorpho?

src/earn/utils/withdraw.ts
```diff
@@ -0,0 +1,28 @@
...
+    args: [amount, receiverAddress, receiverAddress],
+  });
+
+  return [
+    {
+      to: vaultAddress,
+      data: withdrawTxData,
+    },
+  ];
+}
```
brendan-defi, line 28: approvals will always have happened from the deposit tx?

src/earn/components/EarnProvider.tsx
dschlabach: If we're not throwing these errors in our context hooks, that seems like a code smell. It's the standard way of doing it for pretty much all major libraries.

We also do this in:
- `useNFTLifecycleContext`
- `useTheme`
- `useBuyContext`
- `useCheckoutContext`
- `useTransactionContext`

src/earn/utils/withdraw.ts
dschlabach: You don't need an approval to withdraw from a vault as the ERC20 is under the control of the vault

src/earn/components/EarnProvider.tsx
abcrane123: i think we should include this in all providers

src/earn/utils/deposit.ts
abcrane123: agreed

src/earn/constants.ts
dschlabach: I renamed it to `MORPHO_VAULT_ABI`, perhaps that's clearer?

src/earn/components/EarnProvider.tsx
alessey: as far as I know, wallet is the exception so it will silently fail on an unconnected wallet button

src/earn/utils/deposit.ts
```diff
@@ -0,0 +1,42 @@
+import { METAMORPHO_ABI } from '@/earn/constants';
+import type { Call } from '@/transaction/types';
+import { type Address, encodeFunctionData, erc20Abi } from 'viem';
+
+type DepositToMorphoArgs = {
+  vaultAddress: Address;
+  tokenAddress: Address;
+  amount: bigint;
+  receiverAddress: Address;
+};
```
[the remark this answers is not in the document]
dschlabach: Done

src/earn/utils/deposit.ts
dschlabach: Thanks, done. Was initially going to put all the deposit functions in this `deposit.ts` module, but I'll break them up as we have in the rest of the codebase.

src/earn/components/EarnProvider.tsx
brendan-defi: the providers that do not currently do this are:
`OnchainKitProvider`
image

`IdentityProvider`
image

`WalletProvider`
image

can y'all think of a reason we wouldn't want to throw in these cases? if not I can add

src/earn/utils/withdraw.ts
brendan-defi: ah, right, thanks

alessey (approved):

brendan-defi (approved):

alessey:
looks like the identity components are set up so that the provider is optional, same as wallet

abcrane123 (approved):

PR #1864, opened 2025-01-23
feat: add Tabs component

abcrane123:
**What changed? Why?**
- add simple tabs component to be used in Earn
- putting in internal in case we have another use for tabs in the future

```
          <Tabs defaultValue="deposit">
            <TabsList>
              <Tab value="deposit">Deposit</Tab>
              <Tab value="withdraw">Withdraw</Tab>
            </TabsList>
            <TabContent value="deposit">Deposit</TabContent>
            <TabContent value="withdraw">Withdraw</TabContent>
          </Tabs>

```

**Notes to reviewers**

**How has it been tested?**


src/internal/components/Tabs.tsx
```diff
@@ -0,0 +1,94 @@
...
+  className?: string;
+  children: React.ReactNode;
+};
+
+export function TabsList({ className, children }: TabsListReact) {
+  return (
+    <div
+      className={cn(
+        border.lineDefault,
+        'flex overflow-hidden !border-r-0 !border-l-0 !border-t-0',
```
alessey, line 73: why do these need to be important?

src/internal/components/Tabs.tsx
```diff
@@ -0,0 +1,94 @@
...
+      className={cn(
+        text.label1,
+        isSelected ? color.primary : color.foreground,
+        isSelected ? background.washed : background.default,
+        'w-[50%] text-center border-b-none border-r',
+        'cursor-pointer px-3 py-4',
+        className,
+      )}
+      onClick={handleClick}
+    >
```
alessey, line 57: better for accessibility if this was actually a button, divs aren't focusable elements

src/internal/components/Tabs.tsx
abcrane123: when you add the `border.lineDefault` class it seems to take priority even if you pass `border-r-0` after and we need to use `border.lineDefault` to ensure the border that does appear is consistent with rest of theme

abcrane123: but now that i've refactored i think i can remove and rely on card to handle this

brendan-defi (dismissed):

src/internal/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,39 @@
+import { background, cn, color, text } from '@/styles/theme';
+import { useCallback } from 'react';
+import { useTabsContext } from './Tabs';
+
+type TabReact = {
+  value: string;
+  children: React.ReactNode;
+  className?: string;
+  ariaLabel?: string;
```
dschlabach, line 9: Can we follow standard html attributes? That's how Paul's done them in the internal primitives / is standard in React (e.g. `HTMLAttributes<HTMLButtonElement>>`)

```suggestion
  aria-label?: string;
```

src/internal/components/Tabs/TabContent.tsx
```diff
@@ -0,0 +1,17 @@
...
+};
+
+export function TabContent({ children, value, className }: TabContentReact) {
+  const { selectedTab } = useTabsContext();
+
+  if (selectedTab !== value) {
+    return null;
+  }
+
+  return <div className={className}>{children}</div>;
```
dschlabach, line 16:
```suggestion
  return <div className={className} role="tabpanel">{children}</div>;
```

src/internal/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,39 @@
+import { background, cn, color, text } from '@/styles/theme';
+import { useCallback } from 'react';
+import { useTabsContext } from './Tabs';
+
+type TabReact = {
+  value: string;
```
cpcramer, line 6: `value` seems vague here - is this just used to identify the currently selected tab?

cpcramer: I feel like there's a better name

src/internal/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,39 @@
...
+      className={cn(
+        text.label1,
+        isSelected ? color.primary : color.foreground,
+        isSelected ? background.washed : background.default,
+        'w-1/2 text-center',
+        'cursor-pointer px-3 py-4',
+        className,
+      )}
+      onClick={handleClick}
+      aria-label={ariaLabel}
```
dschlabach, line 32:
```suggestion
      aria-label={ariaLabel}
      aria-controls={`${value}-panel`}
```

src/internal/components/Tabs/Tab.tsx
cpcramer: Ah never mind, looks like this is what Radix-UI does

src/internal/components/Tabs/Tab.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { useTabsContext } from './Tabs';
+
+type TabReact = {
+  value: string;
+  children: React.ReactNode;
+  className?: string;
+  ariaLabel?: string;
+};
+
+export function Tab({ value, children, className, ariaLabel }: TabReact) {
```
cpcramer, line 12: I think we should implement / apply theming here

dschlabach: Can we follow that in a separate PR? Let's try and keep this one super focused

src/internal/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { createContext, useContext, useState } from 'react';
+
+type TabsReact = {
+  children: React.ReactNode;
+  defaultValue: string;
+};
+
+export function Tabs({ children, defaultValue }: TabsReact) {
+  return <TabsProvider defaultValue={defaultValue}>{children}</TabsProvider>;
+}
```
dschlabach, line 11: Can you move this to the bottom so that it's easier to find

src/internal/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,38 @@
...
+const TabsContext = createContext<TabsContextType | undefined>(undefined);
+
+type TabsProviderReact = {
+  children: React.ReactNode;
+  defaultValue: string;
+};
+
+function TabsProvider({ children, defaultValue }: TabsProviderReact) {
+  const [selectedTab, setSelectedTab] = useState<string>(defaultValue);
+  const value = useValue({ selectedTab, setSelectedTab });
```
dschlabach, line 27: The only other accessibility / quality thing we need to do is have Left and Right Arrow focus the next tab. Some libraries use a `RovingFocus` or component wrapper for this, but could just implement it as a one off in this component

Here's the W3 accessibility spec if you're curious: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/#wai-ariaroles,states,andproperties

src/internal/components/Tabs/Tab.tsx
abcrane123: i think it makes sense to add here since we'll probably want theming to always apply

abcrane123: but i added at provider level, let me know if you strongly disagree

src/internal/components/Tabs/Tab.test.tsx
```diff
@@ -0,0 +1,70 @@
+import { fireEvent, render, screen } from '@testing-library/react';
```
dschlabach, line 1: This is a nit but could you also change the directory name from `components/Tabs` to `components/tabs`? I know we have a couple here that don't follow that pattern, but it's bad practice that we should try and avoid

e.g.:
https://x.com/kentcdodds/status/1874830406140448943

cpcramer (approved):

dschlabach (approved):

PR #1868, opened 2025-01-23
chore: Standardize comment format to JSDoc-style syntax

cpcramer:
**What changed? Why?**

## chore: Standardize comment format to JSDoc-style syntax

We're implementing a consistent approach to documentation comments across our codebase. This will improve readability and make our documentation more maintainable.

### New Comment Standards

1. **Multi-line comments** (`/** */`) for:
   - Function definitions
   - Class definitions
   - Type/Interface definitions
   - Complex exports

2. **Single-line comments** (`/** */`) for:
   - Properties within types/interfaces
   - Simple constants
   - Basic exports

3. **Regular comments** (`//`) for:
   - Inside function implementations
   - Implementation details
   - Temporary notes
   - Line-by-line explanations
   
### Example:

```
/**
 */
export type isEthereumOptions = {
  /** Chain ID for the network */
  chainId: number;
  /** If the chainId check is only allowed on mainnet */
  isMainnetOnly?: boolean;
};

/**
 * DismissableLayer handles dismissal using outside clicks and escape key events
 */
export function DismissableLayer({}) {
      // This prevents the popover from closing when clicking inside it
      if (layerRef.current?.contains(event.target as Node)) {
        return;
      }

      // Handling for the trigger button (e.g., settings toggle)
      // Without this, clicking the trigger would cause both:
      // 1. The button's onClick to fire (toggling isOpen)
      // 2. This dismissal logic to fire (forcing close)
      // This would create a race condition where the popover rapidly closes and reopens
      const isTriggerClick = (event.target as HTMLElement).closest(
        '[aria-label="Toggle swap settings"]',
      );
}
```
Result: 
Screenshot 2025-01-23 at 12 54 20 PM

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/buy/types.ts
+++ b/src/buy/types.ts
@@ -10,22 +10,32 @@ import type { Token } from '@/token';
 import type { Address, TransactionReceipt } from 'viem';
 
 export type BuyReact = {
-  className?: string; // Optional className override for top div element.
+  /** Optional className override for top div element. */
+  className?: string;
   config?: SwapConfig;
-  disabled?: boolean; // Disables Buy button
+  /** Disables Buy button */
+  disabled?: boolean;
   experimental?: {
-    useAggregator: boolean; // Whether to use a DEX aggregator. (default: true)
+    /** Whether to use a DEX aggregator. (default: false) */
+    useAggregator: boolean;
   };
-  isSponsored?: boolean; // An optional setting to sponsor swaps with a Paymaster. (default: false)
-  onError?: (error: SwapError) => void; // An optional callback function that handles errors within the provider.
-  onStatus?: (lifecycleStatus: LifecycleStatus) => void; // An optional callback function that exposes the component lifecycle state
-  onSuccess?: (transactionReceipt?: TransactionReceipt) => void; // An optional callback function that exposes the transaction receipt
-  fromToken?: Token; // An optional token to swap from
-  toToken: Token; // The token to swap to
+  /** An optional setting to sponsor swaps with a Paymaster. (default: false) */
+  isSponsored?: boolean;
+  /** An optional callback function that handles errors within the provider. */
+  onError?: (error: SwapError) => void;
+  /** An optional callback function that exposes the component lifecycle state */
+  onStatus?: (lifecycleStatus: LifecycleStatus) => void;
+  /** An optional callback function that exposes the transaction receipt */
+  onSuccess?: (transactionReceipt?: TransactionReceipt) => void;
+  /** An optional token to swap from */
+  fromToken?: Token;
+  /** The token to swap to */
+  toToken: Token;
 };
 
 export type BuyContextType = {
-  address?: Address; // Used to check if user is connected in SwapButton
+  /** Used to check if user is connected in SwapButton */
+  address?: Address;
   config: SwapConfig;
   disabled?: boolean;
   fromETH: SwapUnit;
@@ -33,9 +43,10 @@ export type BuyContextType = {
   lifecycleStatus: LifecycleStatus;
   handleAmountChange: (amount: string) => void;
   handleSubmit: (fromToken: SwapUnit) => void;
+  /** A function to set the lifecycle status of the component */
   updateLifecycleStatus: (
     state: LifecycleStatusUpdate<LifecycleStatus>,
-  ) => void; // A function to set the lifecycle status of the component
+  ) => void;
   setTransactionHash: (hash: string) => void;
   fromToken?: Token;
   to?: SwapUnit;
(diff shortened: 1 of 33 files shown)


src/buy/types.ts
```diff
@@ -10,32 +10,43 @@ import type { Token } from '@/token';
...
-  className?: string; // Optional className override for top div element.
+  /** Optional className override for top div element. */
+  className?: string;
   config?: SwapConfig;
-  disabled?: boolean; // Disables Buy button
+  /** Disables Buy button */
+  disabled?: boolean;
   experimental?: {
-    useAggregator: boolean; // Whether to use a DEX aggregator. (default: true)
+    /** Whether to use a DEX aggregator. (default: true) */
```
alessey, line 19: nit: this is actually wrong and it defaults false, same for swap type

alessey (dismissed):

src/buy/types.ts
cpcramer: Updated all references including docs

dschlabach (dismissed):

src/buy/types.ts
```diff
@@ -50,16 +61,22 @@ export type BuyContextType = {
...
   children: React.ReactNode;
   config?: {
-    maxSlippage: number; // Maximum acceptable slippage for a swap. (default: 10) This is as a percent, not basis points
+    /** Maximum acceptable slippage for a swap. (default: 10) This is as a percent, not basis points */
+    maxSlippage: number;
   };
   disabled?: boolean;
   experimental: {
-    useAggregator: boolean; // Whether to use a DEX aggregator. (default: true)
+    /** Whether to use a DEX aggregator. (default: true) */
```
brendan-defi, line 69: per Adam above, this defaults to false, right?

src/fund/constants.ts
```diff
@@ -1,11 +1,21 @@
+/** The base URL for the Coinbase Onramp widget */
 export const DEFAULT_ONRAMP_URL = 'https://pay.coinbase.com';
 // The base URL for the Coinbase Onramp widget.
+
+/** The base URL for the Coinbase Onramp widget */
 export const ONRAMP_BUY_URL = `${DEFAULT_ONRAMP_URL}/buy`;
```
brendan-defi, line 6: identical comment for what appear to be different values

src/identity/types.ts
```diff
@@ -5,10 +5,14 @@ import type { Address, Chain } from 'viem';
...
-  address?: Address | null; // The Ethereum address to render.
-  className?: string; // Optional className override for top span element.
-  isSliced?: boolean; // Determines if the displayed address should be sliced.
-  hasCopyAddressOnClick?: boolean; // Defaults to true. Optional boolean to disable copy address on click functionality.
+  /** The Ethereum address to render. */
+  address?: Address | null;
+  /** Optional className override for top span element. */
+  className?: string;
+  /** Determines if the displayed address should be sliced. */
+  isSliced?: boolean;
```
brendan-defi, line 13: we should say that we default to true, esp because as a dev i wouldn't assume this would default to true

src/buy/types.ts
cpcramer: i'd have to check, but i'd like to keep this PR mostly tied to reformatting comments as is

src/wallet/constants.ts
```diff
@@ -1,18 +1,23 @@
...
+
+/** The Coinbase Smart Wallet factory address. */
 export const CB_SW_FACTORY_ADDRESS =
-  '0x0BA5ED0c6AA8c49038F819E587E2633c4A9F428a';
+  '0x0BA5Ed0c6AA8c49038F819E587E2633c4A9F428a';
+
+/** Maximum height for wallet advanced view */
 export const WALLET_ADVANCED_MAX_HEIGHT = 400;
 export const WALLET_ADVANCED_MAX_WIDTH = 352;
 export const WALLET_ADVANCED_DEFAULT_SWAPPABLE_TOKENS: Token[] = [
```
brendan-defi, line 23: if we're adding comments, could you add for max width and default swappable tokens as well?

cpcramer: Ah that's a mistake, trying to avoid adding comments in this PR

src/swap/types.ts
```diff
@@ -217,24 +234,36 @@ export type SwapLoadingState = {
...
-  fromAmount: string; // The amount of the source token
-  fromAmountUSD: string; // The USD value of the source token
-  hasHighPriceImpact: boolean; // Whether the price impact is high
-  priceImpact: string; // The price impact of the swap
-  slippage: string; // The slippage of the swap
-  to: Token; // The destination token for the swap
-  toAmount: string; // The amount of the destination token
-  toAmountUSD: string; // The USD value of the destination token
-  warning?: QuoteWarning; // The warning associated with the quote
+  /** The reference amount for the quote */
```
brendan-defi, line 245: non-blocking: i don't know what a reference amount is, and this comment doesn't help me understand

brendan-defi (approved):

brendan-defi (dismissed):

dschlabach (approved):

brendan-defi (approved):

PR #1869, opened 2025-01-23
feat: Client Analytics

0xAlec:
**What changed? Why?**
- add `interactionId` on every `OnchainKitProvider` render (user session)
- add `analyticsURL` as a `config` parameter on `OnchainKitProvider` (users can override where analytics are sent)
- add `useAnalytics` hook
- add `sendAnalytics` API

**Notes to reviewers**

**How has it been tested?**
locally

### DIFF
--- a/src/core-react/internal/hooks/useAnalytics.ts
+++ b/src/core-react/internal/hooks/useAnalytics.ts
@@ -1,6 +1,6 @@
+import { sendAnalytics } from '@/core/network/sendAnalytics';
 import type { AnalyticsEvent, AnalyticsEventData } from '@/internal/types';
 import { useOnchainKit } from '@/useOnchainKit';
-import { sendAnalytics } from '@/core/network/sendAnalytics';
 import { useEffect, useState } from 'react';
 
 export const useAnalytics = () => {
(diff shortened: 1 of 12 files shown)


src/core-react/internal/hooks/useAnalytics.ts
```diff
@@ -0,0 +1,35 @@
...
+import { useOnchainKit } from '@/core-react/useOnchainKit';
+import { sendAnalytics } from '@/core/network/sendAnalytics';
+import { useEffect, useState } from 'react';
+
+export const useAnalytics = () => {
+  const { apiKey, interactionId, config } = useOnchainKit();
+  const [appName, setAppName] = useState('');
+
+  useEffect(() => {
+    setAppName(document.title);
```
dschlabach, line 14: Curious why you decided to go with this instead of something like the origin? Given that a website could have different titles for different pages

dschlabach (dismissed):

src/core-react/internal/hooks/useAnalytics.ts
0xAlec: we're already tracking the origin server-side! 

adding `title` as another data point although we could potentially use `config.appearance.appName` as a first preference instead

dschlabach (approved):

abcrane123 (approved):

PR #1871, opened 2025-01-23
feat: add hooks for preparing Morpho transactions

dschlabach:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/earn/hooks/useMorphoVault.ts
```diff
@@ -0,0 +1,68 @@
+import { MORPHO_VAULT_ABI } from '@/earn/abis/morpho';
+import { type Address, erc20Abi, formatUnits } from 'viem';
+import { useReadContract, useReadContracts } from 'wagmi';
+
+type UseMorphoVaultParams = {
+  vaultAddress: Address;
+  address: Address;
```
cpcramer, line 7: This is the Wallet Address?

cpcramer: To help clarify differences between `vaultAddress`, `receiverAddress`, and `address` - Maybe we can update naming for add a comment?

alessey (approved):

src/earn/hooks/useMorphoVault.ts
dschlabach: yes, that's a good point. I'll update the that to `receiverAddress`

PR #1874, opened 2025-01-24
feat: Native Send Component

brendan-defi:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/internal/components/TokenBalance.tsx
```diff
@@ -0,0 +1,129 @@
...
+  onClick,
+  showAction = false,
+  actionText = 'Use max',
+  onActionPress,
+  className,
+}: TokenBalanceProps) {
+  const formattedValueInFiat = new Intl.NumberFormat('en-US', {
+    style: 'currency',
+    currency: 'USD',
+  }).format(token.fiatBalance);
```
dschlabach, line 31: Use `formatFiatAmount`?

src/internal/components/TokenBalance.tsx
```diff
@@ -0,0 +1,129 @@
...
+              color.foreground,
+              'overflow-hidden text-ellipsis whitespace-nowrap',
+            )}
+          >
+            {token.name?.trim()}
+          </span>
+          <span className={cn(text.label2, color.foregroundMuted)}>
+            {`${truncateDecimalPlaces(
+              token.cryptoBalance / 10 ** token.decimals,
+              2,
```
dschlabach, line 52: You probably want to use `formatUnits` from viem here:

https://viem.sh/docs/utilities/formatUnits.html

src/internal/components/TokenBalance.tsx
```diff
@@ -0,0 +1,129 @@
...
+            )} ${token.symbol} ${subtitle}`}
+          </span>
+        </div>
+        <div className="text-right">
+          {showAction ? (
+            <span
```
dschlabach, line 58: Why `<span >` here? Probably should be a button if it's clickable

src/token/components/TokenRow.tsx
```diff
@@ -32,7 +32,14 @@ export const TokenRow = memo(function TokenRow({
...
         {!hideImage && <TokenImage token={token} size={28} />}
         <span className="flex flex-col items-start">
-          <span className={cn(text.headline)}>{token.name}</span>
+          <span
+            className={cn(
+              text.headline,
+              'overflow-hidden text-ellipsis whitespace-nowrap text-left',
+            )}
+          >
+            {token.name.trim()}
```
dschlabach, line 41: Curious what the need for `trim()` is here?

src/wallet/components/WalletAdvancedSend/components/Send.tsx
```diff
@@ -0,0 +1,158 @@
...
+}
+
+function SendDefaultChildren() {
+  const context = useSendContext();
+
+  console.log({
+    context,
+  });
+
+  const activeStep = useMemo(() => {
```
dschlabach, line 52: I see what's going on here but it's quite hard to reason about having not built it myself. 

I think the easiest thing would probably be breaking this into internal subcomponents for each step (we'd have to think about the right names, not sure about that). What do you think about that?

src/wallet/components/WalletAdvancedSend/components/SendAddressSelector.tsx
```diff
@@ -0,0 +1,43 @@
+'use client';
+
+import { Address, Avatar, Identity, Name } from '@/identity';
+import { background, border, cn, pressable } from '@/styles/theme';
+import { useCallback } from 'react';
+import type { Address as AddressType, Chain } from 'viem';
```
dschlabach, line 6: Would avoid the rename if possible

src/wallet/components/WalletAdvancedSend/components/Send.tsx
dschlabach: e.g. it'd be easier to reason about that you're rendering the recipient search/options if this was in a component:

```
        <>
          <SendAddressInput
            selectedRecipientAddress={context.selectedRecipientAddress}
            recipientInput={context.recipientInput}
            setRecipientInput={context.setRecipientInput}
            handleRecipientInputChange={context.handleRecipientInputChange}
          />
          {context.validatedRecipientAddress && (
            <SendAddressSelector
              address={context.validatedRecipientAddress}
              senderChain={context.senderChain}
              handleAddressSelection={context.handleAddressSelection}
            />
          )}
        </>
```

src/wallet/components/WalletAdvancedSend/components/SendAddressSelector.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+type SendAddressSelectorProps = {
+  address: AddressType;
+  senderChain: Chain | null | undefined;
+  handleAddressSelection: (address: AddressType) => void;
+};
+
+export function SendAddressSelector({
```
dschlabach, line 14: How do you feel about `SendAddressOption`?

src/wallet/components/WalletAdvancedSend/components/SendAddressSelector.tsx
```diff
@@ -0,0 +1,43 @@
...
+  address,
+  senderChain,
```
dschlabach, line 16: I'd probably drop the `senderChain` here and get it from context (or global context) since it doesn't seem like something you're going to want to pass to each option?

src/wallet/components/WalletAdvancedSend/components/SendAmountInput.tsx
```diff
@@ -0,0 +1,100 @@
+'use client';
+
+import type { PortfolioTokenWithFiatValue } from '@/api/types';
+import { AmountInput } from '@/internal/components/amount-input/AmountInput';
+import { AmountInputTypeSwitch } from '@/internal/components/amount-input/AmountInputTypeSwitch';
+import { Skeleton } from '@/internal/components/Skeleton';
+import { cn, color, text } from '@/styles/theme';
+import { useSendContext } from '@/wallet/components/WalletAdvancedSend/components/SendProvider';
+
+export function SendAmountInput({
```
dschlabach, line 10: Can you move this to the bottom so it's easier to read when scanning? (e.g. exports at bottom)

src/wallet/components/WalletAdvancedSend/components/SendButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+  label?: string;
+  className?: string;
+} & Pick<
+  TransactionButtonReact,
+  'disabled' | 'pendingOverride' | 'successOverride' | 'errorOverride'
+>;
+
+export function SendButton({
+  label = 'Continue',
+  senderChain,
```
dschlabach, line 38: I'd drop chain here as well

src/wallet/components/WalletAdvancedSend/components/SendButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+  className,
+  disabled,
+  successOverride,
+  pendingOverride,
+  errorOverride,
+  cryptoAmount,
+  selectedToken,
+  callData,
```
dschlabach, line 46: I'd drop and get from context - otherwise people should just use `<Transaction />`

src/wallet/components/WalletAdvancedSend/components/SendButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+  sendTransactionError,
+}: SendButtonProps) {
+  const isSponsored = false;
+
+  const sendButtonLabel = useMemo(() => {
+    if (
+      Number(cryptoAmount) >
+      Number(selectedToken?.cryptoBalance) /
+        10 ** Number(selectedToken?.decimals)
```
dschlabach, line 55: Be careful doing this kind of math + comparison in JS - you should probably use something like `Number(formatUnits(...))` instead of rolling your own converter

src/wallet/components/WalletAdvancedSend/components/SendButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+    if (disabled) {
+      return true;
+    }
+    if (Number(cryptoAmount) <= 0) {
+      return true;
+    }
+    if (
+      Number(cryptoAmount) >
+      Number(selectedToken?.cryptoBalance) /
+        10 ** Number(selectedToken?.decimals)
```
dschlabach, line 72: Would probably extract to helper function and variable

src/wallet/components/WalletAdvancedSend/components/SendButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+      <SendTransactionButton
+        className={className}
+        senderChain={senderChain}
+        pendingOverride={pendingOverride}
+        errorOverride={errorOverride}
+        successOverride={successOverride}
+        disabled={isDisabled}
+        label={sendButtonLabel}
+      />
+      {!sendTransactionError && <TransactionSponsor />}
```
dschlabach, line 94: Is this right?

src/wallet/components/WalletAdvancedSend/components/SendButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+  callData: Call | null;
+  sendTransactionError: string | null;
+  label?: string;
+  className?: string;
+} & Pick<
+  TransactionButtonReact,
+  'disabled' | 'pendingOverride' | 'successOverride' | 'errorOverride'
+>;
+
+export function SendButton({
```
dschlabach, line 36: Could you move to bottom so it's easier to find the export?

src/wallet/components/WalletAdvancedSend/components/SendButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+  const { address } = useWalletContext();
+  const { setShowSend } = useWalletAdvancedContext();
+
+  const { transactionHash, transactionId } = useTransactionContext();
+
+  const accountChainId = senderChain?.id ?? useChainId();
+
+  const defaultSuccessHandler = useCallback(
+    (receipt: TransactionReceipt | undefined) => {
+      // SW will have txn id so open in wallet
```
dschlabach, line 134: Had a bit of a hard time understanding this context
      // SW will have txn id so open in wallet

src/wallet/components/WalletAdvancedSend/components/SendButton.tsx
```diff
@@ -0,0 +1,185 @@
...
+        address
+      ) {
+        const url = new URL('https://wallet.coinbase.com/assets/transactions');
+        url.searchParams.set('contentParams[txHash]', transactionHash);
+        url.searchParams.set(
+          'contentParams[chainId]',
+          JSON.stringify(senderChain?.id),
+        );
+        url.searchParams.set('contentParams[fromAddress]', address);
+        window.open(url, '_blank', 'noopener,noreferrer');
```
dschlabach, line 149: This is getting a bit nasty, could you move it to a util? We can also add testing then

src/wallet/components/WalletAdvancedSend/components/SendProvider.tsx
```diff
@@ -0,0 +1,282 @@
...
+const SendContext = createContext(emptyContext);
+
+export function useSendContext() {
+  return useContext(SendContext);
+}
+
+export function SendProvider({ children }: SendProviderReact) {
+  const [isInitialized, setIsInitialized] = useState(false);
+
+  // state for ETH balance
```
dschlabach, line 33:
```suggestion
```

src/wallet/components/WalletAdvancedSend/components/SendProvider.tsx
```diff
@@ -0,0 +1,282 @@
...
+      },
+    });
+  const { address: senderAddress, chain: senderChain } = useWalletContext();
+  const { tokenBalances } = useWalletAdvancedContext();
+
+  useEffect(() => {
+    const ethBalance = tokenBalances?.find((token) => token.address === '');
+    if (ethBalance && ethBalance.cryptoBalance > 0) {
+      setEthBalance(
+        Number(ethBalance.cryptoBalance / 10 ** ethBalance.decimals),
```
dschlabach, line 76: `formatUnits` - or maybe make a util that wraps that and converts to `number` as we expect

src/wallet/components/WalletAdvancedSend/components/SendProvider.tsx
```diff
@@ -0,0 +1,282 @@
...
+      );
+    } else {
+      updateLifecycleStatus({
+        statusName: 'fundingWallet',
```
dschlabach, line 80: curious why `fundingWallet`?

src/wallet/components/WalletAdvancedSend/components/SendProvider.tsx
```diff
@@ -0,0 +1,282 @@
...
+  useEffect(() => {
+    if (!selectedToken) {
+      return;
+    }
+    useExchangeRate({
+      token: selectedToken,
+      selectedInputType,
+      setExchangeRate,
+      setExchangeRateLoading,
+    });
```
dschlabach, line 169: Where does this get set?

dschlabach: /and read?

src/wallet/components/WalletAdvancedSend/components/SendProvider.tsx
```diff
@@ -0,0 +1,282 @@
...
+      const truncatedValue = truncateDecimalPlaces(value, 8);
+      setCryptoAmount(truncatedValue);
+      updateLifecycleStatus({
+        statusName: 'amountChange',
+        statusData: {
+          isMissingRequiredField: true,
+          sufficientBalance:
+            Number(value) <=
+            Number(selectedToken?.cryptoBalance) /
+              10 ** Number(selectedToken?.decimals),
```
dschlabach, line 198: formatUnits

src/internal/components/TokenBalance.tsx
brendan-defi: nice, didn't see that, will use

src/internal/components/TokenBalance.tsx
brendan-defi: cool, will do

src/internal/components/TokenBalance.tsx
brendan-defi: can't have a button within a button, and sometimes the whole container is a button

src/token/components/TokenRow.tsx
brendan-defi: some tokens have whitespace at the front / end for ***style*** /s. we don't want that

src/wallet/components/WalletAdvancedSend/components/Send.tsx
brendan-defi: yeah this is exactly the kind of thing i was hoping to get your thoughts on. i can see why this would be confusing. 

is the suggestion to put all the conditionals into the sub-components, so it looks like (not real component names):

```
if (!isInitialized) {
  return null;
}

return (
  <FundWallet />
  <AddressSelection />
  <TokenSelection />
  <AmountInputAndConfirmation />
)
```

or keep the conditionals here, but just pull all the return values into sub-components?

src/wallet/components/WalletAdvancedSend/components/SendAddressSelector.tsx
brendan-defi: don't think it's possible right? I need the `Address` component from identity

src/wallet/components/WalletAdvancedSend/components/SendAddressSelector.tsx
brendan-defi: as teh component name? i like Selector better because it signals that the user needs to interact with it

src/wallet/components/WalletAdvancedSend/components/SendAddressSelector.tsx
brendan-defi: right now i'm taking a props-maxi approach, almost entirely avoiding using context in the subcomponents. i _think_ this is better devex because it enables easier customization. alternative would be to export the provider. not sure what's better

src/wallet/components/WalletAdvancedSend/components/SendAmountInput.tsx
brendan-defi: like just function declaration and then at the bottom do the exports? 

tbh i find that way much more confusing because i like to know immediately whether the thing i'm looking at is "part of the reason for the file" or "just a helper in the file", but not a hill i'll die on

src/wallet/components/WalletAdvancedSend/components/SendButton.tsx
brendan-defi: i'll use the viem method you suggested, i think that'll work here too

src/wallet/components/WalletAdvancedSend/components/SendButton.tsx
brendan-defi: sorry, i just copied directly from Transaction. 

smart wallet transactions don't have the same structure as EOAs. so the way we link to them is different

src/wallet/components/WalletAdvancedSend/components/SendButton.tsx
brendan-defi: yeah, will create a util in transaction for this

src/wallet/components/WalletAdvancedSend/components/SendProvider.tsx
brendan-defi: useExchangeRate sets the exchange rate value.

`exchangeRate` is read by `SendAmountInput`

src/wallet/components/WalletAdvancedSend/types.ts
```diff
@@ -0,0 +1,106 @@
...
+  lifecycleStatus: LifecycleStatus;
+  updateLifecycleStatus: (newStatus: LifecycleStatus) => void;
+
+  // Wallet Context
+  senderAddress: Address | null | undefined;
+  senderChain: Chain | null | undefined;
+  ethBalance: number | undefined;
+  tokenBalances: PortfolioTokenWithFiatValue[] | undefined;
+
+  // Recipient Address Context
```
dschlabach, line 22: I like the comments, they're helpful

src/wallet/components/WalletAdvancedSend/components/SendButton.tsx
brendan-defi: going to remove it, it's not part of the design

src/internal/components/TokenBalance.tsx
alessey: may be better to style the second button so it appears over the top using position relative to avoid having an interactive descendent of a button.  Even with an onClick, a span won't be tabbable without a tabindex or role, both of which would make this invalid.

src/wallet/components/WalletAdvancedSend/components/Send.tsx
```diff
@@ -0,0 +1,123 @@
...
+              <SendAmountInput className="p-0" textClassName="text-4xl" />
+              <SendTokenSelector
+                tokenBalances={tokenBalances}
+                selectedToken={context.selectedToken}
+                handleTokenSelection={context.handleTokenSelection}
+                handleResetTokenSelection={context.handleResetTokenSelection}
+                setSelectedInputType={context.setSelectedInputType}
+                handleCryptoAmountChange={context.handleCryptoAmountChange}
+                handleFiatAmountChange={context.handleFiatAmountChange}
+              />
```
alessey, line 105: why are you passing props into these components instead of just reading from context?

src/wallet/components/WalletAdvancedSend/components/SendAddressInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+} from 'react';
+import type { RecipientAddress } from '../types';
+import { resolveAddressInput } from '../utils/resolveAddressInput';
+import { validateAddressInput } from '../utils/validateAddressInput';
+
+type AddressInputProps = {
+  selectedRecipientAddress: RecipientAddress;
+  recipientInput: string;
+  setRecipientInput: Dispatch<SetStateAction<string>>;
+  setValidatedInput: Dispatch<SetStateAction<RecipientAddress>>;
```
alessey, line 19: if the intentional by passing props is to allow this to be reusable, would be better to make these standard callbacks vs. setState callbacks

src/wallet/components/WalletAdvancedSend/components/SendAddressInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+        'px-4 py-3',
+        className,
+      )}
+    >
+      <span className={cn(color.foreground)}>To</span>
+      <TextInput
+        inputMode="text"
+        placeholder="Basename, ENS, or Address"
+        value={displayValue}
+        inputValidator={() => !!validateAddressInput(recipientInput)}
```
alessey, line 72: inputValidator={(value) => !!validateAddressInput(value)}

src/wallet/components/WalletAdvancedSend/components/SendAmountInput.tsx
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  return (
+    <AmountInputTypeSwitch
+      asset={selectedToken?.symbol ?? ''}
+      fiatAmount={fiatAmount ?? ''}
+      cryptoAmount={cryptoAmount ?? ''}
+      exchangeRate={exchangeRate}
+      exchangeRateLoading={false}
+      currency={'USD'}
```
alessey, line 96: nit: unnecessary {}

src/wallet/components/WalletAdvancedSend/components/SendAddressSelector.tsx
alessey: you'd just need to export the useProvider hook, no? so if someone wanted to create a custom piece, they could import that hook to get the values they needed?

src/wallet/components/WalletAdvancedSend/components/SendAddressSelector.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+import { Address, Avatar, Name } from '@/identity';
+import { background, border, cn, pressable } from '@/styles/theme';
+import type { Address as AddressType, Chain } from 'viem';
+
+type SendAddressSelectorProps = {
+  address: AddressType | null;
+  senderChain: Chain | null | undefined;
+  handleClick: () => Promise<void>;
+};
```
alessey, line 11: className?

src/wallet/components/WalletAdvancedSend/components/SendAmountInput.tsx
```diff
@@ -0,0 +1,101 @@
...
+  } = useSendContext();
+
+  return (
+    <div className="flex h-full w-full flex-1 flex-col items-center justify-center">
+      <div className="flex h-full w-full flex-col justify-center">
+        <AmountInput
+          fiatAmount={fiatAmount ?? ''}
+          cryptoAmount={cryptoAmount ?? ''}
+          asset={selectedToken?.symbol ?? ''}
+          currency={'USD'}
```
alessey, line 36: nit: unnecessary {}

src/wallet/components/WalletAdvancedProvider.tsx
```diff
@@ -56,6 +58,10 @@ export function WalletAdvancedProvider({
     setShowQr,
     isQrClosing,
     setIsQrClosing,
+    showSend,
+    setShowSend,
+    isSendClosing,
+    setIsSendClosing,
```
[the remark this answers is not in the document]
brendan-defi: address [here](https://github.com/coinbase/onchainkit/pull/1874/commits/3c19c3b6435b688b096f651625287d78b25d1748)

src/wallet/components/WalletAdvancedSend/components/Send.tsx
```diff
@@ -0,0 +1,123 @@
...
+    </SendProvider>
+  );
+}
+
+function SendDefaultChildren() {
+  const { chain: senderChain } = useWalletContext();
+  const { tokenBalances } = useWalletAdvancedContext();
+  const context = useSendContext();
+
+  const walletHasEth = context.isInitialized && context.ethBalance > 0.000001;
```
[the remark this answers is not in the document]
brendan-defi: extracted to const file

src/wallet/components/WalletAdvancedSend/components/SendAddressInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+};
+
+export function SendAddressInput({
+  selectedRecipientAddress,
+  recipientInput,
+  setRecipientInput,
+  setValidatedInput,
+  handleRecipientInputChange,
+  className,
+}: AddressInputProps) {
```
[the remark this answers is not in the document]
brendan-defi: updated

src/wallet/components/WalletAdvancedSend/components/SendAddressInput.tsx
brendan-defi: refactored to use context

src/wallet/components/WalletAdvancedSend/components/SendAddressInput.tsx
brendan-defi: updated

src/token/components/TokenBalance.tsx
```diff
@@ -141,3 +139,36 @@ function TokenBalanceContent({
...
+    <button
+      type="button"
+      data-testid="ockTokenBalanceAction"
+      aria-label={actionText}
+      onClick={onActionPress}
+      onKeyDown={(e) => {
+        if (e.key === 'Enter' || e.key === ' ') {
+          onActionPress?.();
+        }
+      }}
```
dschlabach, line 160: I'm a little confused why you need an explicit `onKeyDown` handler? Shouldn't `<button />` handle that out of the box?

brendan-defi:
closing as work completed in [2134](https://github.com/coinbase/onchainkit/pull/2134)

PR #1875, opened 2025-01-24
feat: send utils

brendan-defi:
**What changed? Why?**
Added / updated utilities and hooks to be used by `Send` component, including:
- `getAddress` — added chain check and fixed issue with basename resolver
- `getName` — minor refactor for hygiene
- `isEns` — basic test to see if an input could be an ENS name
- `truncateDecimalPlaces` — rounds and truncates numbers / strings
- `getValidatedAddress` — returns a valid EVM address from a basename, ens, or address input
- `useExchangeRate` — uses `getSwapQuote` to get an exchange rate for a token against USDC, handles state updates
- `useAmountInput` — refactored from FundCard, handles crypto/fiat value input changes

**Notes to reviewers**

**How has it been tested?**


src/identity/utils/isEns.ts
```diff
@@ -0,0 +1,9 @@
+export const isEns = (username: string) => {
+  if (username.endsWith('.base.eth') || username.endsWith('.basetest.eth')) {
+    return false;
+  }
+  if (username.endsWith('.eth') || username.endsWith('.test.eth')) {
+    return true;
+  }
+  return false;
+};
```
cpcramer, line 9:
```suggestion
export const isEns = (username: string) => {
  if (username.endsWith('.eth') || username.endsWith('.test.eth')) {
    return true;
  }
  return false;
};
```

src/internal/utils/getValidatedAddress.ts
```diff
@@ -0,0 +1,20 @@
+import { getAddress } from '@/identity/utils/getAddress';
+import { isBasename } from '@/identity/utils/isBasename';
+import { isEns } from '@/identity/utils/isEns';
+import { isAddress } from 'viem';
+import { base, mainnet } from 'viem/chains';
+
+export async function getValidatedAddress(input: string) {
+  if (isBasename(input) || isEns(input)) {
```
cpcramer, line 8: We should update to call isBasename once and store the response

src/identity/utils/isEns.ts
brendan-defi: no, this doesn't work, as it would return true for basenames `.base.eth` also ends with `.eth`

src/internal/utils/getValidatedAddress.ts
brendan-defi: done

dschlabach (approved):

abcrane123 (approved):

cpcramer (approved):

PR #1878, opened 2025-01-24
chore: dedupe utils

brendan-defi:
**What changed? Why?**
* removed fund-specific `useAmountInput` and `truncateDecimalPlaces`
* moved to / replaced with versions in `src/internal`

**Notes to reviewers**

**How has it been tested?**


src/internal/hooks/useAmountInput.test.tsx
```diff
@@ -23,8 +23,8 @@ describe('useAmountInput', () => {
         result.current.handleFiatChange('100.456');
       });
 
-      expect(defaultProps.setFiatAmount).toHaveBeenCalledWith('100.46');
-      expect(defaultProps.setCryptoAmount).toHaveBeenCalledWith('200.92');
+      expect(defaultProps.setFiatAmount).toHaveBeenCalledWith('100.45');
+      expect(defaultProps.setCryptoAmount).toHaveBeenCalledWith('200.9');
```
rustam-cb, line 27: Seems like rounding is not correct here? 
`100.456` should round to `100.46` right?

brendan-defi: @rustam-cb had to revert those changes because it didn't play nicely with your FundCard. So I went back to your straightforward truncation instead

rustam-cb (dismissed):

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -36,7 +39,12 @@ export const FundCardAmountInput = ({
     currencySpanRef,
   );
 
-  const { handleChange } = useAmountInput();
+  const { handleChange } = useAmountInput({
+    setFiatAmount: setFundAmountFiat,
+    setCryptoAmount: setFundAmountCrypto,
+    selectedInputType: selectedInputType as 'fiat' | 'crypto',
```
dschlabach, line 45: We should try and avoid these assertions at all costs because they really reduce the value prop of using TS. Is there a way we can update the types or add type checks to prevent the assertion?

brendan-defi: this is because when you pull from context all types become `| undefined`. 

i think the best approach is to either leave the type assertion or add an early return. 

@rustam-cb since this is touching your stuff, do you have an aversion to somethng like:
```
if (!selectedInputType) {
  return null;
}
```

rustam-cb: @brendan-defi  Looks like `selectedInputType` does not need to be optional anymore (It is a missed artifact). So we can just remove `?` for that type then.

brendan-defi: makes sense, updating, thanks y'all

rustam-cb (dismissed):

dschlabach (dismissed):

rustam-cb (approved):

PR #1892, opened 2025-01-28
feat: more send utilities

brendan-defi:
**What changed? Why?**
* added `buildSendTransaction` API endpoint
* added `useSendTransaction` hook
* added `validateAddressInput` wallet util

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/api/buildSendTransaction.ts
+++ b/src/api/buildSendTransaction.ts
@@ -1,0 +1,43 @@
+import { encodeFunctionData, erc20Abi } from 'viem';
+import type {
+  BuildSendTransactionParams,
+  BuildSendTransactionResponse,
+} from './types';
+
+export function buildSendTransaction({
+  recipientAddress,
+  tokenAddress,
+  amount,
+}: BuildSendTransactionParams): BuildSendTransactionResponse {
+  // if no token address, we are sending native ETH
+  // and the data prop is empty
+  if (!tokenAddress) {
+    return [
+      {
+        to: recipientAddress,
+        data: '0x',
+        value: amount,
+      },
+    ];
+  }
+
+  try {
+    const transferCallData = encodeFunctionData({
+      abi: erc20Abi,
+      functionName: 'transfer',
+      args: [recipientAddress, amount],
+    });
+    return [
+      {
+        to: tokenAddress,
+        data: transferCallData,
+      },
+    ];
+  } catch (error) {
+    return {
+      code: 'AmBSeTa01', // Api Module Build Send Transaction Error 01
+      error: `Something went wrong: ${error}`,
+      message: 'Could not build transfer transaction',
+    };
+  }
+}
--- a/src/api/buildSendTransaction.ts
+++ b/src/api/buildSendTransaction.ts
@@ -12,13 +12,11 @@ export function buildSendTransaction({
   // if no token address, we are sending native ETH
   // and the data prop is empty
   if (!tokenAddress) {
-    return [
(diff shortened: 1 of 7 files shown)


src/api/buildSendTransaction.ts
```diff
@@ -0,0 +1,43 @@
...
+      abi: erc20Abi,
+      functionName: 'transfer',
+      args: [recipientAddress, amount],
+    });
+    return [
+      {
+        to: tokenAddress,
+        data: transferCallData,
+      },
+    ];
```
alessey, line 35: why does this return an array?

src/internal/hooks/useSendTransaction.tsx
```diff
@@ -0,0 +1,44 @@
...
+    return { calls: [] };
+  }
+
+  if (!token.address) {
+    if (token.symbol !== 'ETH') {
+      return { calls: [] };
+    }
+    const parsedAmount = parseUnits(amount, token.decimals);
+    return {
+      calls: buildSendTransaction({
```
alessey, line 27: this will be a weird response format if there's an error

src/wallet/utils/validateAddressInput.ts
```diff
@@ -0,0 +1,20 @@
...
+    const address = await getAddress({
+      name: input,
+      chain: isBasename(input) ? base : mainnet,
+    });
+    if (address) {
+      return address;
+    }
+  } else if (isAddress(input, { strict: false })) {
+    return input;
+  }
```
alessey, line 18: nit: if isAddress then no need to check if its isBasename or isEns?  moving this first could remove the else if and short circuit faster

src/api/buildSendTransaction.ts
brendan-defi: it doesn't need to, I'll just return the object

src/wallet/utils/validateAddressInput.ts
brendan-defi: makes sense, updated

src/api/buildSendTransaction.ts
```diff
@@ -0,0 +1,39 @@
...
+      args: [recipientAddress, amount],
+    });
+    return {
+      to: tokenAddress,
+      data: transferCallData,
+    };
+  } catch (error) {
+    return {
+      code: 'AmBSeTa01', // Api Module Build Send Transaction Error 01
+      error: `Something went wrong: ${error}`,
```
alessey, line 35: could probably just return error here, i think its implicit that something went wrong

src/internal/hooks/useSendTransaction.tsx
```diff
@@ -0,0 +1,50 @@
...
+    const parsedAmount = parseUnits(amount, token.decimals);
+    const sendTransaction = buildSendTransaction({
+      recipientAddress,
+      tokenAddress: null,
+      amount: parsedAmount,
+    });
+    return sendTransaction;
+  }
+
+  const parsedAmount = parseUnits(amount.toString(), token.decimals);
```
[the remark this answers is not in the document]
brendan-defi: yep, you're right, removing the method

src/api/buildSendTransaction.ts
brendan-defi: ack, will update to `Error: ${error}`

alessey (approved):

abcrane123 (approved):

PR #1893, opened 2025-01-28
chore: move amount input

brendan-defi:
**What changed? Why?**
* moved `AmountInput`, `AmountInputTypeSwitch`, and `CurrencyLabel` to `src/internal`
* moved `useInputResize` to `src/internal`
* refactored `FundCard` components with the new internal versions

**Notes to reviewers**

**How has it been tested?**


src/internal/components/TextInput.tsx
```diff
@@ -15,10 +15,15 @@ type TextInputReact = {
...
   onChange: (s: string) => void;
+  onFocus?: (e: React.FocusEvent<HTMLInputElement>) => void;
   placeholder: string;
   setValue?: (s: string) => void;
   value: string;
   inputValidator?: (s: string) => boolean;
+  /** autocomplete attribute handles browser autocomplete, defaults to 'off' */
+  autoComplete?: string;
+  /** data-1p-ignore attribute handles password manager autocomplete, defaults to true */
+  'data-1p-ignore'?: boolean;
```
dschlabach, line 26: Is there ever a case where we don't want to ignore it? I'd argue for just removing it completely and setting our inputs to always ignore

brendan-defi: i think it's probably right that we always want to ignore, but since this is internal i figured i'd leave us the option. wdyt?

dschlabach: If it's internal + not planned to be used I'd rather remove it since it keeps the props more concise, but I won't block on this issue

dschlabach (approved):

src/internal/components/TextInput.tsx
brendan-defi: thanks, i'll create ticket to remove

PR #1899, opened 2025-01-29
feat: add onchainkit-context

alessey:
**What changed? Why?**
Adding context parameter to track the context in which our rpc-proxy API requests were used.  For example, this will allow us to differentiate in our analytics whether a swap was initiated from a Buy component, Swap component or directly from an API.

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/api/buildMintTransaction.ts
+++ b/src/api/buildMintTransaction.ts
@@ -1,5 +1,5 @@
 import { CDP_MINT_TOKEN } from '../core/network/definitions/nft';
-import { type JSONRPCReferrer, sendRequest } from '../core/network/request';
+import { type JSONRPCContext, sendRequest } from '../core/network/request';
 import type {
   BuildMintTransactionParams,
   BuildMintTransactionResponse,
@@ -10,7 +10,7 @@ import type {
  */
 export async function buildMintTransaction(
   params: BuildMintTransactionParams,
-  _referrer: JSONRPCReferrer = 'api',
+  _context: JSONRPCContext = 'api',
 ): Promise<BuildMintTransactionResponse> {
   const { mintAddress, tokenId, network = '', quantity, takerAddress } = params;
 
@@ -29,7 +29,7 @@ export async function buildMintTransaction(
           tokenId,
         },
       ],
-      _referrer,
+      _context,
     );
     if (res.error) {
       return {
--- a/src/api/buildMintTransaction.ts
+++ b/src/api/buildMintTransaction.ts
@@ -1,5 +1,6 @@
+import { REQUEST_CONTEXT } from '@/core/network/constants';
 import { CDP_MINT_TOKEN } from '../core/network/definitions/nft';
-import { type JSONRPCContext, sendRequest } from '../core/network/request';
+import { sendRequest } from '../core/network/request';
 import type {
   BuildMintTransactionParams,
   BuildMintTransactionResponse,
@@ -10,7 +11,7 @@ import type {
  */
 export async function buildMintTransaction(
   params: BuildMintTransactionParams,
-  _context: JSONRPCContext = 'api',
+  _context: REQUEST_CONTEXT = REQUEST_CONTEXT.API,
 ): Promise<BuildMintTransactionResponse> {
   const { mintAddress, tokenId, network = '', quantity, takerAddress } = params;
 
@@ -29,7 +30,7 @@ export async function buildMintTransaction(
           tokenId,
         },
       ],
-      _context,
+      REQUEST_CONTEXT.API,
     );
     if (res.error) {
       return {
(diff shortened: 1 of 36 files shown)


fakepixels (dismissed):

src/api/buildMintTransaction.ts
```diff
@@ -8,26 +8,29 @@ import type {
...
-export async function buildMintTransaction({
-  mintAddress,
-  tokenId,
-  network = '',
-  quantity,
-  takerAddress,
-}: BuildMintTransactionParams): Promise<BuildMintTransactionResponse> {
+export async function buildMintTransaction(
+  params: BuildMintTransactionParams,
+  _context: JSONRPCContext = 'api',
```
brendan-defi, line 13: for my education, why underscore `_context`?

src/buy/components/BuyProvider.tsx
```diff
@@ -351,14 +351,17 @@ export function BuyProvider({
...
+        const response = await buildSwapTransaction(
+          {
+            amount: from.amount,
+            fromAddress: address,
+            from: from.token,
+            maxSlippage: String(maxSlippage),
+            to: to.token,
+            useAggregator,
+          },
+          'buy',
```
cpcramer, line 363: Perhaps this string can be a const that would add  more context to what its purpose: Ex: `BuyComponentContextID`

src/checkout/utils/handlePayRequest.ts
```diff
@@ -17,7 +17,10 @@ export const handlePayRequest = async ({
     buildPayTransactionParams.productId = productId;
   }
 
-  const response = await buildPayTransaction(buildPayTransactionParams);
+  const response = await buildPayTransaction(
+    buildPayTransactionParams,
+    'checkout',
```
cpcramer, line 22: Same comment above, these strings should be a shared const with added clarity, `CheckoutContextID` or something

src/nft/hooks/useMintDetails.ts
```diff
@@ -1,23 +1,27 @@
...
-        takerAddress,
-        tokenId,
-      });
+      const mintDetails = await getMintDetails(
+        {
+          contractAddress,
+          takerAddress,
+          tokenId,
+        },
+        _context,
```
brendan-defi, line 23: this one not hardcoded because we use this hook in multiple contexts?

src/nft/hooks/useTokenDetails.ts
```diff
@@ -1,21 +1,26 @@
...
-      const tokenDetails = await getTokenDetails({
-        contractAddress,
-        tokenId,
-      });
+      const tokenDetails = await getTokenDetails(
+        {
+          contractAddress,
+          tokenId,
+        },
+        _context,
```
brendan-defi, line 22: same question here bout not hardcoding

src/wallet/hooks/usePortfolio.ts
```diff
@@ -8,17 +9,19 @@ import type { Address } from 'viem';
...
     queryKey: ['usePortfolio', address],
     queryFn: async () => {
-      const response = await getPortfolios({
-        addresses: [address as Address], // Safe to coerce to Address because useQuery's enabled flag will prevent the query from running if address is undefined
-      });
+      const response = await getPortfolios(
+        {
+          addresses: [address as Address], // Safe to coerce to Address because useQuery's enabled flag will prevent the query from running if address is undefined
+        },
+        _context,
```
brendan-defi, line 23: same question about hardcoding

brendan-defi (approved):

src/api/buildMintTransaction.ts
alessey: to try to mark it as internal, i wrote that in the jsdoc as well, this isn't really something consumers should be or need to be setting, but its obviously there.

src/buy/components/BuyProvider.tsx
alessey: yea, good call

src/nft/hooks/useMintDetails.ts
alessey: yes, you could use this hook in isolation, in which case _context defaults to 'hook', but this is also used internally in which case that prop will be set by the consumer

src/internal/hooks/useExchangeRate.tsx
```diff
@@ -32,12 +33,15 @@ export async function useExchangeRate({
...
-      useAggregator: false,
-    });
+    const response = await getSwapQuote(
+      {
+        amount: '1', // hardcoded amount of 1 because we only need the exchange rate
+        from: fromToken,
+        to: toToken,
+        useAggregator: false,
+      },
+      RequestContext.Wallet,
```
brendan-defi, line 43: non-blocking question: when should we allow this to be `hook` vs hardcoded to the "default" component in which it's used?

brendan-defi (approved):

dschlabach (approved):

PR #1900, opened 2025-01-29
feat: Fund Card - Refresh exchange rate regularly

rustam-cb:
**What changed? Why?**
We need to make sure users aren't seeing stale exchange rate on the Fund Card.
Thus we are making a request to fetch fresh exchange rate every 5 seconds while changes are being made with rate limiting strategy `Throttling`. 

- Implement throttling for the onramp exchange rate fetching. 

How it works: 
- Initial Amount updated ---> fetch immediately
- Amount updated again within 5 seconds ---> schedule to fetch ---> fetched after 5 seconds
- Amount updated again after 5 seconds ----> fetch immediately

Screenshot 2025-01-30 at 12 39 05 PM

https://github.com/user-attachments/assets/5640874c-a437-4e9b-9105-ddb734a21c94

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/playground/nextjs-app-router/onchainkit/package.json
+++ b/playground/nextjs-app-router/onchainkit/package.json
@@ -1,11 +1,11 @@
 {
   "name": "@coinbase/onchainkit",
-  "version": "0.36.8",
+  "version": "0.36.9",
   "type": "module",
   "repository": "https://github.com/coinbase/onchainkit.git",
   "license": "MIT",
   "scripts": {
-    "build": "packemon build --addEngines --addFiles --loadConfigs --declaration && npx packemon validate --no-license --no-people --no-repo && tailwindcss -i ./src/styles/index.css -o ./src/tailwind.css --minify && tailwindcss -i ./src/styles/index-with-tailwind.css -o ./src/styles.css --minify",
+    "build": "packemon build --addEngines --addFiles --loadConfigs --declaration && tscpaths -p tsconfig.esm.json -s ./src -o ./esm && npx packemon validate --no-license --no-people --no-repo && tailwindcss -i ./src/styles/index.css -o ./src/tailwind.css --minify && tailwindcss -i ./src/styles/index-with-tailwind.css -o ./src/styles.css --minify",
     "check": "biome check --write .",
     "check:unsafe": "biome check . --fix --unsafe",
     "ci:check": "biome ci --formatter-enabled=false --linter-enabled=false",
@@ -77,6 +77,7 @@
     "rimraf": "^5.0.5",
     "storybook": "^8.2.9",
     "tailwindcss": "^3.4.3",
+    "tscpaths": "^0.0.9",
     "tsup": "^8.3.5",
     "typescript": "~5.3.3",
     "vite": "^5.3.3",
--- a/playground/nextjs-app-router/onchainkit/package.json
+++ b/playground/nextjs-app-router/onchainkit/package.json
@@ -5,7 +5,7 @@
   "repository": "https://github.com/coinbase/onchainkit.git",
   "license": "MIT",
   "scripts": {
-    "build": "packemon build --addEngines --addFiles --loadConfigs --declaration && tscpaths -p tsconfig.esm.json -s ./src -o ./esm && npx packemon validate --no-license --no-people --no-repo && tailwindcss -i ./src/styles/index.css -o ./src/tailwind.css --minify && tailwindcss -i ./src/styles/index-with-tailwind.css -o ./src/styles.css --minify",
+    "build": "packemon build --addEngines --addFiles --loadConfigs --declaration && npx packemon validate --no-license --no-people --no-repo && tailwindcss -i ./src/styles/index.css -o ./src/tailwind.css --minify && tailwindcss -i ./src/styles/index-with-tailwind.css -o ./src/styles.css --minify",
     "check": "biome check --write .",
     "check:unsafe": "biome check . --fix --unsafe",
     "ci:check": "biome ci --formatter-enabled=false --linter-enabled=false",
@@ -77,7 +77,6 @@
     "rimraf": "^5.0.5",
     "storybook": "^8.2.9",
     "tailwindcss": "^3.4.3",
-    "tscpaths": "^0.0.9",
     "tsup": "^8.3.5",
     "typescript": "~5.3.3",
     "vite": "^5.3.3",
--- a/src/fund/components/FundCardAmountInput.tsx
+++ b/src/fund/components/FundCardAmountInput.tsx
@@ -5,7 +5,7 @@ import { useOnrampExchangeRate } from '../hooks/useOnrampExhangeRate';
 import type { FundCardAmountInputPropsReact } from '../types';
 import { useFundContext } from './FundCardProvider';
 
-const THROTTLE_DELAY = 5000;
+const THROTTLE_DELAY_MS = 5000;
 
 export const FundCardAmountInput = ({
   className,
@@ -36,7 +36,7 @@ export const FundCardAmountInput = ({
 
   const throttledFetchExchangeRate = useThrottle(
     fetchExchangeRate,
-    THROTTLE_DELAY,
+    THROTTLE_DELAY_MS,
   );
 
   /**
--- a/src/fund/components/FundCardAmountInput.tsx
+++ b/src/fund/components/FundCardAmountInput.tsx
@@ -1,7 +1,7 @@
 import { AmountInput } from '@/internal/components/amount-input/AmountInput';
 import { useThrottle } from '@/internal/hooks/useThrottle';
 import { useCallback } from 'react';
-import { useOnrampExchangeRate } from '../hooks/useOnrampExhangeRate';
+import { useOnrampExchangeRate } from '../hooks/useOnrampExchangeRate';
 import type { FundCardAmountInputPropsReact } from '../types';
 import { useFundContext } from './FundCardProvider';
 
--- a/src/fund/components/FundCardAmountInput.tsx
+++ b/src/fund/components/FundCardAmountInput.tsx
@@ -1,4 +1,6 @@
 import { AmountInput } from '@/internal/components/amount-input/AmountInput';
+import { useCallback, useEffect, useRef } from 'react';
+import { useOnrampExchangeRate } from '../hooks/useOnrampExhangeRate';
 import type { FundCardAmountInputPropsReact } from '../types';
 import { useFundContext } from './FundCardProvider';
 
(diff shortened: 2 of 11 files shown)


src/fund/components/FundCardProvider.tsx
```diff
@@ -92,40 +94,24 @@ export function FundCardProvider({
     onStatus,
   });
 
-  const fetchExchangeRate = useCallback(async () => {
```
rustam-cb, line 95: Moved this into a separate hook

src/fund/hooks/useOnrampExhangeRate.ts
```diff
@@ -0,0 +1,49 @@
...
+        onError?.({
+          errorType: 'handled_error',
+          code: 'EXCHANGE_RATE_ERROR',
+          debugMessage: err.message,
+        });
+      }
+    }
+  }, [asset, country, subdivision, currency, onError, setExchangeRate]);
+
+  return useMemo(() => ({ fetchExchangeRate }), [fetchExchangeRate]);
```
alessey, line 48: could probably use [tanstack/react-query](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery) here which we use elsewhere in the codebase which includes multiple refetch options

rustam-cb: Does that help with throttle?

alessey: these in particular seem like they'd do what you want
```refetchInterval: number | false | ((query: Query) => number | false | undefined)
Optional
If set to a number, all queries will continuously refetch at this frequency in milliseconds
If set to a function, the function will be executed with the query to compute a frequency
refetchIntervalInBackground: boolean
Optional
If set to true, queries that are set to continuously refetch with a refetchInterval will continue to refetch while their tab/window is in the background
```
```
refetchOnMount: boolean | "always" | ((query: Query) => boolean | "always")
Optional
Defaults to true
If set to true, the query will refetch on mount if the data is stale.
If set to false, the query will not refetch on mount.
If set to "always", the query will always refetch on mount.
If set to a function, the function will be executed with the query to compute the value
```
```
refetchOnWindowFocus: boolean | "always" | ((query: Query) => boolean | "always")
Optional
Defaults to true
If set to true, the query will refetch on window focus if the data is stale.
If set to false, the query will not refetch on window focus.
If set to "always", the query will always refetch on window focus.
If set to a function, the function will be executed with the query to compute the value
```
```
refetchOnReconnect: boolean | "always" | ((query: Query) => boolean | "always")
Optional
Defaults to true
If set to true, the query will refetch on reconnect if the data is stale.
If set to false, the query will not refetch on reconnect.
If set to "always", the query will always refetch on reconnect.
If set to a function, the function will be executed with the query to compute the value
```

rustam-cb: @alessey Looks like those are more like intervals to refetch. My solution is to only refetch while amount input is being changed. So if idle we don't want to keep refetching indefinitely

alessey: doesn't the exchange rate being stale depend more on an interval than on whether the user is actively pressing buttons?

rustam-cb: There are pros and cons to both solutions: 
`Interval` solution 
Pros: 
- Always fresh exchange rate (useful when you enter the amount leave and come back after some time to press `Buy`)
Cons: 
- Excessive API requests while FundCard is open.


`Throttle` solution
Pros: 
- Minimal API requests (only when user is active on the component)

Cons:
- Exchange rate will be stale if user enters the amount then goes away for some time and then come back


Given that I have decided to move with the `throttle` solution

src/fund/components/FundCardAmountInput.tsx
```diff
@@ -1,7 +1,12 @@
 import { AmountInput } from '@/internal/components/amount-input/AmountInput';
+import { useThrottle } from '@/internal/hooks/useThrottle';
+import { useCallback } from 'react';
+import { useOnrampExchangeRate } from '../hooks/useOnrampExhangeRate';
 import type { FundCardAmountInputPropsReact } from '../types';
 import { useFundContext } from './FundCardProvider';
 
+const THROTTLE_DELAY = 5000;
```
cpcramer, line 8:
```suggestion
const THROTTLE_DELAY_MS = 5000;
```

src/fund/hooks/useOnrampExhangeRate.ts
```diff
@@ -0,0 +1,49 @@
...
+  asset: string;
+  currency: string;
+  country: string;
+  subdivision?: string;
+  setExchangeRate: (exchangeRate: number) => void;
+  onError?: (error: OnrampError) => void;
+}) => {
+  const fetchExchangeRate = useCallback(async () => {
+    try {
+      console.log('fetching exchange rate');
```
alessey, line 22: was this left intentionally?

cpcramer: nit: remove

rustam-cb: Nope. This was added for debug and successfully forgotten about. 

Removed

src/fund/components/FundCardAmountInput.tsx

src/fund/hooks/useOnrampExhangeRate.ts
rustam-cb: Done

src/internal/hooks/useThrottle.ts
```diff
@@ -0,0 +1,92 @@
+import { useCallback, useEffect, useRef } from 'react';
+
+type AnyFunction = (...args: unknown[]) => unknown;
+
+/**
```
cpcramer, line 5: Maybe we can reduce the comments by linking to a helpful page with more context on throttling?

cpcramer:
I updated the PR title to match our style 😄 

Can we also add a changeset

src/internal/hooks/useThrottle.ts
rustam-cb: Updated

cpcramer (dismissed):

playground/nextjs-app-router/onchainkit/package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "@coinbase/onchainkit",
-  "version": "0.36.8",
+  "version": "0.36.9",
   "type": "module",
   "repository": "https://github.com/coinbase/onchainkit.git",
   "license": "MIT",
   "scripts": {
-    "build": "packemon build --addEngines --addFiles --loadConfigs --declaration && npx packemon validate --no-license --no-people --no-repo && tailwindcss -i ./src/styles/index.css -o ./src/tailwind.css --minify && tailwindcss -i ./src/styles/index-with-tailwind.css -o ./src/styles.css --minify",
+    "build": "packemon build --addEngines --addFiles --loadConfigs --declaration && tscpaths -p tsconfig.esm.json -s ./src -o ./esm && npx packemon validate --no-license --no-people --no-repo && tailwindcss -i ./src/styles/index.css -o ./src/tailwind.css --minify && tailwindcss -i ./src/styles/index-with-tailwind.css -o ./src/styles.css --minify",
```
brendan-defi, line 8: is this change intentional?

playground/nextjs-app-router/onchainkit/package.json
```diff
@@ -77,6 +77,7 @@
     "rimraf": "^5.0.5",
     "storybook": "^8.2.9",
     "tailwindcss": "^3.4.3",
+    "tscpaths": "^0.0.9",
```
brendan-defi, line 80: what is `tscpaths`?

src/fund/hooks/useOnrampExhangeRate.ts
brendan-defi: @rustam-cb i think you could use useQuery's staleTime. Data won't refetch during the specified interval

playground/nextjs-app-router/onchainkit/package.json
rustam-cb: Not intentional. Not sure what that is... Thanks for catching. Removed

playground/nextjs-app-router/onchainkit/package.json
rustam-cb: I don't know where that came from. Removed

src/fund/hooks/useOnrampExhangeRate.ts
rustam-cb: @brendan-defi From what I understood the `staleTime` only allows you to set if the data is stale so that refetch on mount or window focus does not refetch if its stale.  But it won't prevent refetchInterval from fetching again

brendan-defi: but i don't htink you'd set a refetchInterval, right? you'll just fetch when the user makes their request.

and i'm not sure about this, but the docs seem to suggest that the cached value would be returned if the user queries again within the non-stale-period

playground/nextjs-app-router/onchainkit/package.json
brendan-defi: happening to me too. not sure what's up. will try to figure it out tomorrow

brendan-defi (dismissed):
still curious about whether we can use tanstack query, but won't block on it

src/fund/hooks/useOnrampExhangeRate.ts
```diff
@@ -0,0 +1,48 @@
...
+    } catch (err) {
+      if (err instanceof Error) {
+        console.error('Error fetching exchange rate:', err);
+        onError?.({
+          errorType: 'handled_error',
+          code: 'EXCHANGE_RATE_ERROR',
+          debugMessage: err.message,
+        });
+      }
+    }
```
alessey, line 44: possible to keep this as a pure function and just have it return exchange rate | error?

src/fund/hooks/useOnrampExhangeRate.ts
```diff
@@ -0,0 +1,48 @@
...
+        country,
+        subdivision,
+      });
+
+      setExchangeRate(
+        Number(quote.purchaseAmount.value) /
+          Number(quote.paymentSubtotal.value),
+      );
+    } catch (err) {
+      if (err instanceof Error) {
```
alessey, line 36: is it possible to catch an error here that's a string?

rustam-cb: Updated this.

src/fund/hooks/useOnrampExhangeRate.ts
rustam-cb: We already have a pure function `fetchOnrampQuote`. The hook I created is to be reused in two different places.

brendan-defi (approved):

cpcramer (approved):

PR #1901, opened 2025-01-29
feat: Add DropdownMenu UI Primitive

cpcramer:
**What changed? Why?**

Add new UI primitive - `DropdownMenu`

- Menu positioning relative to trigger
- Focus management
- Click outside and escape key dismissal
- Keyboard navigation
 - Proper ARIA attributes for accessibility
 
Why? We are standardizing common UI primitives across OnchainKit to ensure consistent accessibility, keyboard, navigation and styling across OnchainKit components. 

Add `getHorizontalPosition` utility function.

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/internal/components/primitives/DropdownMenu.tsx
+++ b/src/internal/components/primitives/DropdownMenu.tsx
@@ -1,5 +1,6 @@
 import { useTheme } from '@/internal/hooks/useTheme';
 import { getHorizontalPosition } from '@/internal/utils/getHorizontalPosition';
+import { zIndex } from '@/styles/constants';
 import { cn } from '@/styles/theme';
 import type React from 'react';
 import { useCallback, useEffect, useRef } from 'react';
@@ -84,13 +85,19 @@ export function DropdownMenu({
   }
 
   const dropdownMenu = (
-    <div className={cn(componentTheme, 'pointer-events-none fixed z-50')}>
+    <div
+      className={cn(
+        componentTheme,
+        zIndex.dropdown,
+        'pointer-events-none fixed',
+      )}
+    >
       <FocusTrap active={isOpen}>
         <DismissableLayer onDismiss={onClose} triggerRef={trigger}>
           <div
             ref={contentRef}
             className={cn('pointer-events-auto fixed')}
-            role="menu"
+            role="listbox"
             data-testid="ockDropdownMenu"
             aria-label={ariaLabel}
           >
--- a/src/internal/components/primitives/DropdownMenu.tsx
+++ b/src/internal/components/primitives/DropdownMenu.tsx
@@ -1,0 +1,105 @@
+import { useTheme } from '@/internal/hooks/useTheme';
+import { getHorizontalPosition } from '@/internal/utils/getHorizontalPosition';
+import { cn } from '@/styles/theme';
+import type React from 'react';
+import { useCallback, useEffect, useRef } from 'react';
+import { createPortal } from 'react-dom';
+import { DismissableLayer } from './DismissableLayer';
+import { FocusTrap } from './FocusTrap';
+
+type DropdownMenuProps = {
+  /** Determines how the menu aligns with the trigger element */
+  align?: 'start' | 'center' | 'end';
+  /** Content of the dropdown menu */
+  children: React.ReactNode;
+  isOpen?: boolean;
+  /** Space between the trigger and menu in pixels */
+  offset?: number;
+  /** Callback for when the dropdown should close */
+  onClose?: () => void;
+  /** The element that triggers the dropdown menu */
+  trigger: React.RefObject<HTMLElement>;
+  'aria-label'?: string;
+};
+
+/**
+ * DropdownMenu primitive that handles:
+ * - Menu positioning relative to trigger
+ * - Focus management
+ * - Click outside and escape key dismissal
+ * - Keyboard navigation
+ * - Proper ARIA attributes for accessibility
+ */
+export function DropdownMenu({
+  align = 'start',
+  children,
+  isOpen,
+  offset = 8,
+  onClose,
+  trigger,
+  'aria-label': ariaLabel,
+}: DropdownMenuProps) {
+  const componentTheme = useTheme();
+  const contentRef = useRef<HTMLDivElement>(null);
+
+  const updatePosition = useCallback(() => {
+    if (!trigger?.current || !contentRef.current) {
+      return;
+    }
+
+    const triggerRect = trigger.current.getBoundingClientRect();
+    const menuRect = contentRef.current.getBoundingClientRect();
+
+    if (!triggerRect || !menuRect) {
+      return;
+    }
+
+    const left = getHorizontalPosition({
+      triggerRect,
+      contentRect: menuRect,
+      align,
+    });
+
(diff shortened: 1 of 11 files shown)


src/internal/utils/getHorizontalPosition.ts
```diff
@@ -0,0 +1,40 @@
...
+      if (triggerRect.width === 0) {
+        return triggerRect.left - contentRect.width / 2;
+      }
+      return triggerRect.left + (triggerRect.width - contentRect.width) / 2;
+    }
+    case 'end': {
+      // Handle zero-width content case
+      if (contentRect.width === 0) {
+        return triggerRect.right;
+      }
```
alessey, line 36: should this handle triggerRect.right === 0 instead of contentRect?

src/internal/utils/getHorizontalPosition.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      right: 200,
+      width: 100,
+      top: 0,
+      bottom: 0,
+      height: 0,
+      x: 0,
+      y: 0,
+      toJSON: serialized.toJSON,
+    });
+  });
```
alessey, line 147: what is this testing?

src/internal/components/primitives/DropdownMenu.tsx
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  const dropdownMenu = (
+    <div className={cn(componentTheme, 'pointer-events-none fixed z-50')}>
+      <FocusTrap active={isOpen}>
+        <DismissableLayer onDismiss={onClose} triggerRef={trigger}>
+          <div
+            ref={contentRef}
+            className={cn('pointer-events-auto fixed')}
+            role="menu"
```
alessey, line 93: is this role correct?  not listbox?

src/internal/utils/getHorizontalPosition.ts
cpcramer: refactored and removed this - seemed like overkill for our use

src/internal/utils/getHorizontalPosition.test.ts
cpcramer: Removed as part of refactor simplifying func

src/internal/components/primitives/DropdownMenu.tsx
```diff
@@ -0,0 +1,105 @@
...
+      return;
+    }
+
+    updatePosition();
+    window.addEventListener('resize', updatePosition);
+    window.addEventListener('scroll', updatePosition);
+
+    return () => {
+      window.removeEventListener('resize', updatePosition);
+      window.removeEventListener('scroll', updatePosition);
```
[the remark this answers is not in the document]
cpcramer: We still need these listeners to keep the dropdown aligned with its trigger when scrolling or resizing

src/internal/components/primitives/DropdownMenu.tsx
cpcramer: good call, updated!

src/internal/components/primitives/DropdownMenu.tsx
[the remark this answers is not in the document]
cpcramer: Yes I believe that relative positioning would remove the need for these listeners. 

Radix-UI uses this current approach though

src/internal/components/primitives/DropdownMenu.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { FocusTrap } from './FocusTrap';
+
+type DropdownMenuProps = {
+  /** Determines how the menu aligns with the trigger element */
+  align?: 'start' | 'center' | 'end';
+  /** Content of the dropdown menu */
+  children: React.ReactNode;
+  isOpen?: boolean;
+  /** Space between the trigger and menu in pixels */
+  offset?: number;
```
brendan-defi, line 17: possible to make this based on rem instead?

src/internal/components/primitives/DropdownMenu.tsx
```diff
@@ -0,0 +1,105 @@
...
+      window.removeEventListener('scroll', updatePosition);
+    };
+  }, [isOpen, updatePosition]);
+
+  if (!isOpen) {
+    return null;
+  }
+
+  const dropdownMenu = (
+    <div className={cn(componentTheme, 'pointer-events-none fixed z-50')}>
```
brendan-defi, line 87: let's use the zIndex values from styles/constants

src/internal/components/primitives/DropdownMenu.tsx
```diff
@@ -0,0 +1,105 @@
...
+      <FocusTrap active={isOpen}>
+        <DismissableLayer onDismiss={onClose} triggerRef={trigger}>
```
brendan-defi, line 89: should we add a prop to pass through to DismissableLayer's `disableOutsideClick`? i could see that being useful for menus you want to stay open until a close button is pressed

src/internal/components/primitives/Popover.tsx
```diff
@@ -169,7 +169,7 @@ export function Popover({
   }
 
   const popover = (
-    <div className={cn(componentTheme, 'fixed z-50', 'pointer-events-none')}>
+    <div className={cn(componentTheme, 'pointer-events-none fixed z-50')}>
```
brendan-defi, line 172: same zIndex comment

src/internal/components/primitives/DropdownMenu.tsx
cpcramer: Following Radix-ui's lead using pixels. Also, since the offset will always likely be a very small number, I feel pixels are easier to work with

src/internal/components/primitives/DropdownMenu.tsx
cpcramer: Updated the z index constants - I don't think that they were working as expected

src/internal/components/primitives/DropdownMenu.tsx
cpcramer: I think we should add that prop if we see demand for it - I don't see any current use case for it in ock

brendan-defi (approved):

src/token/components/TokenSelectDropdown.tsx
```diff
@@ -12,28 +12,58 @@ export function TokenSelectDropdown({
...
+    // NOTE: this ensures that handleBlur doesn't get called on initial mount
+    //       We need to use non-div elements to properly handle onblur events
+    setTimeout(() => {
+      document.addEventListener('click', handleBlur);
+    }, 0);
+
+    return () => {
+      document.removeEventListener('click', handleBlur);
+    };
+  }, [handleBlur]);
```
alessey, line 49: can this useOutsideClick?  why is this using blur instead?

cpcramer: It was originally using the blur terminology prior to the implementation - I've split up the PRs so TokenSelectDropdown update happens here [feat: Implement DropdownMenu Primitive in TokenSelectDropdown](https://github.com/coinbase/onchainkit/pull/1903#top)

cpcramer: This code is all removed as a part of the implementation PR

dschlabach (approved):

PR #1924, opened 2025-02-05
feat: make Earn component work with arbitrary vaults

dschlabach:
**What changed? Why?**
Previously we had built Earn to only work with USDC vaults. This PR updates token logic to be used with Morpho vaults that has non-USDC asset tokens.

**Notes to reviewers**

**How has it been tested?**


src/earn/components/DepositDetails.tsx
```diff
@@ -1,26 +1,39 @@
...
+  const { apy, vaultToken } = useEarnContext();
 
-  // TODO: update token when we have logic to fetch vault info
   return (
     <EarnDetails
       className={className}
-      token={usdcToken}
-      tag={tag}
+      token={vaultToken}
+      tag={<ApyTag apy={apy} />}
```
dschlabach, line 36: Doing this so that we can show a skeleton when loading

src/internal/utils/formatPercent.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Internal
+ * @param value
+ * @param decimals
+ * @returns
+ */
+export function formatPercent(value: number, decimals = 2) {
```
dschlabach, line 7: We can have an internal/utils/formatters and this can go in there, along with other number formatters

src/earn/components/EarnProvider.tsx
```diff
@@ -22,12 +22,21 @@ export function EarnProvider({ vaultAddress, children }: EarnProviderReact) {
   const [depositAmount, setDepositAmount] = useState('');
   const [withdrawAmount, setWithdrawAmount] = useState('');
 
-  const { convertedBalance } = useGetTokenBalance(address, usdcToken);
+  const { asset, assetDecimals, assetSymbol, balance, totalApy } =
+    useMorphoVault({
+      vaultAddress,
+      address,
+    });
+  const vaultToken = asset
```
alessey, line 30: asset could be null here?  when would that happen?

src/earn/components/DepositDetails.tsx
alessey: ?  looks like its just passed through and displayed?

src/earn/components/WithdrawBalance.tsx
```diff
@@ -20,10 +21,17 @@ export function WithdrawBalance({ className }: WithdrawBalanceReact) {
     return getTruncatedAmount(depositedAmount.toString(), 4);
   }, [depositedAmount]);
 
+  const title = useMemo(() => {
+    if (!vaultToken) {
+      return <Skeleton className="h-6 w-24" />;
+    }
+    return `${balance} ${vaultToken?.symbol}`;
```
alessey, line 28: is it possible to have vaultToken, but not a valid balance yet?  i.e. showing 0 ASSET and then updating?

src/earn/components/DepositDetails.tsx
alessey: oh, i see you are passing different nodes for tag based on type

src/earn/utils/getToken.ts
```diff
@@ -0,0 +1,44 @@
...
+  name,
+  decimals,
+}: { address: Address; symbol?: string; name?: string; decimals?: number }):
+  | Token
+  | undefined {
+  const token = baseTokens.find((token) => token.address === address);
+  if (token) {
+    return token;
+  }
+  if (symbol && name && decimals) {
```
alessey, line 39: do you get this info back from morpho?  is the only Token field you're actually looking up the imageUrl?

src/earn/components/EarnProvider.tsx
dschlabach: We can do a better job naming things, but `asset` is the vault's asset token. We fetch the vault data and so, while fetching that data, `asset` could be `undefined`

src/earn/components/WithdrawBalance.tsx
dschlabach: Yes, that's a good point. `balance` comes from an RPC read. I'll update this so that both have to come in before we show data

src/earn/utils/getToken.ts
dschlabach: Yes. The only reason for this helper function is to move the following logic out of `EarnProvider`

1. finding the token based off its address
2. constructing a fallback of type `Token` to be passed to `TokenChip`

src/earn/components/EarnProvider.tsx
alessey: vaultToken = asset ?? getToken()?

dschlabach: bad naming, but `asset` is the token address of the vault (`${0x}string` | `undefined`). So what works is:

`vaultToken = asset && getToken()`

src/earn/components/DepositBalance.tsx
```diff
@@ -1,29 +1,42 @@
...
       return '0';
     }
     return getTruncatedAmount(convertedBalance.toString(), 6);
   }, [convertedBalance]);
 
+  const title = useMemo(() => {
+    if (!vaultToken || !balance) {
+      return (
+        <Skeleton
+          className={cn('!bg-[var(--ock-bg-alternate-active)] h-6 w-24')}
```
brendan-defi, line 29: overriding the background because you don't want the shimmer?

src/earn/components/DepositDetails.tsx
```diff
@@ -1,27 +1,39 @@
...
+        background.alternate,
+        'flex items-center justify-center rounded-full p-1 px-3',
+      )}
+    >
+      {`APY ${formatPercent(Number(getTruncatedAmount(apy.toString(), 3)))}`}
+    </div>
+  ) : (
+    <Skeleton className="!rounded-full h-7 min-w-28" />
+  );
+}
```
brendan-defi, line 27: nit: i generally prefer early returns to ternary statements like this, especially when the early return is because of malformed data

src/earn/components/EarnDetails.tsx
```diff
@@ -1,16 +1,8 @@
...
-export function EarnDetails({
-  className,
-  token,
-  tag,
-  tagVariant = 'default',
-}: EarnDetailsReact) {
-  if (!token) {
-    return null;
-  }
+export function EarnDetails({ className, token, tag }: EarnDetailsReact) {
```
brendan-defi, line 5: nit: add space between imports and fn

src/token/constants.ts
```diff
@@ -40,3 +40,65 @@ export const daiToken: Token = {
...
+  name: 'EURC',
+  address: '0x60a3E35Cc302bFA44Cb288Bc5a4F316Fdb1adb42',
+  symbol: 'EURC',
+  decimals: 6,
+  image:
+    'https://d3r81g40ycuhqg.cloudfront.net/wallet/wais/54/f4/54f4216472dd25b1ffb5caf32cc0d81f645c84be166cd713f759a80f05a1418f-M2YxNTczYTItNjk3YS00N2FiLThkZjktYzBiYzExZTk1ZTFj',
+  chainId: base.id,
+};
+
+const baseTokens = [
```
brendan-defi, line 92: what is meant by "base" in `baseTokens`?

src/earn/components/EarnDetails.tsx
cpcramer, line 5: nit: space add space in between function and imports

src/earn/utils/getToken.ts
```diff
@@ -0,0 +1,30 @@
+import type { Token } from '@/token';
+import { baseTokens } from '@/token/constants';
+import type { Address } from 'viem';
+import { base } from 'viem/chains';
+
+export function getToken({
```
brendan-defi, line 6: getToken feels like the wrong name here, because you're not really "getting" anything. transform? validate? sanitize?

src/earn/utils/getToken.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (token) {
+    return token;
+  }
+  if (symbol && name && decimals) {
+    return {
+      address,
+      name,
+      symbol,
+      decimals,
+      image: null,
```
brendan-defi, line 24: should we put some sort of placeholder image (eg. blue circle) here?

cpcramer (approved):

src/earn/utils/getToken.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      address: tokenDetails.address,
+      name: tokenDetails.name,
+      symbol: tokenDetails.symbol,
+      decimals: tokenDetails.decimals,
+      image: null,
+      chainId: base.id,
+    });
+  });
+
+  it('returns undefined when address is not provided', () => {
```
brendan-defi, line 32: i must be mistaken because this test is passing, but i don't see how getTokens would fail if symbol, name, and decimals are provided but not address

cpcramer:
nit: Add PR what and why in the description

src/earn/types.ts
```diff
@@ -1,5 +1,6 @@
 import type { Token } from '@/token';
 import type { Call } from '@/transaction/types';
+import type React from 'react';
```
brendan-defi, line 3: is this just for ReactNode? maybe better to import type ReactNode in that case?

src/earn/hooks/useMorphoVault.ts
```diff
@@ -45,29 +46,25 @@ export function useMorphoVault({
...
-      enabled: !!address,
+      enabled: !!vaultAddress,
     },
   });
 
-  const { data: tokenDecimals } = useReadContract({
-    abi: erc20Abi,
-    address: data?.[0].result,
-    functionName: 'decimals',
+  // Fetching separately because user may not be connected
```
brendan-defi, line 60: value of fetching separately is that you can display the vault info to the user even if they aren't connected?

src/earn/components/EarnProvider.tsx
```diff
@@ -22,12 +22,21 @@ export function EarnProvider({ vaultAddress, children }: EarnProviderReact) {
   const [depositAmount, setDepositAmount] = useState('');
   const [withdrawAmount, setWithdrawAmount] = useState('');
 
-  const { convertedBalance } = useGetTokenBalance(address, usdcToken);
+  const { asset, assetDecimals, assetSymbol, balance, totalApy } =
+    useMorphoVault({
+      vaultAddress,
+      address,
+    });
+  const vaultToken =
```
brendan-defi, line 30: nit: ternary seems easier to reason about here

src/earn/components/DepositDetails.tsx
dschlabach: @alessey and I were talking about extracting this Skeleton logic into a Skeleton component that handles the ternaries. So you'd have something like this: 
```
<Skeleton loading={boolean} loader={React.ReactNode // with deafult}>
{children}
</Skeleton >
```

Could we implement in a future PR?

src/earn/components/DepositBalance.tsx
dschlabach: Tried, unfortunately we have serious trouble overriding those styles with `cn()`. This is really ugly but it's the only way to override the background color.

src/token/constants.ts
dschlabach: Base chain

src/earn/utils/getToken.ts
dschlabach: Open to alternatives, but I don't think transform/validate/sanitize convey any better. I am "getting" the token, just from an array of known tokens

src/earn/components/DepositDetails.tsx
```diff
@@ -1,27 +1,39 @@
 import { getTruncatedAmount } from '@/earn/utils/getTruncatedAmount';
-import { usdcToken } from '@/token/constants';
-import { useMemo } from 'react';
+import { Skeleton } from '@/internal/components/Skeleton';
+import { formatPercent } from '@/internal/utils/formatPercent';
+import { background } from '@/styles/theme';
+import { color } from '@/styles/theme';
+import { text } from '@/styles/theme';
+import { cn } from '@/styles/theme';
```
alessey, line 7: nit: combine

src/earn/utils/getToken.ts
dschlabach: This conforms to the `Token` type which is required by `TokenChip`. `TokenChip` handles the fallback image and expects null for image if we want a fallback

src/earn/components/DepositDetails.tsx
alessey: I agree with early returns, although in this case, the issue is bigger, for instance if you had a wrapper div then loading ? stuff : other, to split out to an early return would require duplicating the wrapper, when we fix the bigger issue, this goes away.

src/earn/utils/getToken.test.ts
dschlabach: What do you mean by "fail"? Address is a required parameter so it should never be called without one

src/earn/types.ts
dschlabach: Yeah but we don't have a consistent pattern here. Searching the codebase, there's a ~65/35 split between `React.ReactNode` and `ReactNode` usage

src/earn/hooks/useMorphoVault.ts
dschlabach: Yeah, I think I could make it work in one go but it does feel off to re-fetch all that data when the user connects + also it seems off to fetch something that won't work (balance of address = '')

cpcramer (approved):

src/earn/components/DepositDetails.tsx
brendan-defi: yeah i agree, when there's a wrapper makes sense to have embedded ternary. in this case there's not though

brendan-defi: also agree with the update to Skeleton, that would be really nice (and seems like it would also address the wrapper issues)

src/earn/components/DepositBalance.tsx
brendan-defi: makes sense, thanks

src/earn/hooks/useMorphoVault.ts
brendan-defi: yeah, makes sense, was just curious. although, didn't you refactor above so that you don't show anythiing until the user is connected and we get their balance?

src/earn/types.ts
brendan-defi: fair enough

src/earn/utils/getToken.test.ts
brendan-defi: > Address is a required parameter so it should never be called without one

agreed, but how is getToken actually returning undefined if there isn't an address?

IIUC, first we check the baseTokens for a matching address. if we don't find one, then we see if symbol, name, and decimals exist and if they do return an image-less Token. 

then we return undefined.

src/earn/utils/getToken.ts
brendan-defi: okay nice, thanks

src/token/constants.ts
brendan-defi: sorta feels like the wrong name then. why is this array `baseTokens`? should be somethiing about morpho, right? so like, `morphoBaseTokens`?

brendan-defi (approved):

alessey (dismissed):

src/earn/components/EarnProvider.tsx
brendan-defi: ha, sorry, maybe just me, but i appreciate the change back thanks Dan 🤗

src/earn/utils/getToken.test.ts
dschlabach: If address isn't in `baseTokens` and we don't have `symbol`, `name`, `decimals` then I wanted to return something.

This is only going to happen if we have address come through and it's not in our known tokens list, and there's an error with Morpho's API (which would cause `symbol`, `name`, and/or `decimals` to be undefined)

dschlabach: Are you saying I should add a check for !address to handle cases where people aren't using TypeScript?

src/token/constants.ts
dschlabach: These are just tokens on Base, not anything related to Morpho per se. The reason I changed it from `tokens` was because I didn't want people using this array for non-Base network tokens.

src/earn/hooks/useMorphoVault.ts
dschlabach: You're right, that's actually a bug - I'll revert to what I had

src/earn/components/WithdrawBalance.tsx
dschlabach: Nvm, reverting this - default balance is 0

alessey (approved):

cpcramer (approved):

src/earn/utils/getToken.test.ts
brendan-defi: no, not saying you should change anything. i don't think it's a big deal not to have early return if address isn't provided, though i'm not opposed to that either.

i was really trying to figure out how this test was passing, but i see now that it's because you're only passing in undefined address, so since symbol, name, decimal aren't present either we get undefined return

PR #1928, opened 2025-02-06
feat: add Lifecycle Statuses to Earn

dschlabach:
**What changed? Why?**
This PR adds LifecycleStatuses to `Earn` via the `useLifecycleStatus` hook. It includes lifecycle statuses for changing the amount in the text input, as well as surfaces all relevant lifecycle statuses from the underlying `Transaction` component.

**Notes to reviewers**

**How has it been tested?**


cpcramer (approved):

src/earn/components/WithdrawButton.tsx
```diff
@@ -1,17 +1,29 @@
...
+    },
+    [updateLifecycleStatus],
+  );
+
   return (
-    <Transaction className={className} calls={withdrawCalls}>
+    <Transaction
+      className={className}
+      calls={withdrawCalls}
+      onStatus={handleOnStatus}
```
alessey, line 25: onStatus={updateLifecycleStatus}?

src/earn/types.ts
```diff
@@ -104,3 +107,22 @@ export type DepositButtonReact = {
...
+      statusData: null;
+    }
+  | {
+      statusName: 'amountChange';
+      statusData: {
+        amount: string;
+        token: Token;
+      };
+    }
+  | TransactionLifecycleStatus;
```
alessey, line 128: will this fire multiple inits?   i.e. this init and then TransactionLifecycleStatus init?

dschlabach: Yeah it's a good point - I updated my handler in DepositButton and WithdrawButton to prevent that, but it doesn't affect the types

cpcramer (dismissed):

src/earn/types.ts
alessey: yea, i just happened to be here when i thought about it :)

src/earn/types.ts
```diff
@@ -35,6 +36,8 @@ export type EarnContextType = {
   withdrawAmount: string;
   setWithdrawAmount: (amount: string) => void;
   withdrawCalls: Call[];
+  lifecycleStatus: LifecycleStatus;
+  updateLifecycleStatus: (status: LifecycleStatus) => void;
```
alessey, line 40: updateLifecycleStatus: (status: LifecycleStatusUpdate<LifecycleStatus>)  => void

brendan-defi (approved):

PR #1938, opened 2025-02-07
fix: reset amount inputs after successful deposit/withdrawal

dschlabach:
**What changed? Why?**
This PR includes the following behavior changes for Earn:
- After a successful deposit or withdrawal, reset the inputs and re-fetch the relevant balances
- Re-fetch user balances when user navigates between tabs
- Use `redeem` method instead of `withdraw` method on vault contract for greater decimal precision

**Notes to reviewers**

**How has it been tested?**


src/earn/components/WithdrawButton.tsx
```diff
@@ -41,11 +35,35 @@ export function WithdrawButton({ className }: WithdrawButtonReact) {
     [updateLifecycleStatus],
   );
 
+  const handleOnSuccess = useCallback(
+    (res: TransactionResponse) => {
+      console.log('res:', res);
```
cpcramer, line 40: nit: remove

src/earn/utils/buildWithdrawFromMorphoTx.ts
```diff
@@ -16,9 +16,10 @@ export function buildWithdrawFromMorphoTx({
   amount,
   receiverAddress,
 }: WithdrawFromMorphoParams): Call[] {
+  console.log('amount:', amount);
```
cpcramer, line 19: nit: remove

src/earn/utils/buildWithdrawFromMorphoTx.ts
```diff
@@ -16,9 +16,10 @@ export function buildWithdrawFromMorphoTx({
...
   const withdrawTxData = encodeFunctionData({
     abi: MORPHO_VAULT_ABI,
-    functionName: 'withdraw',
+    functionName: 'redeem', // redeem is the number of *shares*, withdraw is the number of *assets*
```
cpcramer, line 22: I'm still unsure about the difference between `shares` and `assets` after reading this. That doesn't need to be clarified here, just calling out

cpcramer (dismissed):

src/earn/utils/buildWithdrawFromMorphoTx.ts
dschlabach: Fair point, I have another PR in the pipe to clean up some of this stuff, I can try and improve it then

cpcramer (approved):

PR #1941, opened 2025-02-07
fix: validate Earn input

dschlabach:
**What changed? Why?**
This PR includes validation for disabling the deposit and withdraw buttons as well as showing helpful errors to users:

DS 2025-02-07 at 17 30 37@2x

DS 2025-02-07 at 17 30 33@2x

**Notes to reviewers**

**How has it been tested?**


alessey:
since you start out with amount 0, will you now see the disabled button and error initially instead of seeing "deposit"?

dschlabach:
@alessey no, it only happens on change so the initial state won't be an error

alessey (approved):

cpcramer (approved):

alessey (approved):

PR #1964, opened 2025-02-11
feat: Add Swap Telemetry

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**
tested locally in playground

Screenshot 2025-02-25 at 9 38 46 AM

Screenshot 2025-02-25 at 9 39 12 AM

Screenshot 2025-02-25 at 9 40 03 AM

### DIFF
--- a/src/swap/components/SwapAmountInput.tsx
+++ b/src/swap/components/SwapAmountInput.tsx
@@ -1,5 +1,7 @@
 'use client';
 import { useCallback, useEffect, useMemo } from 'react';
+import { useAnalytics } from '../../core/analytics/hooks/useAnalytics';
+import { SwapEvent } from '../../core/analytics/types';
 import { TextInput } from '../../internal/components/TextInput';
 import { useValue } from '../../internal/hooks/useValue';
 import { getRoundedAmount } from '../../internal/utils/getRoundedAmount';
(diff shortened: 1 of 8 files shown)


src/swap/components/SwapAmountInput.tsx
```diff
@@ -51,12 +54,28 @@ export function SwapAmountInput({
...
   );
 
+  const handleAnalyticsTokenSelected = useCallback(
+    (token: Token) => {
+      sendAnalytics(SwapEvent.TokenSelected, {
+        token: token.symbol,
+      });
+    },
+    [sendAnalytics],
+  );
```
alessey, line 64: no need to wrap this

alessey:
I'm seeing 3 init's locally, when I expect 2 because i'm in dev mode.  I think you need to remove state from your analytics hook, you shouldn't need to store document.title

cpcramer:
> I'm seeing 3 init's locally, when I expect 2 because i'm in dev mode. I think you need to remove state from your analytics hook, you shouldn't need to store document.title

I'm not seeing 3 init's anymore when testing in playground

src/swap/components/SwapAmountInput.tsx
dschlabach: Seems like a bigger issue with `sendAnalytics` or the way we're doing callbacks

dschlabach (approved):

brendan-defi (approved):

PR #1965, opened 2025-02-11
feat: Add Transaction Telemetry 

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**
Tested in the Playground

### DIFF
--- a/src/core/analytics/types.ts
+++ b/src/core/analytics/types.ts
@@ -279,27 +279,15 @@ export type MintEventData = {
 export type TransactionEventData = {
   [TransactionEvent.TransactionFailure]: CommonAnalyticsData & {
     error: string;
-    contracts: Array<{
-      contractAddress: string;
-      function: string;
-    }>;
     metadata?: Record<string, unknown>;
   };
   [TransactionEvent.TransactionInitiated]: CommonAnalyticsData & {
-    address: string;
-    contracts: Array<{
-      contractAddress: string;
-      function: string;
-    }>;
+    address?: string;
   };
   [TransactionEvent.TransactionSuccess]: CommonAnalyticsData & {
-    paymaster: boolean;
-    address: string;
-    contracts: Array<{
-      contractAddress: string;
-      function: string;
-    }>;
-    transactionHash: string;
+    paymaster?: boolean;
+    address?: string;
+    transactionHash?: string;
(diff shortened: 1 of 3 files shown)


brendan-defi (approved):

src/core/analytics/types.ts
```diff
@@ -279,27 +279,15 @@ export type MintEventData = {
...
-    paymaster: boolean;
-    address: string;
-    contracts: Array<{
-      contractAddress: string;
-      function: string;
-    }>;
-    transactionHash: string;
+    paymaster?: boolean;
+    address?: string;
+    transactionHash?: string;
```
brendan-defi, line 290: same comment as the other PR — seems strange that these are optional

src/transaction/components/TransactionProvider.tsx
```diff
@@ -161,6 +166,15 @@ export function TransactionProvider({
...
   });
 
+  const { sendAnalytics } = useAnalytics();
+
+  const handleAnalytics = useCallback(
+    (event: TransactionEvent, data: TransactionEventData[TransactionEvent]) => {
+      sendAnalytics(event, data);
+    },
+    [sendAnalytics],
+  );
```
brendan-defi, line 176: this seems like a nice pattern, any reason you are only using here and not in the other PRs?

cpcramer: Some components do use this pattern, it's dependent on the component. Likely the other PRs that you reviewed I thought it would be unnecessary, maybe the analytic calls were straight forward and didn't need a handler function?

src/core/analytics/types.ts
cpcramer: My thinking here: It just helps with type errors and flexibility with how we are handling analytics. Sometimes things break and I don't ever want analytics to block. While i'd prefer all parameters to be passed, I'd still rather have some information about the event then no event at all.

dschlabach: Left a similar comment on #1969 - i'd try and make them value optional if possible

cpcramer: Updated

brendan-defi (approved):

dschlabach (approved):

PR #1968, opened 2025-02-11
feat: Add Checkout Telemetry

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


src/checkout/components/CheckoutProvider.tsx
```diff
@@ -202,9 +205,46 @@ export function CheckoutProvider({
...
+  const handleAnalyticsFailure = useCallback(
+    (error: string, metadata?: Record<string, unknown>) => {
+      sendAnalytics(CheckoutEvent.CheckoutFailure, {
+        error,
+        metadata,
+      });
+    },
+    [sendAnalytics],
+  );
+
```
0xAlec, line 242: I think we can consolidate these to a single function that takes in the analytics event as a param

cpcramer: good call

dschlabach (approved):

alessey (approved):

PR #1969, opened 2025-02-11
feat: Add Buy Telemetry

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/buy/components/BuyDropdown.tsx
+++ b/src/buy/components/BuyDropdown.tsx
@@ -1,3 +1,5 @@
+import { useAnalytics } from '@/core/analytics/hooks/useAnalytics';
+import { BuyEvent, type BuyOption } from '@/core/analytics/types';
 import { openPopup } from '@/internal/utils/openPopup';
 import { useOnchainKit } from '@/useOnchainKit';
 import { useCallback, useEffect, useMemo } from 'react';
@@ -17,10 +19,15 @@ export function BuyDropdown() {
   const { to, fromETH, fromUSDC, from, startPopupMonitor, setIsDropdownOpen } =
     useBuyContext();
   const { address } = useAccount();
+  const { sendAnalytics } = useAnalytics();
 
   const handleOnrampClick = useCallback(
     (paymentMethodId: string) => {
       return () => {
+        sendAnalytics(BuyEvent.BuyOptionSelected, {
+          option: paymentMethodId as BuyOption,
+        });
+
         const assetSymbol = to?.token?.symbol;
         let fundAmount = to?.amount;
         // funding url requires a leading zero if the amount is less than 1
@@ -43,7 +50,7 @@ export function BuyDropdown() {
         }
       };
     },
-    [address, to, projectId, startPopupMonitor],
+    [address, to, projectId, startPopupMonitor, sendAnalytics],
   );
 
   const formattedAmountUSD = useMemo(() => {
--- a/src/buy/components/BuyProvider.tsx
+++ b/src/buy/components/BuyProvider.tsx
@@ -12,6 +12,8 @@ import { useAccount, useConfig, useSendTransaction } from 'wagmi';
 import { useSwitchChain } from 'wagmi';
 import { useSendCalls } from 'wagmi/experimental';
 import { buildSwapTransaction } from '../../api/buildSwapTransaction';
+import { useAnalytics } from '../../core/analytics/hooks/useAnalytics';
+import { BuyEvent } from '../../core/analytics/types';
 import { useCapabilitiesSafe } from '../../internal/hooks/useCapabilitiesSafe';
 import { useValue } from '../../internal/hooks/useValue';
 import { FALLBACK_DEFAULT_MAX_SLIPPAGE } from '../../swap/constants';
@@ -108,6 +110,9 @@ export function BuyProvider({
   // used to detect when the popup is closed in order to stop loading state
   const { startPopupMonitor } = usePopupMonitor(onPopupClose);
 
+  // Analytics
+  const { sendAnalytics } = useAnalytics();
+
   // Component lifecycle emitters
   useEffect(() => {
     // Error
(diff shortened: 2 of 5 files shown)


src/buy/components/BuyProvider.tsx
```diff
@@ -108,6 +111,16 @@ export function BuyProvider({
...
 
+  // Analytics
+  const { sendAnalytics } = useAnalytics();
+
+  const handleAnalytics = useCallback(
+    <T extends BuyEvent>(event: T, data: AnalyticsEventData[T]) => {
+      sendAnalytics(event, data);
+    },
+    [sendAnalytics],
+  );
```
alessey, line 122: why do you need to wrap sendAnalytics?

src/buy/components/BuyProvider.tsx
```diff
@@ -195,10 +222,15 @@ export function BuyProvider({
...
-      // biome-ignore lint/complexity/noExcessiveCognitiveComplexity: TODO Refactor this component
-    ) => {
+    // biome-ignore lint/complexity/noExcessiveCognitiveComplexity: TODO Refactor this component
+    async (amount: string) => {
+      if (amount !== '' && amount !== '.' && Number.parseFloat(amount) !== 0) {
+        handleAnalytics(BuyEvent.BuyInitiated, {
+          amount: Number(amount),
+          token: to?.token?.symbol || '',
+        });
+      }
```
alessey, line 232: could you move this to where a buy is actually initiated?

src/buy/components/BuyProvider.tsx
cpcramer: We don't need to, but i've been memoizing the sendAnalytics calls

src/buy/components/BuyProvider.tsx
cpcramer: whoops - Updating

src/buy/components/BuyProvider.tsx
cpcramer: I can remove if we'd like!

cpcramer: got it, updated

src/buy/components/BuyDropdown.tsx
```diff
@@ -17,10 +19,15 @@ export function BuyDropdown() {
...
   const { address } = useAccount();
+  const { sendAnalytics } = useAnalytics();
 
   const handleOnrampClick = useCallback(
     (paymentMethodId: string) => {
       return () => {
+        sendAnalytics(BuyEvent.BuyOptionSelected, {
+          option: paymentMethodId as BuyOption,
+        });
+
```
brendan-defi, line 30: it seems like in this case you aren't preventing multiple events from being sent. is that intentional? will this send the same event if the user keeps opening the dropdown and selecting the same payment method? is that desired?

src/buy/components/BuyProvider.tsx
```diff
@@ -117,10 +122,19 @@ export function BuyProvider({
...
       setHasHandledSuccess(true);
+
+      sendAnalytics(BuyEvent.BuySuccess, {
+        address,
+        amount: Number(from?.amount || 0),
+        from: from?.token?.address || '',
+        paymaster: !!paymaster,
+        to: to?.token?.address || '',
+        transactionHash: txHash || '',
+      });
```
brendan-defi, line 137: these are probably just typescript issues, but it's a little weird to see all the `||` handling in the success case. 

would it make sense to handle these separately? like, if we don't have a from.token.address, wouldn't we not expect a success? and if it really were the case that `lifecycleStatus.statusName === 'success'` and `from.token.address` was nullish, wouldn't that signal some sort of error?

src/core/analytics/types.ts
```diff
@@ -210,19 +210,19 @@ export type BuyEventData = {
...
-    from: string;
-    paymaster: boolean;
-    to: string;
-    transactionHash: string;
+    address?: string;
+    amount?: number;
+    from?: string;
+    paymaster?: boolean;
+    to?: string;
+    transactionHash?: string;
```
brendan-defi, line 225: related to comment above, optional seems weird on these. if BuyOptionSelected, i'd expect it to be required that there's a BuyOption, no?

cpcramer: My thinking here: It just helps with type errors and flexibility with how we are handling analytics. Sometimes things break and I don't ever want analytics to block. While i'd prefer all parameters to be passed, I'd still rather have some information about the event then no event at all.

src/buy/components/BuyProvider.tsx
cpcramer: Yes, for type errors. 

I don't want sendAnalytics to ever block the component flow / logic! 

For example, if from.token.address was null then we should handle that error outside of sendAnalytics. More appropriate to update that error handling outside of the analytics implementation

src/buy/components/BuyDropdown.tsx
cpcramer: Yes that's the intention. We want to understand the user's journey, if they are opening the dropdown multiple times before using the component then we want to see that!

src/core/analytics/types.ts
dschlabach: I think you should make these _value optional_ and not _key_ optional. That'll ensure that you know exactly what you should pass in the ideal state.

e.g.

```
address: string | undefined
amount number | undefined
...
```

cpcramer: That's a great idea. I should actually refactor all event types to follow this pattern.

src/buy/components/BuyProvider.tsx
brendan-defi: makes sense

brendan-defi (dismissed):

src/buy/components/BuyProvider.tsx
```diff
@@ -117,10 +122,19 @@ export function BuyProvider({
...
+      setTransactionHash(txHash);
       setHasHandledSuccess(true);
+
+      sendAnalytics(BuyEvent.BuySuccess, {
+        address,
+        amount: Number(from?.amount || 0),
+        from: from?.token?.address || '',
+        paymaster: !!paymaster,
+        to: to?.token?.address || '',
+        transactionHash: txHash || '',
```
dschlabach, line 136: Do you need the fallbacks now?

src/buy/components/BuyProvider.tsx
```diff
@@ -415,6 +451,7 @@ export function BuyProvider({
       updateLifecycleStatus,
       useAggregator,
       walletCapabilities,
+      sendAnalytics,
```
dschlabach, line 454: This is a little weird to expose out of `BuyProvider`, the fact that we're returning it from multiple contexts suggests there are some design flaws with `useAnalytics`

src/buy/components/BuyProvider.tsx
cpcramer: Ah we don't, updating

src/buy/components/BuyProvider.tsx
cpcramer: Update - We are not exposing sendAnalytics out of the BuyProvider here. In general, analytics could use a refactor to remain consistent across all components, reduce unnecessary memoization, minimize amount of code needed to support, etc .

dschlabach: yep my bad, read this wrong! sorry

dschlabach (approved):

brendan-defi (approved):

PR #1998, opened 2025-02-18
feat: add `isSponsored` to `<Earn />`

dschlabach:
**What changed? Why?**
This PR updates `<Earn />` to accept an `isSponsored` prop which is then used to sponsor deposit and withdraw transactions

**Notes to reviewers**

**How has it been tested?**
Locally in playground

### DIFF
--- a/src/earn/components/EarnProvider.tsx
+++ b/src/earn/components/EarnProvider.tsx
@@ -21,7 +21,11 @@ import type {
 
 const EarnContext = createContext<EarnContextType | undefined>(undefined);
 
-export function EarnProvider({ vaultAddress, children }: EarnProviderReact) {
+export function EarnProvider({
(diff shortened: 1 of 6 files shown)


src/earn/components/EarnProvider.tsx
```diff
@@ -21,7 +21,11 @@ import type {
 
 const EarnContext = createContext<EarnContextType | undefined>(undefined);
 
-export function EarnProvider({ vaultAddress, children }: EarnProviderReact) {
+export function EarnProvider({
+  vaultAddress,
+  children,
+  isSponsored,
```
cpcramer, line 27: Do we want `isSponsored` to default to `false` somewhere? Perhaps in the provider?

cpcramer (approved):

PR #2027, opened 2025-02-21
fix: NFTMintButton analytics

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**
Tested in playground. 

Screenshot 2025-02-21 at 10 54 57 AM

Screenshot 2025-02-21 at 10 55 09 AM

### DIFF
--- a/src/nft/hooks/useMintAnalytics.ts
+++ b/src/nft/hooks/useMintAnalytics.ts
@@ -1,0 +1,61 @@
+import { useAnalytics } from '@/core/analytics/hooks/useAnalytics';
+import { MintEvent } from '@/core/analytics/types';
+import { useNFTContext } from '@/nft/components/NFTProvider';
+import type { LifecycleStatus } from '@/transaction/types';
+import { useEffect, useMemo, useRef, useState } from 'react';
+
+export const useMintAnalytics = () => {
+  const [transactionState, setTransactionState] = useState<
+    LifecycleStatus['statusName'] | null
+  >(null);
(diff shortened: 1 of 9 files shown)


dschlabach (approved):

alessey (approved):
looks good, but you should try fixing up useAnalytics with a callback and/or removing state to see if that eliminates the need for some of these wrappers when useAnalytics isn't generating a new callback on every render

src/nft/hooks/useMintAnalytics.ts
```diff
@@ -0,0 +1,71 @@
...
+      });
+    }
+
+    if (transactionState === 'error' && !errorSent.current) {
+      errorSent.current = true;
+      sendAnalytics(MintEvent.MintFailure, {
+        error: 'Transaction failed',
+        metadata: analyticsData,
+      });
+    }
```
brendan-defi, line 58: is it possible that a user would encounter multiple different errors? if not, then this seems like it would work fine. but if so, wouldn't this prevent error logging after the first one?

cpcramer: In `useMintAnalytics` I didn't have access to a specific error msg - which is why you see the general error. If I did have a specific error I'd pass it in the `sendAnalytics` call!

When implementing analytics I wanted to avoid changing the surrounding code - it would be more appropriate to update error handling in a separate PR if a user could encounter multiple errors!

dschlabach (approved):

PR #2029, opened 2025-02-21
Onramp support any network

rustam-cb:
**What changed? Why?**
- Added `walletAddress` and `walletNetwork` props to `FundCard and `FundButton` components to enable these components for any network.

**Notes to reviewers**

**How has it been tested?**


src/fund/components/FundCardProvider.tsx
```diff
@@ -25,6 +25,8 @@ import type {
 type FundCardContextType = {
   asset: string;
   currency: string;
+  walletAddress?: string;
+  walletNetwork?: string;
```
alessey, line 29: shouldn't you be able to get the address and network from wagmi if you set this up with the solana wagmi connector?  or does wagmi not work properly?

dgca:
Closing this out since it's out of date. LMK if you wanna work together to get this into the v1 release.

PR #2043, opened 2025-02-25
fix: improve useOutsideClick portal handling

brendan-defi:
**What changed? Why?**
* added `data-portal-origin="true"` property to all portal-using components
* added logic to `useOutsideClick` to exclude clicks from elements with data-porta-origin

why is excluding portal-clicks on `useOutsideClick` okay?
* if there is a portal open then...
  * either it opened from an outside click
  * or it opened from an "inside click"
* in the former case, `useOutsideClick` will call the callback
* in the latter case, given that the portal opened _because of an "inside click"_, clicks on the portal should not constitute outside clicks and should not call the callback

**Notes to reviewers**

**How has it been tested?**
in playground

### DIFF
--- a/.changeset/yellow-toys-sin.md
+++ b/.changeset/yellow-toys-sin.md
@@ -1,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
+
+- **fix** update useOutsideClick to handle portals. fixes issue with WalletAdvancedSwap. by @brendan-defi #2043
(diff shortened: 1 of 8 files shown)


cpcramer (approved):

cpcramer:
Add a changeset please!

cpcramer (approved):

brendan-defi:
@cpcramer added a changeset, could you take another look and re-approve when you've got a chance?

cpcramer (approved):

alessey (approved):

PR #2053, opened 2025-02-27
fix: Checkout component success event

cpcramer:
**What changed? Why?**
We were not emitting Checkout Success events as expected. This update fixes that by moving the analytics call to the CheckoutButton. 

Screenshot 2025-03-03 at 4 06 06 PM

Screenshot 2025-03-03 at 4 05 27 PM

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/checkout/components/CheckoutButton.test.tsx
+++ b/src/checkout/components/CheckoutButton.test.tsx
@@ -1,5 +1,16 @@
-import { fireEvent, render, screen } from '@testing-library/react';
-import { type Mock, beforeEach, describe, expect, it, vi } from 'vitest';
+import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import {
+  type Mock,
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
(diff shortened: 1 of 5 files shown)


src/checkout/components/CheckoutButton.test.tsx
```diff
@@ -8,88 +19,158 @@ vi.mock('./CheckoutProvider', () => ({
...
+vi.mock('../../core/analytics/hooks/useAnalytics', () => ({
+  useAnalytics: vi.fn(),
 }));
 
-const useCheckoutContextMock = useCheckoutContext as Mock;
-
 describe('CheckoutButton', () => {
-  const mockOnSubmit = vi.fn();
+  let mockOnSubmit: Mock;
+  let mockSendAnalytics: Mock;
```
brendan-defi, line 32: IIUC you never reassign these, so would be better to make them consts

brendan-defi (approved):

abcrane123 (approved):

PR #2057, opened 2025-02-28
fix(wallet): Export WalletModal to allow custom opening

RezaRahemtola:
### **What changed? Why?**
Added an export of the `WalletModal` component to allow opening it from custom connect buttons instead of the generic UI.
Spotted during ETH Denver, it was useful for our case as we have multiple login providers, including OnchainKit, and don't want to use the big "Connect" button, but instead our own button that will open the modal.

Using the wagmi connector directly wasn't possible as it only exposes Coinbase Wallet, not the Metamask / Phantom options also provided in this modal

### **Notes to reviewers**
/

### **How has it been tested?**
Straightforward change that shouldn't require testing, but we tested and integrated in our project

### DIFF
--- a/src/wallet/index.ts
+++ b/src/wallet/index.ts
@@ -18,7 +18,7 @@ export { WalletAdvancedTokenHoldings } from './components/WalletAdvancedTokenHol
 export { WalletAdvancedQrReceive } from './components/WalletAdvancedQrReceive';
 export { WalletAdvancedSwap } from './components/WalletAdvancedSwap';
 export { WalletAdvancedWalletActions } from './components/WalletAdvancedWalletActions';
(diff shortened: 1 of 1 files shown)


gmolki (approved):

cpcramer (approved):

cpcramer:
Good idea!

abcrane123:
can you try running `yarn format` from root to fix formatting and then we can merge!

RezaRahemtola:
> can you try running `yarn format` from root to fix formatting and then we can merge!

Done 👍

cpcramer (approved):

abcrane123 (approved):

PR #2059, opened 2025-03-03
fix: fixed strict check for paymasterUrl in useCapabilities

hexcow:
### **Title:**  
Fixed strict check for `paymasterUrl` in `useCapabilities`

### **Description:**  

**What changed? Why?**  
Previously, `paymasterService` was added even when `paymasterUrl` was an empty string (`""`), resulting in an invalid configuration. Now, the check ensures that `paymasterService` is only included if `paymasterUrl` is not empty, preventing incorrect payloads.  

**Notes to reviewers**  
The main change is in the conditional logic—switched from a loose check to an explicit validation. This prevents edge cases where an empty string could lead to an incomplete configuration.  

**How has it been tested?**  
- Added cases with empty and non-empty `paymasterUrl` values.  
- Verified that `paymasterService` is only included when the URL is present.  
- Checked integration with `usePaymaster` to ensure no regressions.

### DIFF
--- a/playground/nextjs-app-router/lib/hooks.ts
+++ b/playground/nextjs-app-router/lib/hooks.ts
@@ -18,10 +18,8 @@ export function useCapabilities() {
 
   return enabled
     ? {
-        ...(paymasterUrl && {
-          paymasterService: { url: paymasterUrl },
-        }),
-      }
+        ...(paymasterUrl ? { paymasterService: { url: paymasterUrl } } : {}),
+       }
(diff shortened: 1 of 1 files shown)


dgca:
Hey @hexcow I'm gonna close this one since it's a PR against the playground which doesn't affect the library.

We're getting close to finalizing the v1 release so the playground is gonna have some significant changes pretty soon. Appreciate the help, though.

PR #2062, opened 2025-03-03
chore: Refactor WalletModal init event

cpcramer:
**What changed? Why?**
We weren't logging the Wallet Provider in our connect wallet events as expected. Rather than having connectWallet and WalletModal share a single handleConnectInitiated call, this creates a new handler inside WalletModal making it easier to pass the provider used. 

- Removed the walletProvider in the ConnectSuccess call. The provider is already being tracked in the initiation event, and can be tied to a successful connecting using the sessionID

**Notes to reviewers**

**How has it been tested?**
WalletModal with Phantom connector: 
Screenshot 2025-03-03 at 12 07 00 PM

Screenshot 2025-03-03 at 12 07 11 PM

ConnectWallet: 
Screenshot 2025-03-03 at 12 11 24 PM

Screenshot 2025-03-03 at 12 10 59 PM

### DIFF
--- a/src/wallet/components/ConnectWallet.tsx
+++ b/src/wallet/components/ConnectWallet.tsx
@@ -92,9 +92,9 @@ export function ConnectWallet({
   }, [setIsConnectModalOpen]);
 
   const handleAnalyticsInitiated = useCallback(
-    (connectorName: string, component = 'ConnectWallet') => {
+    (connectorName: string) => {
       sendAnalytics(WalletEvent.ConnectInitiated, {
-        component,
+        component: 'ConnectWallet',
         walletProvider: connectorName,
       });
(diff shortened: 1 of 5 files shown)


src/wallet/components/ConnectWallet.tsx
```diff
@@ -134,7 +130,7 @@ export function ConnectWallet({
 
   useEffect(() => {
     if (status === 'connected' && accountAddress && connector) {
-      handleAnalyticsSuccess(connector.name, accountAddress);
+      handleAnalyticsSuccess(accountAddress);
```
alessey, line 133: can't you get the actual connector an account is connected with from useAccount()?

PR #2064, opened 2025-03-03
chore: Refactor WalletProvider event to use connector from useAccount

cpcramer:
**What changed? Why?**
Refactor our wallet connector events to use the connector from `useAccount()` to determine and pass the WalletProvider in our analytics events. 
Screenshot 2025-03-03 at 3 23 31 PM

**Notes to reviewers**

**How has it been tested?**
Screenshot 2025-03-03 at 1 58 09 PM

Screenshot 2025-03-03 at 1 58 22 PM

Screenshot 2025-03-03 at 1 58 35 PM

### DIFF
--- a/src/core/analytics/types.ts
+++ b/src/core/analytics/types.ts
@@ -163,8 +163,6 @@ export type WalletEventData = {
   [WalletEvent.ConnectInitiated]: CommonAnalyticsData & {
     /** Component used to connect wallet */
     component: string;
-    /** Coinbase, Metamask, Phantom, etc. */
-    walletProvider: string;
   };
   [WalletEvent.ConnectError]: CommonAnalyticsData & {
     error: string;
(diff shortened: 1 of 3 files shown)


src/core/analytics/types.ts
```diff
@@ -163,15 +163,14 @@ export type WalletEventData = {
   [WalletEvent.ConnectInitiated]: CommonAnalyticsData & {
     /** Component used to connect wallet */
     component: string;
-    /** Coinbase, Metamask, Phantom, etc. */
-    walletProvider: string;
```
alessey, line 167: would this be useful in determining connect issues?  i.e. initiated counts on a provider < success?

alessey (approved):

src/core/analytics/types.ts
cpcramer: I think that it could be nice to have but we do not have access to the walletProvider at the time the wallet connection is being initiated

PR #2065, opened 2025-03-03
Feat/getPriceQuote

brendan-defi:
**What changed? Why?**
* added API endpoint `getPriceQuote`
  * `getPriceQuote` fetches price quotes for the provided token contract addresses and/or ETH
  * `getPriceQuote` is available only on Base mainnet

*`getPriceQuote` interface*

Request
```
type GetPriceQuoteParams = {
  tokens: PriceQuoteToken[];
};
```
Response
```
type GetPriceQuoteResponse =
  | {
      priceQuote: PriceQuote[];
    }
  | APIError;

type PriceQuote = {
  name: string | '';
  symbol: string | '';
  contractAddress: Address | '';
  price: string | '';
  timestamp: number | 0;
};
```
Usage
```
const response = getPriceQuote({ tokens: ["ETH", "0x123...ABC"] });
if (response.error) {
  throw New Error(priceQuotes.error)
}

const tokenPrice = response.priceQuotes[0].price;
```

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/api/getPriceQuote.ts
+++ b/src/api/getPriceQuote.ts
@@ -1,0 +1,59 @@
+import { RequestContext } from '@/core/network/constants';
+import { CDP_GET_PRICE_QUOTE } from '@/core/network/definitions/wallet';
+import { sendRequest } from '@/core/network/request';
+
+import type { GetPriceQuoteParams, GetPriceQuoteResponse } from './types';
+
+/**
+ * Retrieves a price quote for a token
+ *
+ * @param params - The parameters for the price quote. The property `tokens`
+ * must be an array of contract addresses or 'ETH'.
+ * @param _context - The context in which the price quote is retrieved
+ * @returns The price quote for the token
+ */
+export async function getPriceQuote(
+  params: GetPriceQuoteParams,
+  _context: RequestContext = RequestContext.API,
+): Promise<GetPriceQuoteResponse> {
+  const apiParams = validateGetPriceQuoteParams(params);
+  if ('error' in apiParams) {
+    return apiParams;
+  }
+
+  try {
+    const res = await sendRequest<GetPriceQuoteParams, GetPriceQuoteResponse>(
+      CDP_GET_PRICE_QUOTE,
+      [apiParams],
+      _context,
+    );
+    if (res.error) {
+      return {
+        code: `${res.error.code}`,
+        error: 'Error fetching price quote',
+        message: res.error.message,
+      };
+    }
+    return res.result;
+  } catch (error) {
+    return {
+      code: 'UNCAUGHT_PRICE_QUOTE_ERROR',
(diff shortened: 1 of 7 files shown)


src/api/getPriceQuote.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  try {
+    const res = await sendRequest<GetPriceQuoteParams, GetPriceQuoteResponse>(
+      CDP_GET_PRICE_QUOTE,
+      [apiParams],
+      _context,
+    );
+    if (res.error) {
+      return {
+        code: `${res.error.code}`,
```
dschlabach, line 32:
```suggestion
        code: res.error.code,
```

src/api/getPriceQuote.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+function validateGetPriceQuoteParams(params: GetPriceQuoteParams) {
+  const { tokens } = params;
+
+  if (!tokens || tokens.length === 0) {
+    return {
+      code: 'INVALID_INPUT',
+      error: 'Invalid input: tokens must be an array of at least one token',
+      message: '',
```
dschlabach, line 54: May be useful to have `tokens must be an array of at least one token` in the `message` ?

src/api/types.ts
```diff
@@ -384,3 +384,36 @@ export type BuildSendTransactionParams = {
...
+  name: string | '';
+  /** The symbol of the token */
+  symbol: string | '';
+  /** The contract address of the token */
+  contractAddress: Address | '';
+  /** The price of the token */
+  price: string | '';
+  /** The timestamp of the price quote */
+  timestamp: number | 0;
+};
```
dschlabach, line 409: Out of curiosity why the `''`?

dschlabach (approved):

abcrane123 (approved):
lgtm

src/api/getPriceQuote.ts
brendan-defi: I'll change to String() instead of template-stringing to be more explicit about intention here.

src/api/getPriceQuote.ts
brendan-defi: good call, adding

src/api/types.ts
brendan-defi: I haven't been able to discern the pattern, but sometimes the API doesn't return all the fields. 

I don't think this is a problem, because it has so far _always_ returned the contractAddress, and the caller should have all the other info.

alessey: wouldn't '' be covered under string?

brendan-defi: @alessey ah, you're right, i can remove those nullish versions of the types. will handle in a follow-up

PR #2066, opened 2025-03-03
refactor: improve WalletAdvanced sub-component handling

brendan-defi:
**What changed? Why?**
* previously, we had state for the two sub-components: QR and Swap. 
* in preparation for the addition of Send, I refactored how WalletAdvanced handles sub-components
* instead of separate state for each sub-component, there is one state: `activeFeature`. this simplifies the devex for controlling which feature is displayed

**Notes to reviewers**

**How has it been tested?**
test coverage and in playground

### DIFF
--- a/src/wallet/components/WalletAdvancedContent.tsx
+++ b/src/wallet/components/WalletAdvancedContent.tsx
@@ -23,7 +23,7 @@ export function WalletAdvancedContent({
     breakpoint,
   } = useWalletContext();
 
-  const { showQr, showSwap, tokenBalances, animations } =
+  const { activeFeature, tokenBalances, animations } =
     useWalletAdvancedContext();
 
   const handleBottomSheetClose = useCallback(() => {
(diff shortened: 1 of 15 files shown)


src/wallet/components/WalletAdvancedContent.tsx
```diff
@@ -38,15 +38,15 @@ export function WalletAdvancedContent({
   }, [isSubComponentClosing, setIsSubComponentOpen, setIsSubComponentClosing]);
 
   const content = useMemo(() => {
-    if (showQr) {
+    if (activeFeature === 'qr') {
```
alessey, line 41: should activeFeature correspond to the button position?  i.e. do consumers have the ability to use different buttons here and open different panes for those buttons using activeFeature?

brendan-defi: > do consumers have the ability to use different buttons here and open different panes for those buttons using activeFeature?

not yet. we still have not taken on the work to allow bring-your-own sub-components, so right now activeFeature is limited to those that we provide

dschlabach (approved):

alessey (approved):

PR #2068, opened 2025-03-04
chore: better TokenBalance 

brendan-defi:
**What changed? Why?**
* Refactored the "Action Button" (typically "max")
* Refactored types to be discriminated unions that disallow erroneous prop combinations

**Notes to reviewers**

**How has it been tested?**
* locally, playground, test coverage

### DIFF
--- a/src/token/components/TokenBalance.tsx
+++ b/src/token/components/TokenBalance.tsx
@@ -10,7 +10,7 @@ export function TokenBalance({
   token,
   onClick,
   onActionPress,
-  actionText = 'Use max',
+  actionText = 'Max',
   classNames,
   'aria-label': ariaLabel,
   ...contentProps
--- a/src/token/components/TokenBalance.tsx
+++ b/src/token/components/TokenBalance.tsx
@@ -153,11 +153,6 @@ function TokenBalanceActionButton({
       data-testid="ockTokenBalanceAction"
       aria-label={actionText}
       onClick={onActionPress}
-      onKeyDown={(e) => {
-        if (e.key === 'Enter' || e.key === ' ') {
-          onActionPress?.();
-        }
-      }}
       className={cn(
         text.label2,
         color.primary,
--- a/src/token/components/TokenBalance.tsx
+++ b/src/token/components/TokenBalance.tsx
@@ -2,33 +2,47 @@ import { formatFiatAmount } from '@/internal/utils/formatFiatAmount';
 import { truncateDecimalPlaces } from '@/internal/utils/truncateDecimalPlaces';
 import { border, cn, color, text } from '@/styles/theme';
 import { TokenImage } from '@/token/components/TokenImage';
-import { useCallback } from 'react';
+import { useMemo } from 'react';
 import { formatUnits } from 'viem';
 import type { TokenBalanceProps } from '../types';
 
 export function TokenBalance({
   token,
   onClick,
+  onActionPress,
+  actionText = 'Use max',
   classNames,
+  'aria-label': ariaLabel,
   ...contentProps
 }: TokenBalanceProps) {
   if (onClick) {
     return (
-      <button
-        type="button"
-        onClick={() => onClick(token)}
-        className={cn(
-          'flex w-full items-center justify-start gap-4 px-2 py-1',
-          classNames?.container,
+      <div className="relative">
+        <button
+          type="button"
+          aria-label={ariaLabel ?? `${token.name} token balance`}
+          onClick={() => onClick(token)}
+          className={cn(
+            'flex w-full items-center justify-start gap-4 px-2 py-1',
+            classNames?.container,
+          )}
+          data-testid="ockTokenBalanceButton"
+        >
+          <TokenBalanceContent
+            token={token}
+            classNames={classNames}
+            onActionPress={onActionPress}
+            {...contentProps}
+          />
+        </button>
+        {onActionPress && (
+          <TokenBalanceActionButton
+            actionText={actionText}
+            onActionPress={onActionPress}
+            className={classNames?.action}
+          />
         )}
-        data-testid="ockTokenBalanceButton"
-      >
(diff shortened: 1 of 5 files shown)


src/token/components/TokenBalance.tsx
```diff
@@ -141,3 +139,31 @@ function TokenBalanceContent({
     </div>
   );
 }
+
+function TokenBalanceActionButton({
```
dschlabach, line 143: Because this isn't used anywhere outside of TokenBalance I think it'd be better to call this `ActionButton` or even `Action` - `TokenBalance` isn't really adding much

brendan-defi: makes sense, i've renamed

src/token/components/TokenBalance.tsx
```diff
@@ -2,33 +2,47 @@ import { formatFiatAmount } from '@/internal/utils/formatFiatAmount';
...
-import { useCallback } from 'react';
+import { useMemo } from 'react';
 import { formatUnits } from 'viem';
 import type { TokenBalanceProps } from '../types';
 
 export function TokenBalance({
   token,
   onClick,
+  onActionPress,
+  actionText = 'Use max',
```
alessey, line 13: swap is just "max", should this match?

dschlabach (approved):

cpcramer:
nit: title category should fall under chore, fix, feat, or docs. 

`chore: Update TokenBalance`

src/token/types.ts
```diff
@@ -146,13 +144,29 @@ export type TokenBalanceProps = {
...
+      showImage?: true;
+      /** Size of the token image in px (default: 40) */
+      tokenSize?: number;
     }
   | {
-      /** Show an additional action button (eg. "Use max") */
-      actionText?: string;
-      onActionPress: () => void;
+      /** Hide the token image */
+      showImage: false;
```
cpcramer, line 154: nit: 
```suggestion
      /** Hide the token image (default: false) */
      showImage: false;
```

cpcramer, line 154: nit: 
```suggestion
      /** Hide the token image (default: false) */
      showImage: false;
```

src/token/types.ts
```diff
@@ -146,13 +144,29 @@ export type TokenBalanceProps = {
...
+      /** Size of the token image in px (default: 40) */
+      tokenSize?: never;
     }
-);
+) &
+  (
+    | {
+        /** Hide the action button (default)*/
+        onActionPress?: never;
```
cpcramer, line 162: What's the default here?

src/token/components/TokenBalance.tsx
```diff
@@ -141,3 +139,31 @@ function TokenBalanceContent({
     </div>
   );
 }
+
+function ActionButton({
```
cpcramer, line 143: Thoughts on putting `ActionButton` and or `TokenBalanceContent` in a separate file?

cpcramer: Seems like `TokenBalanceContent` should be extracted

src/token/components/TokenBalance.tsx
brendan-defi: I feel like 'Use Max' makes more sense for this context:
image

I don't know where swap is using just "max", but maybe it would make more sense to change over there?

src/token/types.ts
brendan-defi: this type is a discriminated union:
```
(
  | {
      /** Show the token image (default: true) */
      showImage?: true;
      /** Size of the token image in px (default: 40) */
      tokenSize?: number;
    }
  | {
      /** Hide the token image */
      showImage: false;
      /** Size of the token image in px (default: 40) */
      tokenSize?: never;
    }
)
```

i put default true on the top `showImage` prop. 

because false is not the default, i think it'd be a little confusing to add a default to the false case... because it would need to say default true, but in the false case it's not true...

src/token/types.ts
brendan-defi: there's no default. removing the parenthetical

src/token/components/TokenBalance.tsx
cpcramer: I think we use `max` based off Uniswap 
Screenshot 2025-03-13 at 2 46 18 PM

cpcramer (approved):

src/token/components/TokenBalance.tsx
alessey: testability would be the main reason imo, but there's logic here to test

src/token/types.ts
```diff
@@ -146,13 +144,29 @@ export type TokenBalanceProps = {
   };
 } & (
   | {
-      /** Hide the action button (default)*/
-      actionText?: never;
-      onActionPress?: never;
+      /** Show the token image (default: true) */
+      showImage?: true;
```
alessey, line 148: boolean?

src/token/types.ts
```diff
@@ -146,13 +144,29 @@ export type TokenBalanceProps = {
...
+      /** Size of the token image in px (default: 40) */
+      tokenSize?: number;
     }
   | {
-      /** Show an additional action button (eg. "Use max") */
-      actionText?: string;
-      onActionPress: () => void;
+      /** Hide the token image (default: true) */
+      showImage: false;
```
alessey, line 154: boolean, or does this have to be false?  and if so, does it really default true?

alessey (approved):

cpcramer (dismissed):

src/token/types.ts
brendan-defi: updated the comment to clarify the discriminated union of these two types

src/token/types.ts
brendan-defi: updated the comment to clarify the discriminated union of these two types

alessey (approved):

PR #2078, opened 2025-03-04
feat: SendProvider

brendan-defi:
**What changed? Why?**
* implemented SendProvider
* implemented usePriceQuote hook for `getPriceQuote`
* improved useExchangeRate by using `usePriceHook`
* refactored prerequisite components to use context instead of props

**Notes to reviewers**

**How has it been tested?**
* test coverage

### DIFF
--- a/src/internal/hooks/useExchangeRate.tsx
+++ b/src/internal/hooks/useExchangeRate.tsx
@@ -31,10 +31,10 @@ export function useExchangeRate(
     const priceQuote = data.priceQuotes[0];
 
     if (selectedInputType === 'crypto') {
-      return Number(priceQuote.price);
+      return 1 / Number(priceQuote.price);
     }
 
-    return 1 / Number(priceQuote.price);
+    return Number(priceQuote.price);
   }, [data, selectedInputType]);
 
   return {
--- a/src/internal/hooks/useExchangeRate.test.tsx
+++ b/src/internal/hooks/useExchangeRate.test.tsx
@@ -58,7 +58,7 @@ describe('useExchangeRate', () => {
     );
 
     await waitFor(() => {
-      expect(result.current.exchangeRate).toEqual(2400);
+      expect(result.current.exchangeRate).toEqual(1 / 2400);
     });
   });
 
@@ -85,7 +85,7 @@ describe('useExchangeRate', () => {
     );
 
     await waitFor(() => {
-      expect(result.current.exchangeRate).toEqual(1 / 2400);
+      expect(result.current.exchangeRate).toEqual(2400);
     });
   });
 


src/wallet/components/wallet-advanced-send/components/SendAmountInputTypeSwitch.tsx
```diff
@@ -1,38 +1,33 @@
...
+    exchangeRate,
+    exchangeRateLoading,
+    selectedInputType,
+    setSelectedInputType,
+  } = useSendContext();
+
   // AmountInputTypeSwitch uses a skeleton for both loading and error states
-  // SendAmountInputTypeSwitch uses skeleton for the loading display
-  // SendAmountInputTypeSwitch uses a custom error display (see loadingDisplay default)
+  // SendAmountInputTypeSwitch uses skeleton for the loading display but a custom error display (see loadingDisplay default)
```
alessey, line 30: the error state is called loadingDisplay?

src/wallet/components/wallet-advanced-send/components/SendProvider.tsx
```diff
@@ -0,0 +1,293 @@
...
+      });
+    } else {
+      updateLifecycleStatus({
+        statusName: 'fundingWallet',
+        statusData: {
+          isMissingRequiredField: true,
+        },
+      });
+    }
+    setIsInitialized(true);
```
alessey, line 101: can this be derived?

src/wallet/components/wallet-advanced-send/components/SendProvider.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+  // fetch & set exchange rate
+  useEffect(() => {
+    if (!selectedToken) {
+      return;
+    }
+
+    const tokenSymbol = selectedToken.symbol;
+    const tokenAddress = selectedToken.address;
+    let tokenParam: PriceQuoteToken;
```
alessey, line 112: nit: const tokenParam = selectedToken.symbol === 'ETH' ? 'ETH' : selectedToken.address; if (tokenParam) useExchangeRate()?

src/wallet/components/wallet-advanced-send/types.ts
```diff
@@ -21,7 +26,7 @@ export type SendContextType = {
 
   // Sender Context
   /** The balance of the sender's ETH wallet */
-  ethBalance: number | undefined;
+  ethBalance: number;
```
alessey, line 29: should this have an undefined state to differentiate fetched and 0 from haven't fetched?

src/wallet/components/wallet-advanced-send/components/SendAmountInputTypeSwitch.tsx
dschlabach: I think we should let `exchangeRateLoading` be handled by `AmountInputTypeSwitch` and just return early if there's an error - any reason we can't do that @brendan-defi?

brendan-defi: @alessey Yes, right now AmountInputTypeSwitch shows "loadingDisplay" for both loading state and exchange-rate-not-present state:

image

brendan-defi: @dschlabach unfortunately that's not how AmountInputTypeSwitch currently works. 

AmountInputTypeSwitch doesn't handle loading state on its own. Instead, it accepts a prop `exchangeRateLoading` which it uses to display the loading / error state.

src/wallet/components/wallet-advanced-send/components/SendProvider.tsx
brendan-defi: yeah, good call, this is only based on eth balance right now. refactored in [0cc4028](https://github.com/coinbase/onchainkit/pull/2078/commits/0cc4028db6142a8a41272bf96951d16c7db349a6)

src/wallet/components/wallet-advanced-send/types.ts
brendan-defi: yes, now that we are using this as the derivation scheme, we need an undefined option. updated in [0cc4028](https://github.com/coinbase/onchainkit/pull/2078/commits/0cc4028db6142a8a41272bf96951d16c7db349a6)

src/wallet/components/wallet-advanced-send/components/SendProvider.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+  const handleCryptoAmountChange = useCallback(
+    (value: string) => {
+      const truncatedValue = truncateDecimalPlaces(value, 8);
+      setCryptoAmount(truncatedValue);
+      updateLifecycleStatus({
+        statusName: 'amountChange',
+        statusData: {
+          isMissingRequiredField: true,
+          sufficientBalance:
```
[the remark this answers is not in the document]
brendan-defi: hoisted to the top in a memo

src/wallet/components/wallet-advanced-send/components/SendAmountInputTypeSwitch.tsx
```diff
@@ -1,38 +1,33 @@
...
+  loadingDisplay?: React.ReactNode;
+};
 
 export function SendAmountInputTypeSwitch({
-  exchangeRateLoading,
   loadingDisplay = (
     <div className={cn(text.caption, color.foregroundMuted, 'h-[1.625rem]')}>
       Exchange rate unavailable
     </div>
   ),
```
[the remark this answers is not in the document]
brendan-defi: updated how we handle error state

src/internal/hooks/useExchangeRate.tsx
```diff
@@ -2,47 +2,62 @@ import { getPriceQuote } from '@/api';
...
-  } catch (error) {
-    console.error('Uncaught error fetching price quote:', error);
-  } finally {
-    setExchangeRateLoading?.(false);
-  }
+
+    setIsLoading(true);
+    setError(null);
+
+    getPriceQuote({ tokens: [token] }, RequestContext.Wallet)
```
alessey, line 31: you could wrap this in a react-query hook, i.e. https://github.com/coinbase/onchainkit/blob/main/src/nft/hooks/useMintDetails.ts and export that along with the API

src/wallet/components/wallet-advanced-send/components/SendProvider.tsx
```diff
@@ -0,0 +1,279 @@
...
+  if (sendContext === emptyContext) {
+    throw new Error('useSendContext must be used within a SendProvider');
+  }
+  return sendContext;
+}
+
+export function SendProvider({ children }: SendProviderReact) {
+  // state for ETH balance
+  const [ethBalance, setEthBalance] = useState<number | undefined>(undefined);
+  const isInitialized = useMemo(() => ethBalance !== undefined, [ethBalance]);
```
alessey, line 40: this memoization is doing a shallow compare to determine if it should do a shallow compare.  It'd be more performant not to memoize this.

src/internal/hooks/useExchangeRate.tsx
```diff
@@ -2,47 +2,62 @@ import { getPriceQuote } from '@/api';
 import type { PriceQuoteToken } from '@/api/types';
 import { RequestContext } from '@/core/network/constants';
 import { isApiError } from '@/internal/utils/isApiResponseError';
-import type { Dispatch, SetStateAction } from 'react';
+import { useEffect, useState } from 'react';
 
 type UseExchangeRateParams = {
-  token: PriceQuoteToken;
+  token: PriceQuoteToken | undefined | '';
```
alessey, line 8: token?: PriceQuoteToken | ''?

src/internal/hooks/useExchangeRate.tsx
cpcramer, line 31: Will this call always be in the `Wallet` context? If so should this hook be moved into `src/internal/wallet/hooks`?

src/wallet/components/wallet-advanced-send/components/SendAmountInputTypeSwitch.tsx
```diff
@@ -1,50 +1,55 @@
...
-  // SendAmountInputTypeSwitch uses a custom error display (see loadingDisplay default)
+  // SendAmountInputTypeSwitch uses skeleton for the loading display but a custom error display (see loadingDisplay default)
   if (exchangeRateLoading) {
     return <Skeleton className="h-[1.625rem]" />;
   }
 
+  if (!exchangeRate) {
+    if (errorDisplay) {
+      return errorDisplay;
+    }
```
cpcramer, line 34: We can check for this right after the above useSendContext() call

src/wallet/components/wallet-advanced-send/components/SendTokenSelector.tsx
```diff
@@ -4,18 +4,28 @@ import { border, cn, color, pressable, text } from '@/styles/theme';
...
-  classNames,
-}: SendTokenSelectorProps) {
+type SendTokenSelectorProps = {
+  classNames?: {
+    container?: string;
+    tokenName?: string;
+    tokenValue?: string;
+    fiatValue?: string;
+    action?: string;
+  };
```
cpcramer, line 16: It looks like all of these classes are applied in the same element. Is it necessary to separate them in the prop definition?

src/internal/hooks/useExchangeRate.tsx
brendan-defi: When I use `useExchangeRate` in `SendProvider`, I'm passing in a `PortfolioTokenWithFiatValue.address`:
image

`PortfolioTokenWithFiatValue` inherits from `Token` type. 
image

And Token allows empty string for address:
image

i suppose I could add a type assertion in SendProvider if that's preferred:
image

src/internal/hooks/useExchangeRate.tsx
brendan-defi: right now `useExchangeRate` is only used in Wallet, but I could imagine this being useful in many other contexts. Happy to move if you prefer, but I think it's very likely we'll end up moving to internal later

src/wallet/components/wallet-advanced-send/components/SendAmountInputTypeSwitch.tsx
brendan-defi: not sure i understand. the only thing earlier is an earlier return for loading state:
image

src/wallet/components/wallet-advanced-send/components/SendProvider.tsx
brendan-defi: fixed

src/internal/hooks/useExchangeRate.tsx
alessey: could update the type to reflect that address: '' is only valid for ETH, but in general, I try not to pass through implementation details from other components.

brendan-defi: i think i have a better solve. in the provider, i'll test against address, not symbol. this will resolve the type narrowing issue:

```token: selectedToken?.address === '' ? 'ETH' : selectedToken?.address,```

**Review by cpcramer — APPROVED** [review]

File: src/internal/hooks/useExchangeRate.tsx (hunk 7 above)
Comment (brendan-defi) [reply src/internal/hooks/useExchangeRate.tsx]: @alessey, based on this feedback I created `usePriceQuote` hook and integrated into useExchangeRate. please let me know if that's what you were suggesting

File: src/wallet/components/wallet-advanced-send/components/SendProvider.tsx [hunk 12]
```diff
@@ -0,0 +1,279 @@
...
+        error: 'Uncaught send transaction error',
+        message: String(error),
+      });
+    }
+  }, [
+    selectedRecipientAddress,
+    selectedToken,
+    cryptoAmount,
+    handleTransactionError,
+  ]);
```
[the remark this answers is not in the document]
brendan-defi: you're right, `callData` is only used by `SendButton`, so I can move there. That component isn't in this PR. Would you prefer me to remove these lines now, or keep them for clarity and move them over when we incorporate the button?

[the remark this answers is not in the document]
alessey: i don't think they add any clarity if you are going to move them, but up to you.

alessey (approved):

src/internal/hooks/usePriceQuote.ts
```diff
@@ -0,0 +1,52 @@
...
+  params: UsePriceQuoteParams<GetPriceQuoteResponse>,
+  _context: RequestContext = RequestContext.Hook,
+): UseQueryResult<GetPriceQuoteResponse> {
+  const { token, queryOptions } = params;
+  const { enabled, cacheTime, staleTime, refetchOnWindowFocus } = {
+    ...DEFAULT_QUERY_OPTIONS,
+    ...queryOptions,
+  };
+
+  const queryKey = ['getPriceQuote', token];
```
dschlabach, line 27: nit, don't need to define a variable here just use it in the hook

src/wallet/components/wallet-advanced-send/components/SendProvider.tsx
```diff
@@ -0,0 +1,283 @@
...
+  const [selectedToken, setSelectedToken] =
+    useState<PortfolioTokenWithFiatValue | null>(null);
+  const [selectedInputType, setSelectedInputType] = useState<'fiat' | 'crypto'>(
+    'crypto',
+  );
+  const [fiatAmount, setFiatAmount] = useState<string | null>(null);
+  const [cryptoAmount, setCryptoAmount] = useState<string | null>(null);
+
+  // state for transaction data
+  const [callData, setCallData] = useState<Call | null>(null);
```
dschlabach, line 60: nit - calldata is the solidity world so we should just do `calldata, setCalldata`

src/wallet/components/wallet-advanced-send/components/SendProvider.tsx
```diff
@@ -0,0 +1,283 @@
...
+  }, [
+    selectedRecipientAddress,
+    selectedToken,
+    cryptoAmount,
+    handleTransactionError,
+  ]);
+
+  useEffect(() => {
+    fetchTransactionData();
+  }, [fetchTransactionData]);
```
dschlabach, line 258: I think you should move this into an internal util hook that uses React Query:
1. It'll improve readability by moving fetching logic into its own module
2. It'll be easier to handle things like refetching, cache management in case we ever need to tweak those parameters

src/wallet/components/wallet-advanced-send/components/SendProvider.tsx
```diff
@@ -0,0 +1,283 @@
...
+        formatUnits(BigInt(ethBalance.cryptoBalance), ethBalance.decimals),
+      ),
+    );
+    updateLifecycleStatus({
+      statusName: 'selectingAddress',
+      statusData: {
+        isMissingRequiredField: true,
+      },
+    });
+  }, [tokenBalances, updateLifecycleStatus]);
```
dschlabach, line 117: It seems like `ethBalance` could be derived state (it doesn't feel like it should be a side effect of `tokenBalances` loading in). 

I wasn't sure what you were doing with the lifecycle statuses here but maybe there's a way to move them somewhere else where it's not a side effect of `tokenBalances` either?

src/wallet/components/wallet-advanced-send/components/SendProvider.tsx
brendan-defi: makes sense. does this hook belong in src/transaction?

src/wallet/components/wallet-advanced-send/components/SendProvider.tsx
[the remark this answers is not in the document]
brendan-defi: per @dschlabach suggestion below, i'll extract this into a separate hook

src/internal/hooks/usePriceQuote.ts
brendan-defi: updated

src/wallet/components/wallet-advanced-send/hooks/useSendTransaction.tsx
```diff
@@ -0,0 +1,97 @@
...
+}: UseSendTransactionParams): UseSendTransactionResponse {
+  const [calldata, setCalldata] = useState<Call | null>(null);
+  const [error, setError] = useState<APIError | null>(null);
+
+  useEffect(() => {
+    const { calldata, error } = getTransactionData(
+      recipientAddress,
+      token,
+      amount,
+    );
```
alessey, line 33: do you need a useEffect and state for this?  it looks like its all sync, could you just return getTransactionData?

src/wallet/components/wallet-advanced-send/hooks/useSendTransaction.tsx
```diff
@@ -0,0 +1,97 @@
...
+      };
+    }
+
+    return {
+      calldata: sendTransaction,
+      error: null,
+    };
+  } catch (err) {
+    return {
+      calldata: null,
```
alessey, line 89: what possible errors is this catching?

src/wallet/components/wallet-advanced-send/components/SendProvider.tsx
```diff
@@ -0,0 +1,250 @@
...
+    handleResetTokenSelection,
+    fiatAmount,
+    handleFiatAmountChange,
+    cryptoAmount,
+    handleCryptoAmountChange,
+    exchangeRate,
+    exchangeRateLoading,
+    selectedInputType,
+    setSelectedInputType,
+    calldata,
```
alessey, line 246: does this callData even need to be in the provider?  can you just build it where you use it?

brendan-defi: i'll move out of the provider

src/wallet/components/wallet-advanced-send/hooks/useSendTransaction.tsx
brendan-defi: i will remove state for this, return the values directly

src/wallet/components/wallet-advanced-send/hooks/useSendTransaction.tsx
brendan-defi: because buildSendTransaction already catches errors and returns an error object instead of throwing, this really only catches issues with parseUnits. the most likely (though still not particularly likely) is more decimals than the token supports.

src/internal/hooks/useSendTransaction.tsx
brendan-defi, line 1: converted to util and renamed getSendTransaction

src/wallet/components/wallet-advanced-send/components/SendProvider.tsx
```diff
@@ -0,0 +1,250 @@
...
+      });
+    },
+    [updateLifecycleStatus],
+  );
+
+  const { calldata, error: buildSendTransactionError } = useSendTransaction({
+    recipientAddress: selectedRecipientAddress.value,
+    token: selectedToken,
+    amount: cryptoAmount,
+  });
```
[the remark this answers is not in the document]
brendan-defi: removing from provider, adding to SendButton in a follow-up PR

src/wallet/components/wallet-advanced-send/utils/getSendTransaction.ts
```diff
@@ -0,0 +1,75 @@
...
+  token: Token | null;
+  amount: string | null;
+};
+
+type GetSendTransactionResponse = {
+  calldata: Call | null;
+  error: APIError | null;
+};
+
+export function getSendTransaction({
```
dschlabach, line 19: IMO could call this `getSendCalldata` or something which expresses what the function's returning - that's the only bit i was a little confused on here

dschlabach (approved):

cpcramer (approved):

PR #2082, opened 2025-03-05
feat: add minikit provider

alessey:
**What changed? Why?**
* Add MiniKitProvider
  * Provides global access to context 
  * Hooks up event listeners
  * Updates internal context state with notification details
  * Adds SafeAreaInsets
* Extract Default wagmi/query provider logic to DefaultOnchainKitProviders to reuse allowing consumers to wrap MiniKitProvider with their own wagmi/react-query

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/OnchainKitProvider.tsx
+++ b/src/OnchainKitProvider.tsx
@@ -4,15 +4,12 @@ import {
   ONCHAIN_KIT_CONFIG,
   setOnchainKitConfig,
 } from '@/core/OnchainKitConfig';
-import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { createContext, useMemo } from 'react';
-import { WagmiProvider } from 'wagmi';
+import { DefaultOnchainKitProviders } from './DefaultOnchainKitProviders';
 import OnchainKitProviderBoundary from './OnchainKitProviderBoundary';
 import { DEFAULT_PRIVACY_URL, DEFAULT_TERMS_URL } from './core/constants';
-import { createWagmiConfig } from './core/createWagmiConfig';
 import type { OnchainKitContextType } from './core/types';
 import { COINBASE_VERIFIED_ACCOUNT_SCHEMA_ID } from './identity/constants';
-import { useProviderDependencies } from './internal/hooks/useProviderDependencies';
 import { checkHashLength } from './internal/utils/checkHashLength';
 import type { OnchainKitProviderReact } from './types';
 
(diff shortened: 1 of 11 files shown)


src/minikit/types.ts
```diff
@@ -0,0 +1,20 @@
+import type { Context, FrameNotificationDetails } from '@farcaster/frame-sdk';
+
+export type UpdateClientContextParams = {
+  details?: FrameNotificationDetails | null;
+  frameAdded?: boolean;
+};
+
+export type MiniKitProviderReact = {
+  children: React.ReactNode;
+  notificationProxyUrl?: string;
```
cpcramer, line 10: What is the `notificationProxyUrl`? Can we add a comment

src/minikit/MiniKitProvider.tsx
```diff
@@ -0,0 +1,148 @@
+'use client';
+
+import { DefaultOnchainKitProviders } from '@/DefaultOnchainKitProviders';
+import { OnchainKitProvider } from '@/OnchainKitProvider';
+import type { OnchainKitProviderReact } from '@/types';
+import sdk, { type Context } from '@farcaster/frame-sdk';
```
cpcramer, line 6: sdk is very general, maybe we can rename this to `frameSDK`?

cpcramer (dismissed):

cpcramer (approved):

src/OnchainKitProvider.tsx
```diff
@@ -4,15 +4,12 @@ import {
   ONCHAIN_KIT_CONFIG,
   setOnchainKitConfig,
 } from '@/core/OnchainKitConfig';
-import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { createContext, useMemo } from 'react';
-import { WagmiProvider } from 'wagmi';
+import { DefaultOnchainKitProviders } from './DefaultOnchainKitProviders';
```
brendan-defi, line 8: What's the context for moving this to its own file?

src/core/createWagmiConfig.ts
```diff
@@ -11,16 +11,17 @@ export const createWagmiConfig = ({
   apiKey,
   appName,
   appLogoUrl,
+  connectors = [
```
brendan-defi, line 14: do we consistently support non-CBW connectors?

src/minikit/MiniKitProvider.tsx
```diff
@@ -0,0 +1,148 @@
...
+      });
+    },
+    [],
+  );
+
+  const connectors = useMemo(() => {
+    return [
+      context // if context is set, the app is running in a frame, use farcasterFrame connector
+        ? farcasterFrame()
+        : coinbaseWallet({
```
brendan-defi, line 109: why only coinbaseWallet as the fallback connector?

src/minikit/MiniKitProvider.tsx
```diff
@@ -0,0 +1,148 @@
...
+              paddingTop: context?.client.safeAreaInsets?.top ?? 0,
+              paddingBottom: context?.client.safeAreaInsets?.bottom ?? 0,
+              paddingLeft: context?.client.safeAreaInsets?.left ?? 0,
+              paddingRight: context?.client.safeAreaInsets?.right ?? 0,
+            }}
+          >
+            {children}
+          </div>
+        </OnchainKitProvider>
+      </DefaultOnchainKitProviders>
```
brendan-defi, line 145: doesn't `OnchainKitProvider` already create a `DefaultOnchainKitProviders`?

src/OnchainKitProvider.tsx
alessey: I'm reusing this logic in the MiniKitProvider.  I will be providing default connectors in the MinKitProvider, but want them to be overridable.

src/core/createWagmiConfig.ts
alessey: since this is our internal wagmi config, we have only in the past supported cbwallet.

src/minikit/MiniKitProvider.tsx
alessey: its not a fallback, coinbaseWallet will not work in a frames webview, we can only use it when the app is run outside of a frames context.

src/minikit/MiniKitProvider.tsx
alessey: yes, I extracted the logic that checks whether you have set a wagmi/query provider so I can reuse that in minikit, allowing minikit to set a custom config.

dschlabach (approved):

PR #2085, opened 2025-03-05
fix: Update caching for Identity hooks

cpcramer:
**What changed? Why?**

Fixed excessive network requests by properly implementing React Query caching across multiple hooks:
- Adding caching for Avatar, ENS, Basenames, and Socials (twitter, farcaster, github, etc.)
- Performance benefits impacting multiple different hooks and components such as: Identity, Swap, Wallet, and NFT components.
- Added proper staleTime configuration to prevent redundant fetches on component re-renders
- Simplified query functions for better maintainability
- Ensured consistent cache configuration across all hooks
- This resolves an issue where opening the WalletDropdown triggered ~7 network requests each time (~200ms for each call), potentially causing API throttling. Data is now properly cached between interactions, significantly reducing API calls and improving performance.

Every time the WalletDropdown was clicked and re-rendered, it would trigger new network requests. This happened because:
```
Default staleTime is 0: React Query considers data "stale" immediately after it's fetched when no staleTime is specified. This means:
By default, every time a component re-renders, React Query will trigger a background refetch
Even though the data is cached, it's marked as "needs refreshing" right away
```

Before:
Clipboard-20250305-233243-134

After:
Clipboard-20250305-233414-397

**Notes to reviewers**

**How has it been tested?**


src/identity/hooks/useAddress.test.tsx
```diff
@@ -78,4 +78,69 @@ describe('useAddress', () => {
...
+    );
+
+    // Wait for the hook to finish fetching the ENS address
+    await waitFor(() => {
+      expect(result.current.data).toBe(testEnsAddress);
+    });
+
+    // The query should be executed with the custom staleTime
+    expect(mockGetEnsAddress).toHaveBeenCalled();
+  });
```
brendan-defi, line 145: what about this expect statement ensures that it's been called with the custom staleTime?

cpcramer: thanks the useAddress code is actually not ready yet - removed

src/identity/hooks/useAvatar.ts
```diff
@@ -14,14 +15,19 @@ export const useAvatar = (
...
+
   return useQuery<GetAvatarReturnType>({
-    queryKey: ['useAvatar', ensName, chain.id],
-    queryFn: async () => {
-      return getAvatar({ ensName, chain });
-    },
+    queryKey,
+    queryFn: () => getAvatar({ ensName, chain }),
     gcTime: cacheTime,
+    staleTime: staleTime,
```
brendan-defi, line 29: nit, you can use shorthand property notation

src/identity/hooks/useAvatar.test.tsx
```diff
@@ -134,4 +134,69 @@ describe('useAvatar', () => {
...
+    );
+
+    // Wait for the hook to finish fetching the ENS avatar
+    await waitFor(() => {
+      expect(result.current.data).toBe(testEnsAvatar);
+    });
+
+    // The query should be executed with the custom staleTime
+    expect(mockGetEnsAvatar).toHaveBeenCalled();
+  });
```
brendan-defi, line 201: what about this statement ensures that the query is executed with the custom staleTime?

src/wallet/hooks/useGetETHBalance.ts
```diff
@@ -11,7 +12,14 @@ import type { UseGetETHBalanceResponse } from '../types';
 const ETH_DECIMALS = 18;
 
 export function useGetETHBalance(address?: Address): UseGetETHBalanceResponse {
-  const ethBalanceResponse: UseBalanceReturnType = useBalance({ address });
+  const ethBalanceResponse: UseBalanceReturnType = useBalance({
+    address,
+    query: {
+      gcTime: DEFAULT_QUERY_OPTIONS.gcTime,
+      staleTime: DEFAULT_QUERY_OPTIONS.staleTime,
+      refetchOnWindowFocus: DEFAULT_QUERY_OPTIONS.refetchOnWindowFocus,
```
brendan-defi, line 20: nit: can you spread `DEFAULT_QUERY_OPTIONS`?

src/identity/hooks/useAvatar.ts
cpcramer: good catch, updated

src/identity/hooks/useAvatar.test.tsx
cpcramer: I just removed the test comments

brendan-defi (approved):

dschlabach (approved):

PR #2102, opened 2025-03-07
feat: Batch support for Basenames and ENS names

cpcramer:
**What changed? Why?**
Added identity resolution components for ENS and Basename support in OnchainKit. This includes:
- `useNames` hook for resolving addresses to ENS/Basename names
- Supporting utility function (`getNames`) with proper chain fallback logic

- The implementation follows the chain fallback pattern where Base-specific resolution is attempted first, then falls back to mainnet ENS resolution when needed
- React Query is used for efficient caching and data fetching

**Notes to reviewers**

**How has it been tested?**
Without batching. 48 requests: 
Screenshot 2025-03-10 at 2 34 42 PM

With batching. 14 requests:
Screenshot 2025-03-10 at 2 38 46 PM


src/identity/hooks/useNames.ts
```diff
@@ -0,0 +1,37 @@
...
+  chain?: typeof mainnet;
+};
+
+/**
+ * A React hook that leverages the `@tanstack/react-query` for fetching and optionally caching
+ * multiple Basenames or ENS names in a single batch request.
+ */
+export const useNames = (
+  { addresses, chain = mainnet }: UseNamesOptions,
+  queryOptions?: UseQueryOptions,
```
dschlabach, line 19: Ideally, allow dev to pass through any query options that Tanstack supports - will make the hook a lot more useful to support specific use cases

src/identity/types.ts
```diff
@@ -1,5 +1,6 @@
 import type { HTMLAttributes, ImgHTMLAttributes, ReactNode } from 'react';
 import type { Address, Chain } from 'viem';
+import type { mainnet } from 'viem/_types/chains/definitions/mainnet';
```
alessey, line 3: import type { Chain } from 'viem'; ?

dschlabach (approved):

src/identity/types.ts
```diff
@@ -186,6 +187,16 @@ export type GetName = {
...
 export type GetNameReturnType = string | Basename | null;
 
+/**
+ * Note: exported as public Type
+ */
+export type GetNames = {
+  /** Array of Ethereum addresses to resolve names for */
+  addresses: Address[];
+  /** Optional chain for domain resolution */
+  chain?: typeof mainnet;
```
alessey, line 197: why wouldn't this be Chain?

cpcramer: It should be Chain, good catch

src/identity/hooks/useNames.ts
cpcramer: Will fast follow along with Avatars.

alessey: can you expose gcTime instead of cacheTime so that when you switch to Tanstack query options it won't be a breaking change and we won't have to support both?

alessey (approved):

abcrane123 (approved):

PR #2107, opened 2025-03-11
feat: Batch support for Base and Ethereum Avatars 

cpcramer:
**What changed? Why?**
Added identity resolution components for ENS and Basename support in OnchainKit. This includes:
- `useAvatars` hook for batch fetching ENS/Basename avatars
- Supporting utility function (`getAvatars`) with proper chain fallback logic

- The implementation follows the chain fallback pattern where Base-specific resolution is attempted first, then falls back to mainnet ENS resolution when needed
- React Query is used for efficient caching and data fetching

**Notes to reviewers**

**How has it been tested?**
Without batching. 48 requests: 
Screenshot 2025-03-10 at 2 34 42 PM

With batching. 14 requests:
Screenshot 2025-03-10 at 2 38 46 PM


src/identity/utils/getAvatars.ts
```diff
@@ -0,0 +1,131 @@
...
+import { isBase } from '@/core/utils/isBase';
+import { isEthereum } from '@/core/utils/isEthereum';
+import type { Basename, GetAvatarReturnType } from '@/identity/types';
+import { mainnet } from 'viem/chains';
+import { normalize } from 'viem/ens';
+import { RESOLVER_ADDRESSES_BY_CHAIN_ID } from '../constants';
+import { getBaseDefaultProfilePicture } from './getBaseDefaultProfilePicture';
+import { isBasename } from './isBasename';
+
+export type GetAvatars = {
```
dschlabach, line 11: probably GetAvatarsParams

src/identity/hooks/useAvatars.ts
```diff
@@ -0,0 +1,36 @@
...
+  chain?: typeof mainnet;
+};
+
+/**
+ * A React hook that leverages the `@tanstack/react-query` for fetching and optionally caching
+ * multiple Basenames or ENS avatars in a single batch request.
+ */
+export const useAvatars = (
+  { ensNames, chain = mainnet }: UseAvatarsOptions,
+  queryOptions?: UseQueryOptions,
```
dschlabach, line 18: We should allow the caller to update any of the query options, not just ones that we define. I think @alessey did this for an NFT hook

dschlabach (approved):

src/identity/hooks/useAvatars.ts
cpcramer: Good idea - I think I might implement this as a fast follow for Avatars and Names. Will reference the NFT work!

abcrane123: +1

abcrane123 (approved):
agree with Dan's comments but other than that lgtm

src/identity/types.ts
```diff
@@ -275,21 +286,31 @@ export type UseAvatarOptions = {
...
-  enabled?: boolean;
-  /** Cache time in milliseconds */
-  cacheTime?: number;
-  /** Stale time in milliseconds */
-  staleTime?: number;
+export type UseAvatarsOptions = {
+  /** Array of ENS names to resolve avatars for */
+  ensNames: string[];
+  /** Optional chain for domain resolution */
+  chain?: typeof mainnet;
```
cpcramer, line 293: updating to Chain

src/identity/utils/getAvatars.ts
cpcramer: There were duplicate type definitions so I updated to use the type from the index.ts file. I agree it should be GetAvatarsParams but if we want to remain consistent with the other exported types then it should be left as GetAvatars.

abcrane123 (dismissed):

alessey (approved):

dschlabach (approved):

PR #2108, opened 2025-03-11
feat: implement default children for Transaction

abcrane123:
**What changed? Why?**

Default transaction components can be implemented like:

```
<Transaction />

<TransactionToast />

<TransactionStatus />

```

**Notes to reviewers**

**How has it been tested?**


src/transaction/components/Transaction.tsx
```diff
@@ -3,14 +3,25 @@ import { useTheme } from '../../internal/hooks/useTheme';
...
+import { TransactionToast } from './TransactionToast';
+
+function TransactionDefaultContent() {
+  return (
+    <>
+      <TransactionButton />
+      <TransactionToast />
+    </>
+  );
+}
```
alessey, line 17: this can just be a fragment

src/transaction/components/Transaction.tsx
```diff
@@ -3,14 +3,25 @@ import { useTheme } from '../../internal/hooks/useTheme';
...
 
 export function Transaction({
   calls,
   capabilities,
   chainId,
   className,
-  children,
+  children = <TransactionDefaultContent />,
```
alessey, line 24: the interface of Transaction with no children should match the current TransactionDefault, including the disabled prop

src/transaction/components/TransactionStatus.tsx
```diff
@@ -1,8 +1,19 @@
...
+import { TransactionStatusLabel } from './TransactionStatusLabel';
+
+function TransactionStatusDefaultContent() {
+  return (
+    <>
+      <TransactionStatusLabel />
+      <TransactionStatusAction />
+    </>
+  );
+}
```
alessey, line 13: fragment

src/transaction/components/TransactionToast.tsx
```diff
@@ -2,9 +2,22 @@ import { useCallback } from 'react';
...
+import { TransactionToastIcon } from './TransactionToastIcon';
+import { TransactionToastLabel } from './TransactionToastLabel';
+
+function TransactionToastDefaultContent() {
+  return (
+    <>
+      <TransactionToastIcon />
+      <TransactionToastLabel />
+      <TransactionToastAction />
+    </>
```
alessey, line 15: fragment

playground/nextjs-app-router/components/demo/TransactionDefault.tsx
```diff
@@ -31,14 +31,13 @@ export default function TransactionDefaultDemo() {
...
     <div className="mx-auto grid w-1/2 gap-8">
-      <TransactionDefault
+      <Transaction
         chainId={chainId ?? 84532} // something breaks if we don't have default network?
         {...(transactionType === TransactionTypes.Calls
           ? { calls }
           : { calls: contracts })}
         capabilities={capabilities}
         onStatus={handleOnStatus}
-        disabled={!chainId && !transactionType}
```
alessey, line 41: we should be able to support this

src/transaction/components/Transaction.tsx
```diff
@@ -41,6 +44,7 @@ export function Transaction({
       capabilities={capabilities}
       chainId={accountChainId}
       contracts={contracts}
+      disabled={disabled}
```
alessey, line 47: instead of supporting disabled in the provider and the button, can you just pass disabled through to the `<TransactionButton default/>` and then set up your types to only allow disabled when !children?

dschlabach (approved):

PR #2115, opened 2025-03-12
feat: implement wallet default children

abcrane123:
**What changed? Why?**
- implement default children so Wallet can be rendered like `<Wallet />`

Screenshot 2025-03-17 at 2 28 48 PM

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/wallet/components/Wallet.tsx
+++ b/src/wallet/components/Wallet.tsx
@@ -13,9 +13,15 @@ import { ConnectWallet } from './ConnectWallet';
 import { WalletAdvanced } from './WalletAdvanced';
 import { WalletDropdown } from './WalletDropdown';
 import { WalletProvider, useWalletContext } from './WalletProvider';
+import { Avatar, Name } from '@/identity';
 
 const defaultWalletChildren = {
-  connect: <ConnectWallet />,
+  connect: (
+    <ConnectWallet>
+      <Avatar className="h-6 w-6" key="avatar" />
+      <Name key="name" />
+    </ConnectWallet>
+  ),
   dropdown: <WalletDropdown />,
   advanced: null,
 };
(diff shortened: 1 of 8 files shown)


src/wallet/components/Wallet.tsx
```diff
@@ -60,6 +72,10 @@ function WalletContent({
   useOutsideClick(walletContainerRef, handleClose);
 
   const { connect, dropdown, advanced } = useMemo(() => {
+    // default children implementation
+    if (!children) {
+      return defaultWalletChildren;
```
dgca, line 77: note: Because the default children will have referential equality across renders (since they're defined outside the component), they won't rerender when the parent rerenders. This is almost certainly fine in this case, but just wanted to call it out since it can be unexpected.

src/wallet/components/WalletDropdown.tsx
```diff
@@ -18,7 +38,12 @@ export function WalletDropdown({ children, className }: WalletDropdownReact) {
   } = useWalletContext();
 
   const childrenArray = useMemo(() => {
-    return Children.toArray(children).map((child) => {
+    // default children implementation
+    const childrenToClone = children
+      ? Children.toArray(children)
+      : defaultWalletDropdownChildren;
```
dgca, line 44: note: Related to the previous comment, but here the children _will_ rerender since they're being cloned below and will therefore _not_ have the same reference.

dgca (approved):
suggestion(non-blocking): Super minor and mostly preference, but I'd consider the default children implementation self-commenting, so I'd skip the explicit comment

cpcramer (approved):

cpcramer:
nit: Add changeset

PR #2116, opened 2025-03-12
feat: Update identity hooks to support TanStack Query options

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/identity/hooks/useAddress.ts
+++ b/src/identity/hooks/useAddress.ts
@@ -1,24 +1,28 @@
+import { getAddress } from '@/identity/utils/getAddress';
+import { DEFAULT_QUERY_OPTIONS } from '@/internal/constants';
+import { useQuery } from '@tanstack/react-query';
+import { mainnet } from 'viem/chains';
 import type {
   GetAddressReturnType,
   UseAddressOptions,
   UseQueryOptions,
-} from '@/identity/types';
-import { getAddress } from '@/identity/utils/getAddress';
-import { useQuery } from '@tanstack/react-query';
-import { mainnet } from 'viem/chains';
+} from '../types';
 
 export const useAddress = (
   { name, chain = mainnet }: UseAddressOptions,
-  queryOptions?: UseQueryOptions,
+  queryOptions?: UseQueryOptions<GetAddressReturnType>,
 ) => {
-  const { enabled = true, cacheTime } = queryOptions ?? {};
+  const { enabled, cacheTime, staleTime, refetchOnWindowFocus } = {
+    ...DEFAULT_QUERY_OPTIONS,
+    ...queryOptions,
+  };
+
   return useQuery<GetAddressReturnType>({
     queryKey: ['useAddress', name, chain.id],
-    queryFn: async () => {
-      return await getAddress({ name, chain });
-    },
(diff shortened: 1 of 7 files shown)


src/internal/constants.ts
```diff
@@ -12,9 +12,10 @@ export const DEFAULT_STALE_TIME = 1000 * 60 * 5;
  * Default query options used across hooks
  */
 export const DEFAULT_QUERY_OPTIONS = {
-  enabled: true,
```
cpcramer, line 15: Didn't seem necessary

src/identity/hooks/useAvatar.ts
```diff
@@ -13,21 +13,15 @@ import type {
...
   const queryKey = ['useAvatar', ensName, chain.id];
 
   return useQuery<GetAvatarReturnType>({
     queryKey,
     queryFn: () => getAvatar({ ensName, chain }),
-    gcTime: cacheTime,
-    staleTime,
-    enabled,
-    refetchOnWindowFocus,
+    enabled: !!ensName,
```
alessey, line 23: can you support cacheTime as well so this won't be a breaking change and mark it deprecated?

src/identity/hooks/useAvatar.ts
```diff
@@ -13,21 +13,15 @@ import type {
...
+    ...DEFAULT_QUERY_OPTIONS,
+    ...queryOptions,
```
alessey, line 25: isn't spreading DEFAULT_QUERY_OPTIONS invalid as it has cacheTime which isn't a valid tanstack option?

src/identity/hooks/useAddress.ts
```diff
@@ -4,21 +4,21 @@ import type {
...
   return useQuery<GetAddressReturnType>({
-    queryKey: ['useAddress', name, chain.id],
-    queryFn: async () => {
-      return await getAddress({ name, chain });
-    },
-    gcTime: cacheTime,
-    enabled,
-    refetchOnWindowFocus: false,
+    queryKey,
+    queryFn: () => getAddress({ name, chain }),
```
brendan-defi, line 19: this doesn't need to be async await?

dschlabach (approved):

src/identity/hooks/useAvatar.ts
cpcramer: Do we want `cacheTime` to take priority over `gcTime` if both are provided?

cpcramer: Updated. Added cacheTime as a deprecated param.

src/identity/hooks/useAddress.ts
cpcramer: Both approaches work, but since getAddress already returns a Promise, using async/await would unnecessarily create a second Promise wrapper. TanStack automatically waits for any Promise returned by queryFn, regardless of whether async/await syntax is used.

cpcramer: This is also what we are doing in the other hooks

brendan-defi: gotcha, makes sense thanks

brendan-defi (approved):

alessey (approved):

PR #2133, opened 2025-03-15
fix: Make TokenRow text-overflow: ellipsis work correctly

dgca:
**What changed? Why?**

Fixes the styling in the TokenRow component so that text-overflow ellipsis works as intended. Currently, a token with a long name will cause the container to be horizontally scrollable and the text will overflow out of its container.

Screenshot 2025-03-15 at 3 37 48 PM

**Notes to reviewers**

n/a

**How has it been tested?**

Manually in the playground app


brendan-defi:
before: 
image

after: 
image

brendan-defi (approved):

dschlabach (approved):

PR #2140, opened 2025-03-18
feat: Add Badge tooltip display

cpcramer:
**What changed? Why?**
The attestation badge can be confusing, it's simply a checkmark that displays next to the users wallet name or address. This tooltip will make it clear what the Badge represents. It will also allow the developer to customize the message.
**Notes to reviewers**

**How has it been tested?**

`<IdentityCard /> // No tooltip.` 

`<IdentityCard badgeTooltip={true} /> // Tooltip with attestation name` 
Screenshot 2025-03-18 at 2 51 33 PM

`<IdentityCard badgeTooltipText="Custom text" /> // Tooltip with "Custom text"`
Screenshot 2025-03-18 at 2 52 52 PM

```
<Identity>
       <Badge /> // No tooltip
</Identity>
```

```
<Identity>
       <Badge tooltip={true} /> // Tooltip with attestation name
</Identity>
```

```
<Identity>
       <Badge tooltipText="Custom text" /> // Tooltip with "Custom text
</Identity>
```

With themes:
Screenshot 2025-03-19 at 11 24 00 AM

### DIFF
--- a/src/identity/components/Badge.tsx
+++ b/src/identity/components/Badge.tsx
@@ -28,14 +28,14 @@ export function Badge({ className, tooltip = false }: BadgeReact) {
   const { address, schemaId: contextSchemaId } = useIdentityContext();
   const { chain, schemaId: kitSchemaId } = useOnchainKit();
 
+  // Only enable tooltip if tooltip is true or a string
+  const showTooltipFeature = Boolean(tooltip);
+
   const attestations = useAttestations({
     address,
     chain,
-    schemaId: contextSchemaId ?? kitSchemaId,
+    schemaId: showTooltipFeature ? contextSchemaId ?? kitSchemaId : null,
   });
-
-  // Only enable tooltip if tooltip is true or a string
-  const showTooltipFeature = Boolean(tooltip);
 
   // Get tooltip text from tooltip prop or attestation
   const displayText = useMemo(() => {
--- a/src/identity/components/Badge.tsx
+++ b/src/identity/components/Badge.tsx
@@ -4,11 +4,11 @@ import type { BadgeReact } from '@/identity/types';
 import { badgeSvg } from '@/internal/svg/badgeSvg';
 import { zIndex } from '@/styles/constants';
 import { background, border, cn, color, pressable, text } from '@/styles/theme';
-import { useState } from 'react';
+import { useMemo, useState } from 'react';
 import { useOnchainKit } from '../../useOnchainKit';
 import { useIdentityContext } from './IdentityProvider';
 
-type extractAttestationNameParams = {
+type ExtractAttestationNameParams = {
   decodedDataJson?: string;
   id?: string;
   attester?: string;
@@ -23,13 +23,10 @@ type extractAttestationNameParams = {
 /**
  * Badge component.
  */
-export function Badge({ className, tooltip = false, tooltipText }: BadgeReact) {
+export function Badge({ className, tooltip = false }: BadgeReact) {
   const [showTooltip, setShowTooltip] = useState(false);
   const { address, schemaId: contextSchemaId } = useIdentityContext();
   const { chain, schemaId: kitSchemaId } = useOnchainKit();
-
-  // If tooltipText is provided, tooltip should be enabled
-  const showTooltipFeature = tooltipText ? true : tooltip;
 
   const attestations = useAttestations({
     address,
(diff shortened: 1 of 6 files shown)


src/identity/components/Badge.tsx
```diff
@@ -1,30 +1,132 @@
...
-      {badgeSvg}
-    </span>
+    </div>
   );
 }
+
+/**
+ * Extracts the attestation name from an attestation object
+ */
+function extractAttestationName(
```
cpcramer, line 98: This function could be reused in the future and then will move it to a shared utils folder. But for now it's only used by the Badge component.

src/identity/components/Badge.tsx
```diff
@@ -1,30 +1,132 @@
...
 import type { BadgeReact } from '@/identity/types';
 import { badgeSvg } from '@/internal/svg/badgeSvg';
-import { background, cn } from '@/styles/theme';
+import { zIndex } from '@/styles/constants';
+import { background, border, cn, color, pressable, text } from '@/styles/theme';
+import { useState } from 'react';
+import { useOnchainKit } from '../../useOnchainKit';
+import { useIdentityContext } from './IdentityProvider';
+
+type extractAttestationNameParams = {
```
dgca, line 11: nit: Should this be capitalized?

src/identity/components/Badge.tsx
```diff
@@ -1,30 +1,132 @@
...
+        }}
+        data-testid="ockBadge"
+        {...(showTooltipFeature && {
+          onMouseEnter: () => setShowTooltip(true),
+          onMouseLeave: () => setShowTooltip(false),
+        })}
+      >
+        {badgeSvg}
+      </span>
+      {showTooltip && showTooltipFeature && (
```
dgca, line 69: question: I noticed we have a `<Tooltip>` component in `/internal`. Is it possible to use that?

Mainly thinking about accessibility and while the `<Tooltip>` isn't very accessible right now, we could make it more accessible and then not have to write custom a11y logic in other places that use a tooltip.

src/identity/components/Badge.tsx
```diff
@@ -1,30 +1,132 @@
...
+
+  try {
+    const decodedData = JSON.parse(attestation.decodedDataJson);
+
+    if (Array.isArray(decodedData) && decodedData[0]?.name) {
+      return decodedData[0].name;
+    }
+
+    const firstKey = Object.keys(decodedData)[0];
+    const value = decodedData[firstKey];
```
dgca, line 113: suggestion: You could use Object.values and do this in one line. E.g. `const value = Object.values(decodedData)[0]`

src/identity/types.ts
```diff
@@ -67,6 +67,10 @@ export type AvatarReact = {
 export type BadgeReact = {
   /** Optional className override for top span element. */
   className?: string;
+  /** Controls whether the badge shows a tooltip on hover. Defaults to false. */
+  tooltip?: boolean;
+  /** Custom text to display in the badge tooltip. Defaults to the attestation name if not provided. */
+  tooltipText?: string;
```
dgca, line 73: suggestion: This is mostly personal preference, but I tend to avoid props that can contradict each other.

E.g. with the current implementation, this would show the tooltip even though `tooltip={false}`

```tsx
<BadgeReact
  tooltip={false}
  tooltipText="hello"
/>
```

You could avoid this by getting rid of the `tooltipText` prop and making `tooltip: boolean | string`. Then, the content could be derived like

```tsx
const displayText = useMemo(() => {
  if (!tooltip) return null;
  
  return typeof tooltip === "string"
    ? tooltip
    : extractAttestationName(attestations[0]);
}, [tooltip]);
```

Just a suggestion, though. There might be other ways to avoid the conflicting props too

src/identity/components/Badge.tsx
```diff
@@ -1,30 +1,132 @@
...
+  const { chain, schemaId: kitSchemaId } = useOnchainKit();
+
+  // If tooltipText is provided, tooltip should be enabled
+  const showTooltipFeature = tooltipText ? true : tooltip;
+
+  const attestations = useAttestations({
+    address,
+    chain,
+    schemaId: contextSchemaId ?? kitSchemaId,
+  });
```
dgca, line 38: suggestion:  This is gonna make a network request even if `tooltip = false`. It'd be nice to avoid that, if possible.

src/identity/components/Badge.tsx
cpcramer: I'm not sure if the existing component is ready for use. I know @dschlabach looked at it a few weeks ago, but he encountered integration issues. Overall, I noticed that the tooltip component wasn't being used anywhere, so I decided against implementing it. As a follow-up PR, we could update all of our tooltips to use this component/primitive.

src/identity/components/Badge.tsx
cpcramer: Nice, updated

src/identity/types.ts
cpcramer: Great callout. I think i'll update to only show the tooltip if it's explicitly enabled

src/identity/components/Badge.tsx
cpcramer: Good catch, updated to not execute the call when tooltip is false!

dgca (approved):

dgca (approved):

dschlabach (approved):

PR #2142, opened 2025-03-19
chore: refactor Wallet component

abcrane123:
**What changed? Why?**
- refactor Wallet component

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/wallet/components/Wallet.tsx
+++ b/src/wallet/components/Wallet.tsx
@@ -8,23 +8,22 @@ import { useTheme } from '@/internal/hooks/useTheme';
 import { findComponent } from '@/internal/utils/findComponent';
 import { cn } from '@/styles/theme';
 import { Children, useMemo, useRef } from 'react';
-import type { WalletReact, WalletSubComponentReact } from '../types';
+import type { WalletReact } from '../types';
 import { getWalletDraggableProps } from '../utils/getWalletDraggableProps';
 import { ConnectWallet } from './ConnectWallet';
 import { WalletAdvanced } from './WalletAdvanced';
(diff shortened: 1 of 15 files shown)


src/wallet/components/Wallet.tsx
```diff
@@ -51,23 +68,31 @@ function WalletContent({
...
     return {
-      connect: childrenArray.find(findComponent(ConnectWallet)),
       dropdown: childrenArray.find(findComponent(WalletDropdown)),
       advanced: childrenArray.find(findComponent(WalletAdvanced)),
     };
   }, [children]);
 
+  // cannot use advanced and dropdown,
+  // default to dropdown
+  const childrenToRender = useMemo(() => {
```
abcrane123, line 87: putting together a P/PS on how we can do a bigger refactor to remove the need for this and the `findComponent` logic above

src/wallet/components/Wallet.tsx
```diff
@@ -51,23 +68,31 @@ function WalletContent({
...
+    return Children.map(children, (child: ReactNode) => {
+      if (isValidElement(child) && child.type === WalletAdvanced && dropdown) {
+        return null;
```
dgca, line 90: suggestion(non-blocking): Not sure if this would be a breaking change and we don't need to address it since a lot of this is in flux, but as a consumer, I would rather an invalid state throw a descriptive error in this case. It'd be easier to understand why something isn't working as expected

dgca: Another suggestion would be to just return `defaultWalletChildren` here and not do the `||` check on line 118.

dgca (approved):

PR #2156, opened 2025-03-23
fix correction

sky-coderay:
**What changed? Why?**
playground/nextjs-app-router/components/AppProvider.tsx
paymasters is - paymasters are

playground/nextjs-app-router/components/Demo.tsx
(to_bottom,#f0f0f0_1px,transparent_1px)] --`deleted ]`


dgca:
Hey @sky-coderay, I'm gonna close this one since it only affects the playground which doesn't affect the core library.

PR #2162, opened 2025-03-25
chore: Consolidate WalletDropdown and WalletAdvanced into single component

abcrane123:
**What changed? Why?**

- `WalletDropdown` effectively replaces `WalletAdvanced`, supporting both basic and advanced dropdown configurations based on the children provided
- deprecate WalletAdvanced
- refactor `WalletAdvancedProvider` logic into `WalletProvider` and remove `WalletAdvancedProvider`

```
<Wallet>
  <ConnectWallet />
  <WalletDropdown>
      <WalletAdvancedWalletActions />
      <WalletAdvancedAddressDetails />
      <WalletAdvancedTransactionActions />
      <WalletAdvancedTokenHoldings />
  </WalletDropdown>
</Wallet>
```

```
<Wallet>
  <ConnectWallet />
  <WalletDropdown>
    <WalletDropdownLink
        icon="wallet"
        key="wallet-dd-link"
        href="https://keys.coinbase.com"
        target="_blank"
      >
      Wallet
    </WalletDropdownLink>
    <WalletDropdownDisconnect key="wallet-dd-disconnect" />
  </WalletDropdown>
</Wallet>
```

```
<Wallet>
  <ConnectWallet />
  <WalletDropdown />
</Wallet>
```

**Notes to reviewers**

**How has it been tested?**


src/wallet/components/Wallet.tsx
```diff
@@ -51,23 +68,34 @@ function WalletContent({
...
+        return null;
+      }
+      return child;
+    });
+  }, [dropdown, children]);
+
+  // Note: this can be removed after deprecating WalletAdvanced
   if (dropdown && advanced) {
     console.error(
       'Defaulted to WalletDropdown. Wallet cannot have both WalletDropdown and WalletAdvanced as children.',
```
abcrane123, line 101: We cannot remove lines 77-101 until we deprecate `WalletAdvanced` (devs could pass both and get unexpected behavior)

src/wallet/components/WalletAdvancedDefault.tsx
```diff
@@ -4,10 +4,10 @@ import { Avatar, Name } from '@/identity';
...
 import { Wallet } from './Wallet';
-import { WalletAdvanced } from './WalletAdvanced';
+import { WalletDropdown } from './WalletDropdown';
+import { WalletAdvancedTransactionActions } from './WalletAdvancedTransactionActions';
 import { WalletAdvancedAddressDetails } from './WalletAdvancedAddressDetails';
 import { WalletAdvancedTokenHoldings } from './WalletAdvancedTokenHoldings';
-import { WalletAdvancedTransactionActions } from './WalletAdvancedTransactionActions';
 import { WalletAdvancedWalletActions } from './WalletAdvancedWalletActions';
 
 export function WalletAdvancedDefault() {
```
abcrane123, line 13: I know we want to move away from this pattern but given that  we will only have WalletDropdown, the default children implementation will be the basic version. Should we keep this for our suggested advanced version or remove ?

src/wallet/components/WalletDropdown.tsx
```diff
@@ -1,32 +1,48 @@
...
-  const childrenArray = useMemo(() => {
-    return Children.toArray(children).map((child) => {
-      if (isValidElement(child) && child.type === Identity) {
-        // @ts-ignore
-        return cloneElement(child, { address });
-      }
-      return child;
-    });
-  }, [children, address]);
-
```
abcrane123, line 29: i updated Identity to use `useAccount` to fetch address so we no longer need to pass explicitly here.

src/identity/components/Identity.test.tsx
```diff
@@ -73,6 +82,25 @@ describe('Identity Component', () => {
     });
   });
 
+  it('should return null if no address provided', async () => {
+    useAvatarMock.mockReturnValue({
+      data: 'avatar_url',
+      isLoading: false,
+    });
+    useNameMock.mockReturnValue({ data: 'name', isLoading: false });
+    (useAccount as Mock).mockReturnValue({ address: undefined });
```
brendan-defi, line 91: nit — it would be nice if you followed the same mocking pattern as useNameMock and useAvatarMock

src/identity/components/Identity.tsx
```diff
@@ -14,10 +15,15 @@ export function Identity({
 }: IdentityReact) {
   const { chain: contextChain } = useOnchainKit();
   const accountChain = chain ?? contextChain;
+  const { address } = useAccount();
+
+  if (!address && !addressProp) {
+    return null;
+  }
```
brendan-defi, line 22: i'm not sure this is worth solving for, but...

right now, Identity will make a useAccount call on every render. it might be nice to create some sort of intermediary hook that intercepts and returns addressProp if available, and only calls useAccount if addressProp is nullish

src/identity/components/Identity.tsx
```diff
@@ -14,10 +15,15 @@ export function Identity({
...
 
   return (
     <IdentityProvider
-      address={address}
+      address={addressProp || address}
```
brendan-defi, line 26: if you did the above, you wouldn't need this conditional, you could just use the return value from the intercepter hook

src/wallet/components/Wallet.test.tsx
```diff
@@ -224,106 +258,4 @@ describe('Wallet Component', () => {
 
     expect(screen.getByTestId('ockDraggable')).toBeDefined();
   });
-
-  it('should render WalletAdvanced right-aligned when there is not enough space on the right', () => {
```
brendan-defi, line 228: how come these tests are no longer needed?

src/wallet/components/Wallet.tsx
```diff
@@ -1,19 +1,36 @@
...
+  <>
+    <ConnectWallet>
+      <Avatar className="h-6 w-6" key="avatar" />
+      <Name key="name" />
+    </ConnectWallet>
+    <WalletDropdown />
+  </>
+);
+
 export const Wallet = ({
```
brendan-defi, line 34: nit: since we're refactoring, i think we should make this a proper function not an arrow function

src/wallet/components/Wallet.tsx
```diff
@@ -51,23 +68,34 @@ function WalletContent({
     isConnectModalOpen,
     handleClose,
     connectRef,
-    showSubComponentAbove,
-    alignSubComponentRight,
```
brendan-defi, line 55: nice, i like how you handled these

src/wallet/components/Wallet.test.tsx
brendan-defi: nvm, i see, they're moved into the component tests

src/wallet/components/WalletAdvancedAddressDetails.tsx
```diff
@@ -44,7 +44,7 @@ export function WalletAdvancedAddressDetails({
     <div
       data-testid="ockWalletAdvanced_AddressDetails"
       className={cn(
-        'mt-2 flex flex-col items-center justify-center',
+        'mt-2 flex w-88 flex-col items-center justify-center px-4 py-3',
```
brendan-defi, line 47: i still don't love adding these the width and padding here, even though they can be overwritten via classNames.container.

what if, instead of adding here, you put in a wrapper around the default children?

src/wallet/components/WalletAdvancedDefault.tsx
brendan-defi: IMHO we should preserve in some way, but i don't feel tied to this approach

src/wallet/components/WalletDropdown.tsx
```diff
@@ -35,33 +51,28 @@ export function WalletDropdown({ children, className }: WalletDropdownReact) {
...
-        if (isSubComponentClosing) {
-          setIsSubComponentOpen(false);
-          setIsSubComponentClosing(false);
-        }
-      }}
-      data-testid="ockWalletDropdown"
-    >
-      {childrenArray}
-    </div>
+    <WalletAdvancedProvider>
```
brendan-defi, line 59: IMHO we should rename these WalletDropdownProvider and WalletDropdownContent now

brendan-defi:
@abcrane123 let's take this opportunity to export WalletProvider and WalletAdvancedProvider (or WalletDropdownProvider if we rename)?

src/wallet/components/WalletDropdown.tsx
```diff
@@ -35,33 +51,28 @@ export function WalletDropdown({ children, className }: WalletDropdownReact) {
...
+      <div
+        data-testid="ockWalletDropdown"
+        className={cn(
+          'absolute',
+          showSubComponentAbove ? 'bottom-full' : 'top-full',
+          alignSubComponentRight ? 'right-0' : 'left-0',
+          className,
+        )}
+      >
+        <WalletAdvancedContent
```
brendan-defi, line 69: one nice thing we get "for free" with this refactor is that all dropdowns get the positioning logic. 

right now, if you create a draggable "non-advanced" wallet, you get this:
image

but with this change, you get this:
image

src/wallet/components/WalletAdvancedContent.tsx
```diff
@@ -74,7 +74,7 @@ export function WalletAdvancedContent({
       );
     }
 
-    return <ContentWrapper className="px-4 py-3">{children}</ContentWrapper>;
+    return <ContentWrapper className="p-1">{children}</ContentWrapper>;
```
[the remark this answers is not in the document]
abcrane123: there is a minor style issue when we have no padding where the border radius isn't fully respected

Screenshot 2025-03-26 at 12 03 51 PM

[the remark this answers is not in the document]
abcrane123: removing

src/wallet/components/WalletDropdown.tsx
abcrane123: renamed WalletDropdownContent and deleted WalletAdvancedProvider (moved logic to WalletProvider)

src/wallet/components/WalletAdvancedAddressDetails.tsx
abcrane123: we need to have w-88 when advanced children are rendered so i think i need to add the width here, otherwise it looks strange when devs render only basic wallet components. adding wrapper to default wallet children wouldnt work for that case. open to other ideas if you have them though !

src/identity/components/Identity.tsx
abcrane123: yeah good idea i think i'll explore in follow up since this is getting big, unless you're opposed

src/wallet/components/Wallet.tsx
```diff
@@ -108,52 +131,8 @@ function WalletContent({
...
-        alignSubComponentRight={alignSubComponentRight}
-        showSubComponentAbove={showSubComponentAbove}
-      />
+      {/* Note: update childrenToRender to children after deprecating WalletAdvanced */}
+      <div ref={connectRef}>{childrenToRender || defaultWalletChildren}</div>
     </div>
   );
 }
-
-function WalletSubComponent({
```
dgca, line 124: Pretty cool that we can just delete this now 🤘

src/wallet/components/WalletAdvancedTransactionActions.tsx
```diff
@@ -89,14 +92,14 @@ export function WalletAdvancedTransactionActions({
...
-    return <Skeleton className="my-3 h-16 w-80" />;
+    return <Skeleton className="mx-auto my-3 h-16 w-80 px-4 py-3" />;
   }
 
   return (
     <div
       data-testid="ockWalletAdvanced_TransactionActions"
       className={cn(
-        'my-3 flex w-full flex-row justify-between gap-2',
+        'my-3 flex w-88 w-full flex-row justify-between gap-2 px-4 py-3',
```
dgca, line 102: issue: w-88 and w-full clash, no?

src/wallet/components/WalletAdvancedWalletActions.tsx
```diff
@@ -80,7 +83,7 @@ export function WalletAdvancedWalletActions({
     <div
       data-testid="ockWalletAdvanced_WalletActions"
       className={cn(
-        'flex w-full items-center justify-between',
+        'flex w-88 w-full items-center justify-between px-4 py-3',
```
dgca, line 86: Ditto w-88 & w-full comment

src/wallet/components/WalletProvider.tsx
```diff
@@ -65,6 +86,16 @@ export function WalletProvider({ children }: WalletProviderReact) {
...
+    activeFeature,
+    setActiveFeature,
+    isActiveFeatureClosing,
+    setIsActiveFeatureClosing,
+    tokenBalances,
+    portfolioFiatValue,
+    isFetchingPortfolioData,
+    portfolioDataUpdatedAt,
+    refetchPortfolioData,
+    animations,
```
dgca, line 98: nit(non-blocking): Since animations returns an object, this will break memoization. Any non-primitive type that's recreated on render will cause a dependency array to think the value has changed.

It doesn't really matter, though, because `useValue` passes an object (its only argument) to the dependency array.

Personally, I'd get rid of `useValue` and either pass the object directly to the context value (which would make things work the same as they currently do) or use `useMemo` to memoize. If the latter, though, you'd wanna make sure that all the members of this object are also memoized.

src/wallet/components/Wallet.tsx
```diff
@@ -51,23 +68,34 @@ function WalletContent({
...
+  // cannot use advanced and dropdown,
+  // default to dropdown
+  const childrenToRender = useMemo(() => {
+    return Children.map(children, (child: ReactNode) => {
+      if (isValidElement(child) && child.type === WalletAdvanced && dropdown) {
+        return null;
+      }
+      return child;
+    });
+  }, [dropdown, children]);
```
alessey, line 96: if you used advanced and dropdown could you just render children from each in dropdown?

src/wallet/components/WalletAdvancedDefault.tsx
alessey: another possibility would be to mark deprecated and in default implementations, i.e. no children, support a type prop

alessey: another option would be to deprecate this component and provide a type attribute in the default (i.e. no children) state to enable different default children.  The majority of our implementations are most likely default implementations and keeping this default will be odd when all the others are deprecated.

src/wallet/components/WalletAdvancedQrReceive.test.tsx
```diff
@@ -141,13 +135,16 @@ describe('WalletAdvancedQrReceive', () => {
...
       ...defaultMockUseWalletAdvancedContext,
       activeFeature: 'qr',
     });
 
+    // mockUseWalletAdvancedContext.mockReturnValue({
+    //   ...defaultWalletContext,
+    //   ...defaultMockUseWalletAdvancedContext,
+    //   activeFeature: 'qr',
+    // });
+
```
alessey, line 147: nit: delete

src/wallet/components/WalletAdvancedTokenHoldings.tsx
```diff
@@ -30,13 +30,13 @@ export function WalletAdvancedTokenHoldings({
...
   const { tokenBalances, isFetchingPortfolioData, animations } =
-    useWalletAdvancedContext();
+    useWalletContext();
 
   if (isFetchingPortfolioData || !tokenBalances || tokenBalances.length === 0) {
     return (
       <div
         data-testid="ockWalletAdvanced_LoadingPlaceholder"
-        className="my-2 h-44 w-80"
+        className="mx-auto my-2 h-44 w-80 px-4 py-3"
```
alessey, line 39: should this be w-88 to match below?

src/wallet/components/WalletAdvancedTransactionActions.tsx
alessey: would it be possible to w-full all these children and just w-88 on the parent?  allowing width to be overridden in one place?

src/wallet/components/WalletAdvancedDefault.tsx
brendan-defi: yeah, wallet could have two different defaults, and advanced could be accessed via :

`<Wallet default="advanced" />` or `<Wallet advancedDefault />

src/wallet/components/WalletAdvancedTransactionActions.tsx
abcrane123: agree it is not ideal but i'm trying to prevent changing the display for developers who have already integrated the component

src/wallet/components/WalletDropdown.tsx
```diff
@@ -35,33 +50,26 @@ export function WalletDropdown({ children, className }: WalletDropdownReact) {
...
-        pressable.default,
-        color.foreground,
-        zIndex.dropdown,
-        'absolute right-0 mt-1.5 flex w-max min-w-[300px] cursor-default flex-col overflow-hidden rounded-xl',
-        isSubComponentClosing
-          ? 'fade-out slide-out-to-top-1.5 animate-out fill-mode-forwards ease-in-out'
-          : 'fade-in slide-in-from-top-1.5 animate-in duration-300 ease-out',
+        'absolute',
+        showSubComponentAbove ? 'bottom-full' : 'top-full',
+        alignSubComponentRight ? 'right-0' : 'left-0',
```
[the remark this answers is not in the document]
abcrane123: logic handled in `WalletProvider`

src/wallet/components/WalletProvider.tsx
abcrane123: let me know how it looks now !

brendan-defi:
wanted to confirm we're okay with wallet dropdown default alignment change (the alignment will be based on the position of the button/dropdown)
before:
image

after:
image

brendan-defi:
@alissacrane-cb i think we should add a fixed height to top-level send:
image
image

PR #2171, opened 2025-03-26
chore: fix onramp bug

abcrane123:
**What changed? Why?**

The USD estimate for $5 of USDC is slightly off (4.9999). There is a minimum amount of $5 required for onramp options so this PR updates that check to use amount instead of amountUSD for USDC buys. We need to keep amountUSD for non-USDC tokens.

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/src/buy/components/BuyOnrampItem.tsx
+++ b/src/buy/components/BuyOnrampItem.tsx
@@ -1,5 +1,5 @@
 'use client';
-import { SwapUnit } from '@/swap/types';
+import type { SwapUnit } from '@/swap/types';
 import { usdcToken } from '@/token/constants';
 import { useCallback, useMemo } from 'react';
 import { applePaySvg } from '../../internal/svg/applePaySvg';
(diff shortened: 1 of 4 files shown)


dgca (approved):

src/buy/components/BuyOnrampItem.tsx
```diff
@@ -35,9 +37,19 @@ export function BuyOnrampItem({
     onClick();
   }, [onClick, setIsDropdownOpen]);
 
+  const fiatAmount = useMemo(() => {
+    // if token is USDC the usd estimate
+    // can be slightly off (4.9999999) so
+    // use amount instead to prevent disabling of onramp
+    if (to?.token?.symbol === usdcToken?.symbol) {
```
brendan-defi, line 44: would not use symbol here, should use address. symbols are non-unique

brendan-defi (approved):

dgca (dismissed):

dgca (approved):

brendan-defi (approved):

PR #2176, opened 2025-03-27
feat: add minikit to playground

alessey:
**What changed? Why?**
Add a sample MiniKit app to playground/minikit

**Notes to reviewers**

**How has it been tested?**


cpcramer:
How should we test this? Where is MiniKit in the playground UI?

packages/playground/tailwind.config.ts
```diff
@@ -51,11 +95,11 @@ const config: Config = {
...
-    borderRadius: {
-      lg: 'var(--radius)',
-      md: 'calc(var(--radius) - 2px)',
-      sm: 'calc(var(--radius) - 4px)',
-    },
+    // borderRadius: {
+    //   lg: 'var(--radius)',
+    //   md: 'calc(var(--radius) - 2px)',
+    //   sm: 'calc(var(--radius) - 4px)',
+    // },
```
cpcramer, line 102: Delete?

packages/playground/lib/redis.ts
```diff
@@ -0,0 +1 @@
+export const redis = null;
```
cpcramer, line 1: Is this const needed? It just seems a little strange to keep a const always set to null

packages/playground/app/minikit/theme.css
```diff
@@ -0,0 +1,55 @@
+@import url("https://fonts.googleapis.com/css2?family=Pixelify+Sans:wght@400..700&display=swap");
+
+:root {
+  --snake-background: #e5e5e5;
+}
+
+.snake-light,
+.snake-dark {
```
cpcramer, line 8: Should we add these to the Theme selection dropdown?

alessey:
> How should we test this? Where is MiniKit in the playground UI?

its set up as /minikit, locally will be localhost:3000/minikit
https://onchainkit-playground-fgkph52zq-coinbase-vercel.vercel.app/minikit

when deployed it'll be onchainkit.xyz/playground/minikit

packages/playground/app/minikit/theme.css
cpcramer, line 8: How do we adjust the theme?

cpcramer (approved):
No blocking issues

packages/playground/app/api/notify/route.ts
```diff
@@ -0,0 +1,33 @@
...
+      fid,
+      title: notification.title,
+      body: notification.body,
+    });
+
+    if (result.state === "error") {
+      return new Response(JSON.stringify({ error: result.error }), {
+        status: 500,
+        headers: { "Content-Type": "application/json" },
+      });
```
dgca, line 18: suggestion(non-blocking): Just FYI, there's a `NextResponse.json()` util that lets you skip stringifying and setting content type.

E.g.

```ts
NextResponse.json({ error: result.error }, { status: 500 })
```

packages/playground/app/api/webhook/route.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  const { header: encodedHeader, payload: encodedPayload } = requestJson;
+
+  const headerData = decode(encodedHeader);
+  const event = decode(encodedPayload);
+
+  const { fid, key } = headerData;
+
+  const valid = await verifyFidOwnership(fid, key);
+  console.log("valid", valid);
```
dgca, line 70: nit: I know it's the playground but I feel like this could either be deleted or logged in a more descriptive way (e.g. `Is fid valid: {valid}`)

packages/playground/app/api/webhook/route.ts
```diff
@@ -0,0 +1,128 @@
...
+          fid,
+          title: `Welcome to ${appName}`,
+          body: `Thank you for adding ${appName}`,
+        });
+      } else {
+        console.log(
+          "frame_added",
+          "event.notificationDetails",
+          event.notificationDetails,
+        );
```
dgca, line 98: issue: This console.log is the same as lines 82-86. Is that intentional? If so, we could put the log log before the if statement since it gets called unconditionally

packages/playground/app/minikit/components/snake.tsx
```diff
@@ -0,0 +1,1118 @@
...
+          attestationUid: attestation.id,
+          transactionHash: attestation.txid,
+        };
+      }
+      return null;
+    })
+    .sort((a: Score, b: Score) => b.score - a.score);
+}
+
+function useKonami(gameState: number) {
```
dgca, line 185: image

packages/playground/app/minikit/theme.css
alessey: you cannot adjust the theme, this is the theme for the minikit game

packages/playground/app/api/webhook/route.ts
alessey: I can remove from here.  I had left these in the actual demo template since it's a demo.

packages/playground/lib/redis.ts
alessey: I left this in in case we want to enable in the future, but short circuited here instead of removing the files that reference redis.

dgca (approved):

dschlabach (approved):

PR #2178, opened 2025-03-27
feat: Add Rabby, Trust, and Frame wallet connection support

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


packages/onchainkit/src/core/OnchainKitConfig.ts
```diff
@@ -23,6 +23,11 @@ export const ONCHAIN_KIT_CONFIG: OnchainKitConfig = {
       display: null,
       termsUrl: null,
       privacyUrl: null,
+      supportedWallets: {
+        rabby: false,
+        trust: false,
+        frame: false,
+      },
```
cpcramer, line 30: To test locally, enable the wallets here

packages/onchainkit/src/OnchainKitProvider.tsx
```diff
@@ -61,6 +61,11 @@ export function OnchainKitProvider({
           display: config?.wallet?.display ?? 'classic',
           termsUrl: config?.wallet?.termsUrl || DEFAULT_TERMS_URL,
           privacyUrl: config?.wallet?.privacyUrl || DEFAULT_PRIVACY_URL,
+          supportedWallets: {
+            rabby: config?.wallet?.supportedWallets?.rabby ?? false,
+            trust: config?.wallet?.supportedWallets?.trust ?? false,
+            frame: config?.wallet?.supportedWallets?.frame ?? false,
+          },
```
alessey, line 68: since we already allow users to pass in their own wagmi config, can we just use that to generate the supported wallets?

cpcramer: I'm not sure if we can do that because we need to add the wallet's icon and i'm finding that connecting to certain wallets can be unreliable. For example, the Frame wallet implementation is slightly different that the Trust and Rabby implementation. 

Also, we want to avoid decision fatigue, so we purposely do not support every wallet by default.

dgca (approved):

packages/onchainkit/src/OnchainKitProvider.tsx
jxom: Why not support [EIP-6963](https://eips.ethereum.org/EIPS/eip-6963) which is what every "connect wallet" library uses (RainbowKit, ConnectKit, Thirdweb, Dynamic, Privy, etc)? Wagmi has this [built-in](https://wagmi.sh/react/api/createConfig#multiinjectedproviderdiscovery) and turned on by default – you just need to iterate through the `connectors` (they also have their icons encoded as base64).

cpcramer: @jxom TIL! This looks great, will look into supporting!

brendan-defi:
on very small screens (iphone se) the modal is a bit too tall when all connectors are enabled:
image

brendan-defi (approved):

PR #2179, opened 2025-03-27
chore: refactor Wallet

abcrane123:
**What changed? Why?**

- `WalletDropdown` effectively replaces `WalletAdvanced`, supporting both basic and advanced dropdown configurations based on the children provided
- deprecate WalletAdvanced
- refactor `WalletAdvancedProvider` logic into `WalletProvider` and remove `WalletAdvancedProvider`

```
<Wallet>
  <ConnectWallet />
  <WalletDropdown>
      <WalletAdvancedWalletActions />
      <WalletAdvancedAddressDetails />
      <WalletAdvancedTransactionActions />
      <WalletAdvancedTokenHoldings />
  </WalletDropdown>
</Wallet>
```

```
<Wallet>
  <ConnectWallet />
  <WalletDropdown>
    <WalletDropdownLink
        icon="wallet"
        key="wallet-dd-link"
        href="https://keys.coinbase.com"
        target="_blank"
      >
      Wallet
    </WalletDropdownLink>
    <WalletDropdownDisconnect key="wallet-dd-disconnect" />
  </WalletDropdown>
</Wallet>
```

```
<Wallet>
  <ConnectWallet />
  <WalletDropdown />
</Wallet>
```

**Notes to reviewers**

**How has it been tested?**


brendan-defi (commented):
@abcrane123 I'm sorry I missed this earlier, but with no fixed height on WalletAdvanced anymore, the various active screens for WalletAdvanced are different heights

"Homescreen":
image

QR:
image

Swap:
image

Send:
image


482px is the current implementation.

packages/onchainkit/src/wallet/components/WalletAdvancedSwap.tsx
```diff
@@ -61,7 +61,7 @@ export function WalletAdvancedSwap({
         isActiveFeatureClosing
           ? 'fade-out slide-out-to-right-5 animate-out fill-mode-forwards ease-in-out'
           : 'fade-in slide-in-from-right-5 linear animate-in duration-150',
-        'relative',
+        'relative w-88',
```
brendan-defi, line 64: add `h-120`?

brendan-defi (dismissed):

packages/onchainkit/src/wallet/components/WalletAdvancedAddressDetails.tsx
```diff
@@ -37,14 +35,14 @@ export function WalletAdvancedAddressDetails({
...
-    return <div className="mt-1 h-28 w-10" />; // Prevent layout shift
+    return <div className="mt-1 h-28 w-10 px-4 py-3" />; // Prevent layout shift
   }
 
   return (
     <div
       data-testid="ockWalletAdvanced_AddressDetails"
       className={cn(
-        'mt-2 flex flex-col items-center justify-center',
+        'mt-2 flex w-88 flex-col items-center justify-center px-4 py-3',
```
dgca, line 45: thought: I know we're using this width in a few places to grow the layout of the wallet dropdown container when people are using advanced features. It's usually `w-88 px-4 py-3`

It might be cleaner to make a layout component that just sets those things, and lets the caller override with its own classNames since I think you tweak the `py` in a couple of those

Pretty trivial, but if we did that, I'd recommend using `min-w-88` in case the container was bigger (it won't be, but might as well do things that make things more composable)

Not suggesting we do this in this PR, but something we could do later on

dgca (approved):

packages/onchainkit/src/wallet/components/WalletAdvancedAddressDetails.tsx
abcrane123: ok i can explore in followup !

dgca (approved):

brendan-defi (approved):

PR #2180, opened 2025-03-27
chore: Add Attestation caching

cpcramer:
**What changed? Why?**
Add caching to the `useAttestation` hook.

This saves us one network call on every render of the identity component - reducing network overhead and latency.

Removes:
Screenshot 2025-03-27 at 2 10 27 PM

**Notes to reviewers**

**How has it been tested?**
Tested in playground using logs and network tab.

### DIFF
--- a/packages/onchainkit/src/identity/hooks/useAttestations.ts
+++ b/packages/onchainkit/src/identity/hooks/useAttestations.ts
@@ -14,16 +14,10 @@ export function useAttestations(
   { address, chain, schemaId }: UseAttestations,
   queryOptions?: UseQueryOptions<Attestation[]>,
 ): Attestation[] {
-  if (!schemaId) {
-    return [];
-  }
-
-  const queryKey = ['useAttestations', address, chain?.id, schemaId];
-
   const result = useQuery<Attestation[]>({
-    queryKey,
+    queryKey: ['useAttestations', address, chain?.id, schemaId],
     queryFn: async () => {
-      return getAttestations(address, chain, { schemas: [schemaId] });
+      return getAttestations(address, chain, { schemas: [schemaId!] });
     },
     enabled: !!address && !!schemaId,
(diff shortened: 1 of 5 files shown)


packages/onchainkit/src/identity/hooks/useAttestations.ts
```diff
@@ -1,29 +1,35 @@
...
-  useEffect(() => {
-    const fetchData = async () => {
-      const foundAttestations = await getAttestations(address, chain, {
-        schemas: [schemaId],
-      });
-      setAttestations(foundAttestations);
-    };
-    fetchData();
-  }, [address, chain, schemaId]);
+  const queryKey = ['useAttestations', address, chain?.id, schemaId];
```
dgca, line 21: nit(non-blocking): Personal preference, but since this is just used once and declaring it as variable isn't really adding much in terms of readability, it could just be inlined

packages/onchainkit/src/identity/hooks/useAttestations.ts
```diff
@@ -1,29 +1,35 @@
...
 
-  return attestations;
+  const result = useQuery<Attestation[]>({
```
dgca, line 23: issue: Rules of hooks should be yelling at you about calling this conditionally (lines 17-19).

You should be able to remove that guard clause altogether since you're using the enabled option.

cpcramer: Good catch

packages/onchainkit/src/identity/hooks/useAttestations.ts
cpcramer: that's fair, updated!

dgca (dismissed):

dgca (approved):

brendan-defi (approved):

PR #2182, opened 2025-03-28
fix: useFundCardSetupOnrampEventListeners unmount reset

teimurjan:
**What changed? Why?**

Updated cleanup of useFundCardSetupOnrampEventListeners on unmount. Currently, if the provider unmounts within 3000ms after an action, state mutation happens on an unmounted provider.

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/packages/onchainkit/src/fund/hooks/useFundCardSetupOnrampEventListeners.ts
+++ b/packages/onchainkit/src/fund/hooks/useFundCardSetupOnrampEventListeners.ts
@@ -1,4 +1,4 @@
-import { useCallback, useEffect } from 'react';
+import { useCallback, useEffect, useRef } from 'react';
 import { useFundContext } from '../components/FundCardProvider';
 import { FUND_BUTTON_RESET_TIMEOUT } from '../constants';
 import type { EventMetadata, SuccessEventData } from '../types';
@@ -6,6 +6,13 @@ import { setupOnrampEventListeners } from '../utils/setupOnrampEventListeners';
 
 export const useFundCardSetupOnrampEventListeners = () => {
   const { setSubmitButtonState, updateLifecycleStatus } = useFundContext();
+  const timeoutRef = useRef<NodeJS.Timeout | null>(null);
+
+  const scheduleFundButtonReset = useCallback(() => {
+    timeoutRef.current = setTimeout(() => {
+      setSubmitButtonState('default');
+    }, FUND_BUTTON_RESET_TIMEOUT);
+  }, [setSubmitButtonState]);
 
   const handleOnrampEvent = useCallback(
     (data: EventMetadata) => {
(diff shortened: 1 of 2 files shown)


packages/onchainkit/src/fund/hooks/useFundCardSetupOnrampEventListeners.ts
```diff
@@ -1,11 +1,18 @@
...
 
 export const useFundCardSetupOnrampEventListeners = () => {
   const { setSubmitButtonState, updateLifecycleStatus } = useFundContext();
+  const timeoutRef = useRef<NodeJS.Timeout | null>(null);
+
+  const scheduleFundButtonReset = useCallback(() => {
+    timeoutRef.current = setTimeout(() => {
+      setSubmitButtonState('default');
+    }, FUND_BUTTON_RESET_TIMEOUT);
+  }, [setSubmitButtonState]);
```
dgca, line 15: I don't think this is likely to happen, but if `handleOnrampEvent` and `handleOnrampSuccess` are both called, and then the component unmounts, the cleanup function returned from `useEffect` will only clean up the latest timeout.

We could fix that by clearing the current timeout before setting a new one.

dgca (changes requested):
Ty for the PR! Definitely an issue that should be fixed. I think we might still the issue with the proposed changes in certain circumstances. Let me know what you think 👍

packages/onchainkit/src/fund/hooks/useFundCardSetupOnrampEventListeners.ts
```diff
@@ -21,9 +32,7 @@ export const useFundCardSetupOnrampEventListeners = () => {
         });
 
         setSubmitButtonState('error');
-        setTimeout(() => {
-          setSubmitButtonState('default');
-        }, FUND_BUTTON_RESET_TIMEOUT);
+        scheduleFundButtonReset();
```
dgca, line 35: issue: We gotta add this to the dependency array or the lint CI check will fail

teimurjan: Got it, done ✅

dgca:
@teimurjan Heyo, sorry for all the back and forth but now it looks like the test coverage is under 100%. We need it at 100% to merge. Btw, you can check that locally by doing `pnpm f:ock test:coverage`

teimurjan:
> @teimurjan Heyo, sorry for all the back and forth but now it looks like the test coverage is under 100%. We need it at 100% to merge. Btw, you can check that locally by doing `pnpm f:ock test:coverage`

I get it. I've just committed a fix with a new test checking the missing lines.

dgca (approved):

cb-heimdall:
Review Error for dgca @ 2025-05-28 21:03:59 UTC
User failed mfa authentication, public email is not set on your github profile. see [go/mfa-help](http://go/mfa-help)

dgca:
@teimurjan Thank you so much for bearing with us 🙏

PR #2186, opened 2025-03-28
chore: update ConnectWallet and add default children

abcrane123:
**What changed? Why?**
- implement default children for `<ConnectWallet />`
- refactor ConnectWallet logic
- introduce new prop `disconnectedLabel` on ConnectWallet that enables devs to customize disconnected state (children will represent connected state)
- deprecate `ConnectWalletText` component

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/packages/onchainkit/src/wallet/components/ConnectWallet.tsx
+++ b/packages/onchainkit/src/wallet/components/ConnectWallet.tsx
@@ -24,7 +24,7 @@ import { ConnectWalletText } from './ConnectWalletText';
 import { WalletModal } from './WalletModal';
 import { useWalletContext } from './WalletProvider';
 
-const connectWalletDefaultchildren = (
+const connectWalletDefaultChildren = (
   <>
     <Avatar className="h-6 w-6" />
     <Name />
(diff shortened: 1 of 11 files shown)


packages/onchainkit/src/wallet/components/ConnectWallet.tsx
```diff
@@ -23,13 +24,21 @@ import { ConnectWalletText } from './ConnectWalletText';
...
+    <Name />
+  </>
+);
+
 export function ConnectWallet({
   children,
   className,
   // In a few version we will officially deprecate this prop,
   // but for now we will keep it for backward compatibility.
   text = 'Connect Wallet',
```
abcrane123, line 39: i think we should remove this prop but that would be a breaking change. thoughts ?

cpcramer: I think I prefer this prop over the ConnectWalletText component...but curious what others think

packages/onchainkit/src/wallet/components/ConnectWallet.tsx
```diff
@@ -23,13 +24,21 @@ import { ConnectWalletText } from './ConnectWalletText';
 import { WalletModal } from './WalletModal';
 import { useWalletContext } from './WalletProvider';
 
+const connectWalletDefaultchildren = (
```
cpcramer, line 27: nit 
```suggestion
const connectWalletDefaultChildren = (
```

dgca (approved):

cpcramer (approved):

brendan-defi (approved):

PR #2187, opened 2025-03-28
fix: Changeset version script

dgca:
**What changed? Why?**

- Removed `--since=origin/main` flag in the changeset script which was causing it to not report a version
- Fall back to package.json if changeset doesn't produce a version change
- More robust error handling

**Notes to reviewers**

**How has it been tested?**

- Manually

### DIFF
--- a/packages/onchainkit/scripts/get-next-version.js
+++ b/packages/onchainkit/scripts/get-next-version.js
@@ -36,7 +36,7 @@ function getNextVersion() {
 
     if (!nextVersion) throw new Error('No onchainkit version found');
   } catch (error) {
-    console.error('Error checking changeset status:\n', error);
+    console.error('Error checking changeset status:\n', error.message);
   }
 
   if (!nextVersion) {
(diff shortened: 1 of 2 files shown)


cpcramer (approved):

cpcramer (approved):

packages/onchainkit/scripts/get-next-version.js
```diff
@@ -2,37 +2,59 @@ import { execSync } from 'child_process';
 import process from 'process';
 import fs from 'fs';
 import path from 'path';
+import { fileURLToPath } from 'url';
+import { dirname } from 'path';
```
brendan-defi, line 6: nit: can put this named import with the default import on line 4

dgca: Removed this and using `path.dirname` below for consistency.

brendan-defi (approved):

packages/onchainkit/scripts/get-next-version.js
```diff
@@ -2,37 +2,58 @@ import { execSync } from 'child_process';
...
+    });
 
     const lines = output.split('\n');
     const onchainkitVersionLine = lines.find((line) =>
       line.includes(packageName),
     );
 
-    if (!onchainkitVersionLine) return;
+    if (!onchainkitVersionLine)
+      throw new Error('No onchainkit version line found');
```
brendan-defi, line 32: would have expected linter to complain about no brackets, but as long as it's passing lint i'm fine with it

dgca: I kinda line one liners for these and the linter's fine with it, but if we wanna enforce brackets later we can turn that rule on and `--fix`

cpcramer (approved):

PR #2192, opened 2025-03-31
fix: Changesets PR title

dgca:
**What changed? Why?**

Fixes the issue that was making Changeset's PR title all wonky

**Notes to reviewers**

**How has it been tested?**

Manually

### DIFF
--- a/.changeset/common-places-enjoy.md
+++ b/.changeset/common-places-enjoy.md
@@ -1,5 +1,0 @@
----
-"@coinbase/onchainkit": patch
----
-
-Testing changeset PR title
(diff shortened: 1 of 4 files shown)


packages/onchainkit/scripts/get-next-version.js
```diff
@@ -33,6 +34,8 @@ function getNextVersion() {
 
     nextVersion = onchainkitVersionLine.split(packageName)[1].trim();
 
+    console.log('Version bump detected: ', nextVersion);
```
dschlabach, line 37: remove?

dschlabach (approved):

cpcramer (approved):

packages/onchainkit/scripts/get-next-version.js
dgca: I figured it’d be helpful to have that be available in the action logs for the release  action. Is it cool if we keep it?

PR #2198, opened 2025-03-31
🚧 chore: eslint enable exhaustive deps rule

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


packages/onchainkit/scripts/get-next-version.test.ts
```diff
@@ -9,6 +9,9 @@ vi.mock('child_process');
 vi.mock('fs');
 vi.mock('path');
 vi.mock('url');
+vi.mock('util', () => ({
+  stripVTControlCharacters: vi.fn((str) => str),
+}));
```
cpcramer, line 14: All of these tests were failing locally for me due to my specific nodejs version

packages/onchainkit/src/internal/components/QrCode/useLogo.tsx
```diff
@@ -60,7 +61,7 @@ export function useLogo({
             width={logoSize}
             height={logoSize}
             preserveAspectRatio="xMidYMid slice"
-            href={String(logo)}
+            href={String(transformedLogo)}
```
cpcramer, line 64: QR code logo still works

packages/onchainkit/src/minikit/MiniKitProvider.tsx
```diff
@@ -81,7 +81,7 @@ export function MiniKitProvider({
     return () => {
       sdk.removeAllListeners();
     };
-  }, []);
+  });
```
dgca, line 84: question: Is this intentional? This is gonna cause the hook to be called on every render and the event listeners to be cleaned up + reinitialized

Having a useMount might be handy. E.g.

```tsx
export function useMount(cb: () => void) {
  const isMounted = useRef(false);

  useEffect(() => {
    if (isMounted.current) return;
    isMounted.current = true;
    cb();

  }, [func]);
}
```

cpcramer: Reverted this change. For now, we will leave this as an empty dependency array and ignore the eslint warning. A follow up can be to refactor with `useMount`

dgca (dismissed):

packages/onchainkit/src/fund/hooks/usePaymentMethods.ts
```diff
@@ -70,5 +70,6 @@ export const usePaymentMethods = ({
   // biome-ignore lint/correctness/useExhaustiveDependencies: initial effect
```
alessey, line 70: have we removed biome?  can we ditch these comments?

cpcramer: Yep! I'll remove all Biome related comments in its own PR

packages/onchainkit/src/fund/hooks/useFundCardSetupOnrampEventListeners.ts
```diff
@@ -65,5 +65,5 @@ export const useFundCardSetupOnrampEventListeners = () => {
     return () => {
       unsubscribe();
     };
-  }, []);
+  });
```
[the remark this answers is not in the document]
cpcramer: Thanks, will revert and ignore linting

packages/onchainkit/src/fund/hooks/usePaymentMethods.ts
cpcramer: Removing Biome here https://github.com/coinbase/onchainkit/pull/2207

alessey (approved):

dgca (approved):

PR #2211, opened 2025-04-02
fix: Remove unused @coinbase/onchainkit/core export + enable alpha releases

dgca:
**What changed? Why?**

- Removes \@coinbase/onchainkit/core export from package.json that's not included in the build
- Removes --dry-run from alpha release script

**Notes to reviewers**

Figured this change would be a good minor but testable change to test alpha releases with.

**How has it been tested?**

Manually

### DIFF
--- a/.changeset/nice-spoons-kiss.md
+++ b/.changeset/nice-spoons-kiss.md
@@ -1,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
+
+Removes unused @coinbase/onchainkit/core export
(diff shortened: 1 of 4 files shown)


.changeset/nice-spoons-kiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
+
+Removes unused \@coinbase/onchainkit/core export
```
cpcramer, line 5: nit: The format we try to follow is:
Type (chore, feat, etc): description-past-tense. By @github-username #PR-number 

- **chore**: Removed unused \@coinbase/onchainkit/core export. By @dgca #2211.

dgca: Updated. I don't add the bullet point, right? Those are added by changesets?

brendan-defi (approved):

cpcramer (approved):

PR #2221, opened 2025-04-03
chore: Remove onchainkit version file that is now autogenerated

dgca:
**What changed? Why?**

Remove version file since that's now autogenerated

**Notes to reviewers**

**How has it been tested?**

Manually


packages/onchainkit/src/index.ts
```diff
@@ -7,7 +7,6 @@ export {
 } from './core/OnchainKitConfig';
 export { OnchainKitProvider } from './OnchainKitProvider';
 export { useOnchainKit } from './useOnchainKit';
-export { version } from './version';
```
cpcramer, line 10: Just thinking out loud, is there any benefit to keeping the version exported?

cpcramer (approved):

PR #2224, opened 2025-04-03
refactor: cli

alessey:
**What changed? Why?**
* refactor cli to only be the entry point into either the minikit or onchainkit flows

**Notes to reviewers**

**How has it been tested?**


packages/create-onchain/src/utils.ts
```diff
@@ -1,11 +1,41 @@
...
+    } else {
+      if (!excludeFiles.includes(entry.name)) {
+        await optimizedCopy(srcPath, destPath);
+      }
+    }
+  }
+}
+
+
+async function optimizedCopyFile(src: string, dest: string) {
```
dgca, line 34: nit(non-blocking): This wrapper function isn't doing anything, so we could just use fs.copyFile directly. I see it was like this before, though, so no biggie

packages/create-onchain/src/utils.ts
```diff
@@ -1,11 +1,41 @@
...
   await fs.copyFile(src, dest);
 }
 
-async function copyDir(src: string, dest: string) {
+async function optimizedCopyDir(src: string, dest: string) {
```
dgca, line 38: suggestion: I _think_ we could use `fs.cp(src, dir, { recursive: true })` and eliminate the need for this function. [See docs](https://nodejs.org/api/fs.html#fscpsrc-dest-options-callback)

dgca (approved):

abcrane123 (approved):
agree with Dan's comments but not blocking

cpcramer (approved):

PR #2227, opened 2025-04-04
refactor: Replace Packemon with Vite

dgca:
**What changed? Why?**

Removes Packemon and tsup in favor of Vite.

_**Why?**_

Currently, we use Packemon for building the production build, and tsup for building the development build.

Packemon uses Rollup under the hood for bundling, while tsup uses ESBuild. This was making it really hard to do things like adding a babel plugin since ESBuild doesn't support that out of the box.

Vite uses Rollup for bundling, and though ESBuild is faster than Rollup, I'm seeing that the build is actually faster due to the type generation step. Not by much, only a second or two.

I also did the following:

- Removed storybook
  - IMO, we should consider using Storybook, but with only a handful of stories and ongoing discussions about what docs and the V1 component API looks like, I figured it was a good time to cut it. We can bring it back in very easily if we want to adopt it more fully
  - Cleaned up some scripts. There were scripts that were doing the same thing, and some that weren't used.
- Renamed the `earn/mocks/index.ts` file to `mocks.test.ts`. Otherwise, it would get picked up by the build script.

**Notes to reviewers**

**How has it been tested?**

- Tests are passing
- Verified file generation manually


alessey (approved):
LGTM

cpcramer (approved):

PR #2243, opened 2025-04-08
Update _gitignore

Hack666r:
old link to a new one


dgca:
We appreciate the help, but we're not accepting PRs for comment updates, especially since the link redirects are working as intended.

PR #2244, opened 2025-04-08
Update optimism.ts

petryshkaCODE:
fix


dgca:
We appreciate the help, but we're not accepting PRs for comment updates, especially since the link redirects are working as intended.

PR #2245, opened 2025-04-08
feat: Add uninstalled wallet redirects

cpcramer:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


packages/onchainkit/src/wallet/components/WalletModal.tsx
```diff
@@ -146,11 +163,10 @@ export function WalletModal({
...
     try {
-      // Check if it's not Frame wallet
-      if (!window.ethereum.isFrame) {
-        throw new Error(
-          'Frame is not the active wallet. Please activate Frame before connecting.',
-        );
+      if (!window.ethereum?.isFrame) {
+        window.open('https://frame.sh/download', '_blank');
+        onClose();
+        return;
```
cpcramer, line 169: Note that Frame wallet must be handled differently than the other injected wallets

packages/onchainkit/src/wallet/utils/checkWalletAndRedirect.ts
```diff
@@ -0,0 +1,84 @@
...
+ */
+export interface WindowWithPhantom extends Window {
+  phantom?: {
+    ethereum?: {
+      isPhantom?: boolean;
+    };
+  };
+}
+
+export function isWalletInstalled(walletType: string): boolean {
```
abcrane123, line 15: will this work on mobile ?

cpcramer: No, this is web only

alessey: will these links deeplink on mobile if someone does have the wallet installed?  i.e. currently metamask will open the wallet app on mobile, will this continue to work?  phantom currently does not open even with the app, will this deeplink to the wallet?

cpcramer: Phantom still does not deeplink - It looks like we will need to add support for wallets where we are using the injected provider.

alessey: also, imo we should be respecting the users wagmi config to some extent, we currently ignore it completely.

alessey (approved):

abcrane123 (approved):

packages/onchainkit/src/wallet/utils/checkWalletAndRedirect.ts
cpcramer: Update: It seems like there's a lot of improvements that we can do for wallet connection (deep linking, 6963 support via mipd, automatically detect and support user wagmi config connectors) but these will need to be scoped and prioritized!

PR #2258, opened 2025-04-10
fix: only call useportfolio when subcomponents need it

alessey:
**What changed? Why?**
Move usePortfolio out of wallet context so its only called when its actually used

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/packages/onchainkit/src/wallet/components/WalletAdvancedAddressDetails.tsx
+++ b/packages/onchainkit/src/wallet/components/WalletAdvancedAddressDetails.tsx
@@ -6,6 +6,9 @@ import { zIndex } from '@/styles/constants';
 import { border, cn, color, pressable, text } from '@/styles/theme';
 import { useCallback, useState } from 'react';
 import { useWalletContext } from './WalletProvider';
+import { usePortfolio } from '../hooks/usePortfolio';
+import { RequestContext } from '@/core/network/constants';
+import { useAccount } from 'wagmi';
 
 type WalletAdvancedAddressDetailsProps = {
   classNames?: {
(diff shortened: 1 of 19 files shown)


packages/onchainkit/src/wallet/hooks/usePortfolio.ts
```diff
@@ -5,12 +5,17 @@ import { isApiError } from '@/internal/utils/isApiResponseError';
...
+  enabled?: boolean;
+};
+
 /**
  * Retrieves the portfolio for the provided address
  * portfolio includes the address, the balance of the address in USD, and the tokens in the address
  */
 export function usePortfolio(
-  { address }: { address: Address | undefined | null },
+  { address, enabled = true }: UsePortfolioProps,
```
dgca, line 18: suggestion(non-blocking): Personal preference, but for cases like this, I'd skip the enabled argument, and pass `null` to address if I want to disable the fetch.

dgca (approved):

packages/onchainkit/src/wallet/hooks/usePortfolio.ts
alessey: I think it would be nice to export this hook and allow users to pass through all tanstack params, I didn't want to make that change in this PR, but this is kinda a start to that.

packages/onchainkit/src/wallet/components/WalletAdvancedAddressDetails.tsx
```diff
@@ -93,7 +96,12 @@ export function WalletAdvancedAddressDetails({
 }
 
 function AddressBalanceInFiat({ className }: { className?: string }) {
-  const { portfolioFiatValue, isFetchingPortfolioData } = useWalletContext();
+  const { address } = useAccount();
```
brendan-defi, line 99: why use useAccount here vs. useWalletContext?

PR #2261, opened 2025-04-10
chore: add context exports

abcrane123:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/packages/onchainkit/src/checkout/index.ts
+++ b/packages/onchainkit/src/checkout/index.ts
@@ -7,7 +7,4 @@ export type {
   CheckoutReact,
   CheckoutStatusReact,
 } from './types';
-export {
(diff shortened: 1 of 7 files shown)


packages/onchainkit/src/checkout/index.ts
```diff
@@ -7,3 +7,7 @@ export type {
   CheckoutReact,
   CheckoutStatusReact,
 } from './types';
+export {
+  CheckoutProvider,
```
alessey, line 11: why export the providers?  isn't that basically what the base component export does?

abcrane123: removed

alessey:
this is a breaking change now that you are removing those

PR #2263, opened 2025-04-10
chore: improve TypeScript typing for Providers Component

mdqst:
**What changed? Why?**

I refactored the typing for the `Providers` component. Instead of specifying `{ children: ReactNode }` directly, I used `React.PropsWithChildren<{}>`. This improves compatibility with React and provides better type inference, making it easier to extend the component with additional props in the future.

**Notes to reviewers**

This change should ensure the component is more flexible without introducing any breaking changes. Please check if the updated type works well with the rest of the codebase and doesn’t cause any issues with children handling.

**How has it been tested?**

I ran the component in the existing environment and confirmed that no type errors were introduced and that it works seamlessly with the current usage of the `Providers` component.


dgca:
Appreciate you contributing, but unfortunately we don't accept PRs which are purely stylistic

PR #2264, opened 2025-04-10
chore: Add bidirectional resolution validation when batching Basenames and ENS names

cpcramer:
**What changed? Why?**
This PR fixes a vulnerability in the identity resolution system that allowed users to spoof their displayed names.

The current implementation of reverse resolution (address → name) didn't verify name ownership:

- It's possible to spoof the address by setting a name on the reverse registry without actually owning the name
- We would display that name without verification

The Solution: Bidirectional Validation

- Reverse Resolution: Get name from address (getName)
- Forward Resolution: Resolve name back to address (getAddress)
- Validation: Only display the name if addresses match

This prevents malicious actors from falsely displaying names they don't control, ensuring all displayed names are properly verified and legitimate.

**Notes to reviewers**
This is a follow up PR from [chore: Add bidirectional resolution validation to Basenames and ENS names](https://github.com/coinbase/onchainkit/pull/2255#top)

**How has it been tested?**
Unit tests.

### DIFF
--- a/packages/onchainkit/src/identity/utils/getNames.ts
+++ b/packages/onchainkit/src/identity/utils/getNames.ts
@@ -1,17 +1,19 @@
 import type { Basename, GetNameReturnType, GetNames } from '@/identity/types';
-import { mainnet } from 'viem/chains';
+import { base, mainnet } from 'viem/chains';
 import { getChainPublicClient } from '../../core/network/getChainPublicClient';
 import { isBase } from '../../core/utils/isBase';
 import { isEthereum } from '../../core/utils/isEthereum';
 import L2ResolverAbi from '../abis/L2ResolverAbi';
 import { RESOLVER_ADDRESSES_BY_CHAIN_ID } from '../constants';
 import { convertReverseNodeToBytes } from './convertReverseNodeToBytes';
+import { getAddress } from './getAddress';
 
 /**
  * An asynchronous function to fetch multiple Basenames or Ethereum Name Service (ENS)
  * names for a given array of Ethereum addresses in a single batch request.
  * It returns an array of ENS names in the same order as the input addresses.
  */
+// eslint-disable-next-line complexity
 export const getNames = async ({
   addresses,
   chain = mainnet,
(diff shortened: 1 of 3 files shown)


packages/onchainkit/src/identity/utils/getNames.ts
```diff
@@ -48,11 +50,31 @@ export const getNames = async ({
...
+        const result = batchResults[index];
         if (result.status === 'success' && result.result) {
-          results[index] = result.result as Basename;
+          const basename = result.result as Basename;
+          try {
+            // Verify basename with forward resolution
+            const resolvedAddress = await getAddress({
+              name: basename,
+              chain: base,
+            });
```
alessey, line 62: can we verify in a multicall too?

cpcramer: Yep! I would need to add `getAddresses` and `useAddress` first though - this would make more sense in a separate PR

packages/onchainkit/src/identity/utils/getNames.ts
```diff
@@ -48,11 +50,31 @@ export const getNames = async ({
...
+            }
+          } catch (error) {
+            console.error(
+              `Error during basename forward resolution verification for ${addresses[index]}:`,
+              error,
+            );
+          }
         }
-      });
+      }
```
dgca, line 77: suggestion(non-blocking): Is it worth parallelizing these async calls? In a for loop, they'll have to run one after the other, we could map to an array of promises and Promise.all to run them in parallel.  Same thought for below.

dgca (approved):

alessey (approved):

packages/onchainkit/src/identity/utils/getNames.ts
cpcramer: Yes I think so. I could add support for getAddresses and useAddresses which would accomplish this for forward resolution then integrate that into useNames. It also would be a nice utility function that we could export.

PR #2277, opened 2025-04-11
feat: Add batch ENS / Basename resolution with useAddresses hook and getAddresses utility

cpcramer:
**What changed? Why?**
Added new multicall ENS resolution utilities with `useAddresses` hook and `getAddresses` utility function to efficiently resolve multiple ENS or Basenames to Ethereum addresses, reducing network requests and improving performance for batch resolution scenarios.

**Notes to reviewers**

Follow up from [chore: Add bidirectional resolution validation when batching Basenames and ENS names](https://github.com/coinbase/onchainkit/pull/2264#top). As an additional fast follow, I will be implementing this function into `getNames`.

**How has it been tested?**
Unit tests.

### DIFF
--- a/packages/onchainkit/src/identity/hooks/useAddresses.ts
+++ b/packages/onchainkit/src/identity/hooks/useAddresses.ts
@@ -1,0 +1,31 @@
+import { getAddresses } from '@/identity/utils/getAddresses';
+import { DEFAULT_QUERY_OPTIONS } from '@/internal/constants';
+import { useQuery } from '@tanstack/react-query';
+import { mainnet } from 'viem/chains';
+import type {
+  GetAddressReturnType,
+  UseAddressesOptions,
+  UseQueryOptions,
+} from '../types';
+
+/**
+ * A React hook that leverages the `@tanstack/react-query` for fetching and optionally caching
+ * multiple Ethereum addresses from ENS names or Basenames in a single batch request.
+ */
+export const useAddresses = (
+  { names, chain = mainnet }: UseAddressesOptions,
+  queryOptions?: UseQueryOptions<GetAddressReturnType[]>,
+) => {
+  const namesKey = names.join(',');
+  const queryKey = ['useAddresses', namesKey, chain.id];
+
+  return useQuery<GetAddressReturnType[]>({
+    queryKey,
+    queryFn: () => getAddresses({ names, chain }),
+    enabled: !!names.length,
+    ...DEFAULT_QUERY_OPTIONS,
+    // Use cacheTime as gcTime for backward compatibility
+    gcTime: queryOptions?.cacheTime,
+    ...queryOptions,
+  });
+};
--- a/packages/onchainkit/src/identity/utils/getAddresses.ts
+++ b/packages/onchainkit/src/identity/utils/getAddresses.ts
@@ -1,0 +1,74 @@
+import { getChainPublicClient } from '@/core/network/getChainPublicClient';
+import { isBase } from '@/core/utils/isBase';
+import { isEthereum } from '@/core/utils/isEthereum';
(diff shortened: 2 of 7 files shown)


packages/onchainkit/src/identity/hooks/useAddresses.ts
```diff
@@ -0,0 +1,31 @@
...
+  queryOptions?: UseQueryOptions<GetAddressReturnType[]>,
+) => {
+  const namesKey = names.join(',');
+  const queryKey = ['useAddresses', namesKey, chain.id];
+
+  return useQuery<GetAddressReturnType[]>({
+    queryKey,
+    queryFn: () => getAddresses({ names, chain }),
+    enabled: !!names.length,
+    ...DEFAULT_QUERY_OPTIONS,
```
dgca, line 26: nit: I'd recommend always putting defaults at the top of the object, otherwise if someone adds a default that conflicts with the keys above (unlikely in this case), things won't work as expected.

E.g. if someone added `enabled: true` to the default options, it'll take precedence over `enabled: !!names.length`

packages/onchainkit/src/identity/hooks/useAddresses.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export const useAddresses = (
+  { names, chain = mainnet }: UseAddressesOptions,
+  queryOptions?: UseQueryOptions<GetAddressReturnType[]>,
+) => {
+  const namesKey = names.join(',');
+  const queryKey = ['useAddresses', namesKey, chain.id];
+
+  return useQuery<GetAddressReturnType[]>({
+    queryKey,
```
dgca, line 23: suggestion(non-blocking): 100% personal preference, but if you wanna save some variable declaraions...

```tsx
queryKey: ['useAddresses', ...names, chain.id],
```

packages/onchainkit/src/identity/utils/getAddresses.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  const client = getChainPublicClient(chain);
+  const results: GetAddressReturnType[] = Array(names.length).fill(null);
+
+  try {
+    // Filter out null or undefined names
+    const validItems = names
+      .map((name, index) => (name ? { name, index } : null))
+      .filter((item): item is { name: string; index: number } => item !== null);
```
dgca, line 38: suggestion(non-blocking): Filtering before mapping lets you skip some operations. Super trivial in terms of perf gains in this case, but in some cases it can be meaningful

dgca: Nvm, I see why this is the way it is. The user can pass "" as a name and you wanna preserve index order.

In that case, you could go with something like this if you wanna be more succinct (this would replace lines 32-73)

```
  const addressPromises = names.map(async (name) => {
    if (!name) return null;

    try {
      const address = await client.getEnsAddress({
        name,
        universalResolverAddress: isBasename(name)
          ? RESOLVER_ADDRESSES_BY_CHAIN_ID[chain.id]
          : undefined,
      });
      return address;
    } catch (error) {
      console.error(`Error resolving address for ${name}:`, error);
      return null;
    }
  });

  try {
    return await Promise.all(addressPromises);
  } catch (err) {
    console.error('Error resolving addresses in batch:', err);
    return Array(names.length).fill(null);
  }
```

dgca (approved):

brendan-defi (approved):

brendan-defi:
nice, love this

packages/onchainkit/src/identity/utils/getAddresses.ts
cpcramer: Saw this after merging, will look at refactoring!

PR #2281, opened 2025-04-14
chore: Refactor getNames utility to leverage batch processing with getAddresses

cpcramer:
**What changed? Why?**

Refactors the `getNames` function to use the `getAddresses` utility for batch name verification, replacing individual address verifications with more efficient batch processing. 

The change improves performance by reducing network calls when resolving multiple names.

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/packages/onchainkit/src/identity/utils/getNames.ts
+++ b/packages/onchainkit/src/identity/utils/getNames.ts
@@ -6,7 +6,7 @@ import { isEthereum } from '../../core/utils/isEthereum';
 import L2ResolverAbi from '../abis/L2ResolverAbi';
 import { RESOLVER_ADDRESSES_BY_CHAIN_ID } from '../constants';
 import { convertReverseNodeToBytes } from './convertReverseNodeToBytes';
-import { getAddress } from './getAddress';
+import { getAddresses } from './getAddresses';
 
 /**
  * An asynchronous function to fetch multiple Basenames or Ethereum Name Service (ENS)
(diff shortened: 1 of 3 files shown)


packages/onchainkit/src/identity/utils/getNames.ts
```diff
@@ -50,29 +50,46 @@ export const getNames = async ({
...
         if (result.status === 'success' && result.result) {
           const basename = result.result as Basename;
-          try {
-            // Verify basename with forward resolution
-            const resolvedAddress = await getAddress({
-              name: basename,
-              chain: base,
-            });
+          basenamesWithIndices.push({ basename, index });
+        }
```
alessey, line 62: could create your separate basenames array for getAddresses here since you are already looping over them

packages/onchainkit/src/identity/utils/getNames.ts
```diff
@@ -50,29 +50,46 @@ export const getNames = async ({
...
+          );
+          const resolvedAddresses = await getAddresses({
+            names: basenames,
+            chain: base,
+          });
+
+          // Update results with validated basenames
+          for (let i = 0; i < basenamesWithIndices.length; i++) {
+            const { basename, index } = basenamesWithIndices[i];
+            const resolvedAddress = resolvedAddresses[i];
```
alessey, line 79: is this index match dependable because getAddresses returns all resolved or all null?

packages/onchainkit/src/identity/utils/getNames.ts
```diff
@@ -50,29 +50,46 @@ export const getNames = async ({
...
+        }
+      }
+
+      if (basenamesWithIndices.length > 0) {
+        try {
+          // Verify basenames with forward resolution using batch processing
+          const basenames = basenamesWithIndices.map(
+            ({ basename }) => basename,
+          );
+          const resolvedAddresses = await getAddresses({
```
alessey, line 71: this function doesn't look to be batching

cpcramer: We are passing in an array of Basenames and `getAddresses` is using `Promise.all` to parallelize the multiple individual `getENSAddress` requests

packages/onchainkit/src/identity/utils/getNames.ts
```diff
@@ -50,29 +50,46 @@ export const getNames = async ({
...
+            names: basenames,
+            chain: base,
+          });
```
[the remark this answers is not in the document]
cpcramer: No. Each individual promise has its own .catch() handler.

Failed resolutions return `{ index, address: null }`, they do not cause the entire Promise.all to reject

packages/onchainkit/src/identity/utils/getNames.ts
```diff
@@ -50,29 +50,46 @@ export const getNames = async ({
...
+
+          // Update results with validated basenames
+          for (let i = 0; i < basenamesWithIndices.length; i++) {
```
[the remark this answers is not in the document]
cpcramer: Correct - The loop prevents spoofing or invalid name resolutions when getAddresses successfully resolves at least one address

alessey (approved):

PR #2283, opened 2025-04-14
refactor: send display name

brendan-defi:
**What changed? Why?**
* refactored `RecipientState` type to include all state related to recipient:
  * phase
  * user input
  * address — valid EVM address
  * displayValue — what we show after selection
* created `useRecipientState` to colocate all state and state updating related to `RecipientState`
  * integrated in `SendProvider`
  * removed `SendAddressSelection` as the business logic now resides in the hook, and the aggregating component was no longer performing useful work
  * `SendAddressInput` and `SendAddressSelector` now integrated directly into `Send`

**Notes to reviewers**
behavior change from prod:
  * if you input an address, and the address has a basename / ens name, when you select the address, the input field should have the name, not the address

example: `0x8038FAF7Ca8B1A09dcf1B91f6f807F99e26C759e`
before:
image

after:
image

**How has it been tested?**
* playground, test coverage


cpcramer:
Not blocking - could be a fast follow: 

I think

cpcramer (approved):

PR #2294, opened 2025-04-17
fix: Updates playground to remove outdated imports

dgca:
**What changed? Why?**

We had some type imports that were still using imports from before the monorepo migration. This removes those imports.

**Notes to reviewers**

I went with quick and dirty since we're starting to land on what v1 looks like. IMO, this is a good example of more types that should be exported. Personally, I'd err on the side of exporting all internal types.

**How has it been tested?**

Manually

### DIFF
--- a/packages/playground/components/demo/FundCard.tsx
+++ b/packages/playground/components/demo/FundCard.tsx
@@ -1,6 +1,6 @@
 import { FundCard } from '@coinbase/onchainkit/fund';
 
-export type PresetAmountInputs = readonly [string, string, string];
+type PresetAmountInputs = readonly [string, string, string];
 
 export default function FundCardDemo() {
   const presetAmountInputs: PresetAmountInputs = ['10', '20', '100'];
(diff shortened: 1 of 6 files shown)


packages/playground/components/demo/Transaction.tsx
```diff
@@ -3,8 +3,7 @@ import {
   clickContracts,
   heterogeneousClickCalls,
 } from '@/lib/transactions';
-import type { Call } from '@/onchainkit/esm/transaction/types';
-import type { LifecycleStatus } from '@/onchainkit/src/transaction';
+import type { LifecycleStatus } from '@coinbase/onchainkit/checkout';
```
brendan-defi, line 6: why checkout's `LifecycleStatus`?

packages/playground/components/demo/FundCard.tsx
```diff
@@ -1,5 +1,6 @@
 import { FundCard } from '@coinbase/onchainkit/fund';
-import type { PresetAmountInputs } from '../../onchainkit/esm/fund/types';
+
+export type PresetAmountInputs = readonly [string, string, string];
```
brendan-defi, line 3: why would we export from the demo component?

brendan-defi (approved):

cpcramer (approved):

PR #2307, opened 2025-04-22
feat: Support CLI option to setup ai editor rules

xtbase:
**What changed? Why?**

- Adding prompt for CLI to provide some default rules for popular ai powered editors

https://github.com/user-attachments/assets/6a721ffc-c8fd-416d-82d6-d44951850b83

**Notes to reviewers**

**How has it been tested?**

`pnpm test`


packages/create-onchain/rules/onchainkit.copilot.txt
```diff
@@ -0,0 +1,79 @@
+# Code Style & TypeScript
```
dgca, line 1: suggestion: Could we consolidate the three txt files into one and set the name of the file based on the user's selection?

xtbase: yea we could, the reason i split it up was in case each one had different optimizations, for example copilot/windsurf have a single rules file that contains overall context whereas cursor you could have multiple files. Lemme know if you have a strong preference for a single file

dgca: IMO, since we currently only have one file per model, it'd be better to just have one main source of content. It'll make reviewing easier too. There's some things that I think we should reconsider (e.g. use english seems unnecessary since this is for consumers, not the library itself) and reviewing that will be easier if we just have one source of truth.

xtbase:
@dgca updated ty

packages/create-onchain/src/utils.ts
```diff
@@ -108,3 +108,14 @@ export function detectPackageManager(): string {
   }
   return 'npm'; // default to npm if unable to detect
 }
+
+export async function addToGitignore({
```
dgca, line 112: question: Does it make sense to gitignore this stuff? I feel like I would want it checked in so things are more consistent, and in case I have to check out the project again.

I noticed windsurf mentions gitignoring, but neither Cursor or Github do. IMO, there's nothing wrong with checking the file in and if the user wants to gitignore, they can do so manually.

packages/create-onchain/rules/onchainkit.txt
```diff
@@ -0,0 +1,65 @@
...
+  - Manage swap transaction states correctly
+
+- Frame Components:
+  - Use FrameMetadata for proper frame configuration
+  - Handle frame messages and validation correctly
+  - Implement proper frame response handling
+  - Follow frame security best practices
+
+Best Practices
+- Always wrap components with OnchainKitProvider
```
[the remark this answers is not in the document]
xtbase: This is pulled from current rule 

https://cursor.directory/onchainkit

we should update this if we want more accurate rules

packages/create-onchain/rules/onchainkit.txt
```diff
@@ -0,0 +1,65 @@
...
+
+Error Handling
+- Implement proper error boundaries
+- Handle API errors gracefully
+- Provide user-friendly error messages
+- Use proper TypeScript error types
+- Handle edge cases appropriately
+
+Key Conventions
+1. Always use OnchainKitProvider at the app root
```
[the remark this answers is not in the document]
xtbase: same as above

packages/create-onchain/src/utils.ts
xtbase: yea i don't have strong feelings on gitignoring it, figured some people might want to customize but that's probably in the small minority

xtbase: removing it for now unless we see people complaining

xtbase:
Updated @dgca I think regarding the exact rules we can even pull that from some remote place later on so it's more dynamic , this should be fine for now wdyt

packages/create-onchain/rules/onchainkit.txt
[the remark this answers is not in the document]
dgca: In that case, I think updating those rules should happen before we merge this PR.

I don't think we should include rules that tell OnchainKit to do things incorrectly.

Alternatively, we can make the content in the codebase the source of truth, and update https://cursor.directory/onchainkit afterwards. I don't have a preference on order of operations, but I think we should figure out how we wanna do that before merging this PR

dgca:
Gonna close this out and we can revisit when we're at v1

PR #2311, opened 2025-04-23
chore: refactor fund components

abcrane123:
**What changed? Why?**
- add render prop to FundButton and remove display props 
- add children prop to FundButton (mutually exclusive with render prop)
- remove `text`, `successText`, `errorText`, `hideIcon`, `hideText` props

```js

<FundCardSubmitButton render={customRender} />

<FundButton render={customRender} />


```

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/packages/onchainkit/src/fund/components/FundButton.tsx
+++ b/packages/onchainkit/src/fund/components/FundButton.tsx
@@ -17,7 +17,7 @@ import { SuccessSvg } from '../../internal/svg/successSvg';
 import { background } from '../../styles/theme';
 import { ConnectWallet } from '../../wallet/components/ConnectWallet';
 import { useGetFundingUrl } from '../hooks/useGetFundingUrl';
-import type { FundButtonReact } from '../types';
+import type { FundButtonProps } from '../types';
 import { getFundingPopupSize } from '../utils/getFundingPopupSize';
 
 export function FundButton({
@@ -33,7 +33,7 @@ export function FundButton({
   onPopupClose,
   onClick,
   render,
-}: FundButtonReact) {
+}: FundButtonProps) {
   const componentTheme = useTheme();
   // If the fundingUrl prop is undefined, fallback to our recommended funding URL based on the wallet type
   const fallbackFundingUrl = useGetFundingUrl({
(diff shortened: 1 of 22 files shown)


packages/onchainkit/src/fund/types.ts
```diff
@@ -111,32 +112,74 @@ type GetOnrampBuyUrlOptionalProps = {
...
+  fiatCurrency?: string;
+  /* A callback function that will be called when the popup window is closed */
+  onPopupClose?: () => void;
+  /* A callback function that will be called when the button is clicked */
+  onClick?: () => void;
+};
+
+export type FundButtonRenderParams = {
+  /* The state of the button component, only relevant when using FundCardSubmitButton */
+  state: FundButtonStateReact;
```
abcrane123, line 146: thinking this should be status instead of state

packages/onchainkit/src/fund/components/FundButton.tsx
```diff
@@ -168,33 +169,23 @@ export function FundButton({
...
-  }
+  }, [buttonState, buttonIcon, buttonTextContent]);
 
   if (shouldShowConnectWallet) {
     return <ConnectWallet className={cn('w-full', className)} />;
   }
 
+  if (render) {
+    return render({
+      state: buttonState,
```
dgca, line 183: follow-up: I think this is named `status` in the Transaction PR. Assuming it's more or less doing the same thing, we should standardize that across components

dgca: I think I prefer status, but no strong feelings

packages/onchainkit/src/fund/components/FundCardSubmitButton.tsx
```diff
@@ -34,17 +40,79 @@ export function FundCardSubmitButton() {
...
+      </>
+    );
+  }, [submitButtonState, buttonIcon, buttonTextContent]);
+
+  // FundButton accepts render prop or children but not both
+  if (render) {
+    return (
+      <FundButton
+        disabled={isButtonDisabled}
+        className="w-full"
```
dgca, line 94: nit: Since this is passing render down to FundButton, this className won't get rendered

packages/onchainkit/src/fund/types.ts
dgca: +1

dgca (approved):

PR #2313, opened 2025-04-24
feat: upgrade tailwind to v4

xtbase:
**What changed? Why?**

- Upgrade tailwind version in the following places to v4
  - onchainkit
  - onchainkit next template
  - minikit-basic template
  - minikit-snake template
  - playground

- Upgrade to @tailwindcss/postcss in the following places
  - onchainkit
  - onchainkit next template
  - minikit-basic template
  - minikit-snake template
  - playground

- Clean up css files to adhere to new rules
  - onchainkit
  - onchainkit next template
  - minikit-basic template
  - minikit-snake template
  - playground 

- Update playground shadcn to use v4

- Deprecate tailwind.css export as it was just exporting legacy ock classes which will be deprecated in the future

**Notes to reviewers**

**How has it been tested?**

- Test playground - preview link: https://onchainkit-playground-git-tea-tailwind-v4-coinbase-vercel.vercel.app/
- Test onchainkit creation
- Test minikit creation

Steps to test onchainkit creation

- go to onchainkit cli folder
- run `pnpm build && node ./dist/esm/cli.js`
- cd into your project
- edit package.json by changing `"@coinbase/onchainkit": "latest"` to `"@coinbase/onchainkit": "file:../../onchainkit"`
- `npm install && npm run dev` in your project
- check to make sure onchainkit components are good


packages/create-onchain/templates/minikit-basic/app/globals.css
```diff
@@ -1,25 +1,32 @@
-@tailwind base;
-@tailwind components;
-@tailwind utilities;
+@import "tailwindcss";
+@custom-variant dark (&:where(.dark, .dark *));
```
dgca, line 2: follow-up: We should consider scoping this or using data-attributes for this. This technically leaks OnchainKit internals to the parent, which we should avoid if we can figure out a good way to do so

packages/onchainkit/src/styles/tailwind-base.css
```diff
@@ -1,4 +1,25 @@
-@tailwind base;
+@import 'tailwindcss';
+@custom-variant dark (&:where(.dark, .dark *));
```
dgca, line 2: follow-up: We should consider scoping this or using data-attributes for this. This technically leaks OnchainKit internals to the parent, which we should avoid if we can figure out a good way to do so

packages/create-onchain/templates/next/app/page.tsx
```diff
@@ -31,16 +32,32 @@ const components = [
...
-  { name: 'Commerce', url: 'https://github.com/coinbase/onchain-commerce-template'},
+  {
+    name: 'Commerce',
+    url: 'https://github.com/coinbase/onchain-commerce-template',
+  },
   { name: 'Fund', url: 'https://github.com/fakepixels/fund-component' },
 ];
 
 export default function App() {
+  const [theme, setTheme] = useState<'light' | 'dark'>('light');
```
dgca, line 43: issue: This will reset the theme to light whenever the page is refreshed. We could do something similar to what [Tailwind suggests here](https://tailwindcss.com/docs/dark-mode#with-system-theme-support) to default to the user's preferred mode

packages/onchainkit/src/styles/tailwind-base.css
xtbase: lemme add a task for this as i'm going through the ock classes deprecation

packages/create-onchain/templates/next/app/page.tsx
dgca: I feel like it'd be better not to have the toggle than to have it not work properly. Can we just omit it altogether?

xtbase: yea that's fine

dgca (approved):

PR #2317, opened 2025-04-26
feat: <Connected /> component

dgca:
**What changed? Why?**

Adds <Connected /> component that only renders children if user has connected their account.

**Notes to reviewers**

**How has it been tested?**

- Manually
- Unit tests


packages/onchainkit/src/connected/components/Connected.tsx
```diff
@@ -0,0 +1,27 @@
...
+  children: ReactNode;
+  /** The component to render when the user is not connected. Defaults to <ConnectWallet />. */
+  fallback?: ReactNode;
+};
+
+/**
+ * Renders children only when the user has connected their wallet.
+ * Otherwise, renders the fallback.
+ */
+export function Connected({
```
xtbase, line 16: do we want to add support for the loading states (isConnecting/isReconnecting)? It might promote overall better handling of components that require connectedness to force developers to think about these states

dgca: I started doing this but after looking at [the docs](https://wagmi.sh/react/api/hooks/useAccount#status), it looks like the connecting and reconnecting status can be active even when you're already connected.

I.e. you can have an address and be connecting. I'm not sure what people would expect when we get into one of the states where you have an account connected, but you're not in the connected state.

To make this a bit easier to think about, I updated it so that we render `children` if there's an available address from `useAccount()`. If there's no address but `isConnecting` is true, we render the `connecting` content. Otherwise we render the fallback. Not doing anything with `isReconnecting` ATM cause it feels like an edge case and we could add it later.

Does that make sense? I can change it more but honestly I'm starting to lose the plot. Not sure if this is better than just having the consumer use `useAccount` directly.

dschlabach (approved):

brendan-defi (approved):

xtbase (approved):

Issue #2309, opened 2025-04-23
Bug: `crypto.randomUUID` is not a function


JFrankfurt:
### Describe the bug and the steps to reproduce it

![Image](https://github.com/user-attachments/assets/ebea0cd1-076d-4cec-9725-ab5af0f9b31e)

`crypto.randomUUID is not a function`

This occurs when accessing the dev build on a remote device on your local network (i.e., for testing mobile). These connections are made over http, and the web crypto api requires an https connection

### What's the expected behavior?

I'd expect a wrapper that checks for `crypto` availability and uses `Math.random` for `randomUUID` when it isn't there.

### What version of the libraries are you using?

`latest`

xtbase:
should be fixed in latest version

PR #2318, opened 2025-04-28
fix: support an insecure fallback for session id to allow for testing on http contexts

xtbase:
**What changed? Why?**

- Added a fallback for non-critical uuid usages to allow for testing in http environments

See issue https://github.com/coinbase/onchainkit/issues/2309

**Notes to reviewers**

Did not introduce `uuid` dependency to avoid bloating package/security concerns (though extremely unlikely), but it's a consideration we can make

**How has it been tested?**

### DIFF
--- a/packages/onchainkit/src/utils/crypto.ts
+++ b/packages/onchainkit/src/utils/crypto.ts
@@ -1,0 +1,20 @@
+/**
+ * Generates a UUID using the crypto API if available, otherwise falls back to a
+ * more insecure method. Only use this for non-critical purposes.
+ *
+ * @returns string - A UUID string that may not be cryptographically secure
+ */
+export const generateUUIDWithInsecureFallback = () => {
+  if (
+    typeof crypto !== 'undefined' &&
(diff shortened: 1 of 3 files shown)


packages/onchainkit/src/utils/crypto.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * Generates a UUID using the crypto API if available, otherwise falls back to a
+ * more insecure method. Only use this for non-critical purposes.
+ *
+ * @returns string - A UUID string that may not be cryptographically secure
```
dgca, line 5: nit: The type checker doesn't check JSDoc comments, so it's better to omit types in comments and add them to the function itself. You can just rely on type inference, in this case, but up to you

xtbase: sg i'll update this along with a test

dgca (approved):

brendan-defi (approved):

PR #2326, opened 2025-04-30
chore: Don't ask to generate manifest during app creation in create-onchain

dgca:
**What changed? Why?**

Skips the manifest prompt during the create-onchain process. This also adds messaging that's more explicit about what the account manifest is and what to do in terms of steps before going live.

We were getting feedback that it was confusing to users since they likely don't have a domain set up at that point.

Also gets rid of the timer and has the user explicitly close the account manifest window. When I was going through that process, I didn't have enough time to read the whole thing before it closed and I felt like something might have broken without me knowing.

**Notes to reviewers**

**How has it been tested?**

- Manually
- Unit tests


abcrane123 (approved):
nice

brendan-defi (approved):

PR #2360, opened 2025-05-13
v1: Signature audit pt 1

dgca:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**


packages/onchainkit/package.json
```diff
@@ -40,6 +40,7 @@
     "graphql-request": "^6.1.0",
     "qrcode": "^1.5.4",
     "tailwind-merge": "^2.3.0",
+    "usehooks-ts": "^3.1.1",
```
dgca, line 43: The current implementation of `useIsMounted` in `src/internal/hooks/useIsMounted` doesn't let you check if the component is mounted at the end of the component lifecycle.

Rather than update the hook and its usage internally, I figured bringing in this library would be better since it also covers a _lot_ of the hooks we have in our codebase, and it's been extremely battle tested.

packages/onchainkit/src/internal/types.ts
```diff
@@ -66,3 +67,31 @@ export type LifecycleStatusUpdate<T extends AbstractLifecycleStatus> =
...
+
+/**
+ * Utility for typing component props that support the render prop pattern.
+ * Returns a type where the render prop and the props it may clash with
+ * are separated as a discriminated union.
+ *
+ * It defaults the clashing props to `className` and `children`.
+ * This can be overridden by passing a second argument to the generic.
+ */
+export type WithRenderProps<
```
dgca, line 79: Is this useful?

packages/onchainkit/src/ui/Button.tsx
```diff
@@ -0,0 +1,31 @@
+import { type ButtonHTMLAttributes } from 'react';
+import { cn, pressable, text } from '@/styles/theme';
+
+type ButtonProps = ButtonHTMLAttributes<HTMLButtonElement>;
+
+export function Button({
```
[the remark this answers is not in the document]
xtbase: I think having a reusable button component is useful, but what do you think about just pulling into radix-ui button and applying our default styles over it, this way we can also support their loading prop and standardize loading button pattern

xtbase (approved):

packages/onchainkit/src/internal/types.ts
xtbase: gotcha 👍

packages/onchainkit/src/signature/components/SignatureButton.tsx
```diff
@@ -1,79 +1,79 @@
...
+}>;
 
 export function SignatureButton({
   className,
   disabled = false,
   label = 'Sign',
-  connectLabel = 'Connect Wallet',
   errorLabel = 'Try again',
   successLabel = 'Signed',
   pendingLabel = 'Signing...',
```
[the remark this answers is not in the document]
dgca: I wanna punt on this cause I see your point but also the benefits of making this easy to customize. I can revisit in a follow up

abcrane123 (approved):

PR #2361, opened 2025-05-13
Feat: TextInput updates

xtbase:
**What changed? Why?**

- a11y updates
- support error state
- support existing input props being able to be passed in 

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/packages/onchainkit/src/internal/components/TextInput.tsx
+++ b/packages/onchainkit/src/internal/components/TextInput.tsx
@@ -4,10 +4,15 @@ import {
   type ForwardedRef,
   useCallback,
   forwardRef,
+  HTMLProps,
 } from 'react';
 import { useDebounce } from '../hooks/useDebounce';
+import { cn } from '@/styles/theme';
 
-type TextInputReact = {
+type TextInputProps = Omit<
+  HTMLProps<HTMLInputElement>,
+  'aria-label' | 'className' | 'onChange' | 'onBlur' | 'onFocus'
+> & {
   'aria-label'?: string;
   className: string;
   delayMs?: number;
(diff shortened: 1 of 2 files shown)


packages/onchainkit/src/internal/components/TextInput.tsx
```diff
@@ -63,11 +63,14 @@ export const TextInput = forwardRef(
 
     return (
       <input
+        aria-disabled={disabled}
+        aria-errormessage={errorMessage}
```
dgca, line 67: issue: The value of `aria-errormessage` should be the ID of the element that renders the error message. Since we're not rendering the message, this doesn't really follow spec as is.

packages/onchainkit/src/internal/components/TextInput.tsx
```diff
@@ -63,11 +63,14 @@ export const TextInput = forwardRef(
 
     return (
       <input
+        aria-disabled={disabled}
```
dgca, line 66: nit: This shouldn't be necessary since inputs' disabled attribute is already accessible. This would only be needed for custom inputs (e.g. using divs or spans) or if the value was different than that of the input's disabled attribute

packages/onchainkit/src/internal/components/TextInput.tsx
```diff
@@ -2,25 +2,23 @@ import {
...
   inputMode?: InputHTMLAttributes<HTMLInputElement>['inputMode'];
-  onBlur?: () => void;
   onChange: (s: string) => void;
-  onFocus?: (e: React.FocusEvent<HTMLInputElement>) => void;
   placeholder: string;
   setValue?: (s: string) => void;
-  value: string;
   inputValidator?: (s: string) => boolean;
+  /** specify 'message' to show error state (change in color), field is used for a11y purposes, not actually rendered currently */
+  errorMessage?: string;
```
dgca, line 21: suggestion: Name this prop this `error` and type it as `string | boolean`. That way we don't have to pass a string that isn't rendered, and we can also show a text input in an error state without needing to show an error message.

xtbase: yea i was thinking maybe we could render an invisible error message

abcrane123 (approved):

PR #2363, opened 2025-05-13
chore: switch enums -> object consts

xtbase:
**What changed? Why?**

- Switch off of enums in favor of object consts ( can use string literals as a drop in replacement for referencing enum specifically )

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -53,4 +53,11 @@ export default [
       ],
     },
   },
+  {
+    settings: {
+      react: {
+        version: 'detect',
+      },
+    },
+  },
 ];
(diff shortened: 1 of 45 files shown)


abcrane123 (commented):
this mostly looks great! i just noticed that in quite a few places you've defined the map but aren't using it. i.e writing `"success"` instead of `CHECKOUT_LIFECYCLE_STATUS.SUCCESS`, is that intentional ?

xtbase:
@abcrane123  yea since it's using object enums, strings can be directly used where type can be inferred. I'm also okay switching them to be object usage, i figured using strings directly can increase legibility

abcrane123 (approved):

PR #2419, opened 2025-07-03
chore: Create new templates directory and add basic OnchainKit template

dgca:
**What changed? Why?**

- Stubs out a new template format for create-onchain
    - I'm putting these in their own directory in the root to to be able to run them and verify that they work correctly. It also feels cleaner, and this will let us add Vercel deploy buttons directly in the templates
    - This _just_ adds the template, I still have to make `create-onchain` use the new template. Then I can migrate the minikit templates as well
- Updates wagmi/viem
- Fixes some styling issues that hadn't been caught

**Notes to reviewers**

**How has it been tested?**


packages/onchainkit/plugins/babel-prefix-react-classnames.ts
```diff
@@ -103,6 +103,29 @@ export function babelPrefixReactClassNames({
...
+                    arg.alternate = types.stringLiteral(
+                      prefixStringParts(arg.alternate.value, prefix),
+                    );
+                  }
+                  if (types.isTemplateLiteral(arg.consequent)) {
+                    processTemplateLiteral(arg.consequent, prefix);
+                  }
+                  if (types.isTemplateLiteral(arg.alternate)) {
+                    processTemplateLiteral(arg.alternate, prefix);
+                  }
```
dgca, line 128: Added these so the babel plugin prefixes stuff like:

```
className={cn(
  someBool ? "text-red-500" : "text-blue-500",
  otherBool && "border-red-500"
)}
```

packages/onchainkit/src/checkout/components/CheckoutProvider.tsx
```diff
@@ -322,15 +324,23 @@ export function CheckoutProvider({
...
+          };
+        }),
+        capabilities:
+          isSponsored && paymaster
+            ? {
+                paymasterService: {
+                  url: paymaster,
+                },
+              }
+            : undefined,
```
dgca, line 343: Updated to the latest wagmi, so I had to update this to use `useSendCalls` instead of the experimental `useWriteContracts`

packages/onchainkit/src/wallet/components/Wallet.tsx
```diff
@@ -80,7 +81,7 @@ export function Wallet({
 
   // prevents SSR hydration issue
   if (!isMounted) {
-    return null;
+    return <Button disabled>Connect Wallet</Button>;
```
dgca, line 84: This prevents the button from just popping into view and causing a layout shift

packages/onchainkit/src/earn/components/RenderDepositButton.tsx
```diff
@@ -17,14 +17,16 @@ export function RenderDepositButton({
...
 }: RenderDepositButtonProps) {
-  const classNames = cn(
-    pressable.primary,
-    'rounded-ock-default',
-    'w-full rounded-xl',
-    'px-4 py-3 font-medium leading-6',
-    isDisabled && pressable.disabled,
-    text.headline,
-    'text-ock-foreground-inverse',
+  const classNames = prefixClassName(
```
xtbase, line 20: is it possible to have this be handled internally? it feels like additional thing to keep in mind if we need to manually prefix in certain places. what about something like `cn` reading a special env var we inject during build to prefix?

packages/onchainkit/src/fund/components/FundButton.tsx
```diff
@@ -105,7 +105,7 @@ export function FundButton({
 
   const buttonColorClass = useMemo(() => {
     if (buttonState === 'error') {
-      return 'bg-ock-error';
+      return prefixClassName('bg-ock-error');
```
xtbase, line 108: if we leverage some env var logic in `cn` we may be able to just go with convention of using `cn` everywhere as a wrapper

xtbase (approved):
LGTM open to keeping the prefix util but feel like it may be able to be solved better

PR #2420, opened 2025-07-04
chore: Simplify MiniKit template

dgca:
**What changed? Why?**

- Adds simplified template for MiniKit
- Consolidates manifest config to root `manifest.config.ts` file
- Adds utility to validate manifest from OnchainKit

**Notes to reviewers**

**How has it been tested?**


templates/next-basic/app/test/route.ts
```diff
@@ -0,0 +1,7 @@
+import { NextRequest, NextResponse } from "next/server";
```
dgca, line 1: Remove this test route

xtbase (approved):

PR #2433, opened 2025-07-24
refactor: Use floating UI for wallet dropdown

dgca:
**What changed? Why?**

- Wraps WalletDropdown with floating-ui to position the dropdown using fixed position and in a portal. This fixes an issue where the dropdown causes its container to become scrollable if it has overflow-y auto

**Notes to reviewers**

**How has it been tested?**

- Unit tests
- Manually


xtbase (approved):
LGTM though i would prefer us re-using radix stuff since we're already using that. We can leave it for a future refactor though

PR #2479, opened 2025-09-09
feat: support farcaster username and avatar in identity components

teimurjan:
**What changed? Why?**

**Notes to reviewers**

**How has it been tested?**

### DIFF
--- a/packages/onchainkit/src/identity/utils/getAvatar.ts
+++ b/packages/onchainkit/src/identity/utils/getAvatar.ts
@@ -1,71 +1,36 @@
-import { getChainPublicClient } from '@/core/network/getChainPublicClient';
-import { isBase } from '@/core/utils/isBase';
-import { isEthereum } from '@/core/utils/isEthereum';
-import type {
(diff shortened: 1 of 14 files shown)


packages/onchainkit/src/identity/utils/getAvatar.ts
```diff
@@ -1,71 +1,36 @@
-import { getChainPublicClient } from '@/core/network/getChainPublicClient';
-import { isBase } from '@/core/utils/isBase';
-import { isEthereum } from '@/core/utils/isEthereum';
-import type {
-  Basename,
-  GetAvatarParams,
-  GetAvatarReturnType,
-} from '@/identity/types';
+import type { GetAvatarParams, GetAvatarReturnType } from '@/identity/types';
```
teimurjan, line 1: At some point, minikit name and avatar will be fetched for all users via an API. To facilitate future additions, I added atomic functions for avatar and name (base and ENS). These functions are consumed by the final `getAvatar` and `getName`

PR #2502, opened 2025-09-22
fix: Minikit template prepends process.env.VERCEL_URL with https://

dgca:
**What changed? Why?**

- Updates the minikit template to prepend VERCEL_URL with https:// since it doesn't include the protocol
- Adds an example tag to manifest so it's fully validated

**Notes to reviewers**

**How has it been tested?**


cb-heimdall:
Review Error for soheimam @ 2025-09-22 20:18:38 UTC
User must have write permissions to review

PR #2676, opened 2026-09-03
fix: give TransactionButton an accessible name while a transaction is in progress

### LINKED ISSUE #2675: Bug: TransactionButton has no accessible name while a transaction is approving or pending

`TransactionButton` swaps its entire label for a bare spinner while a
transaction is being submitted or is pending, with no accessible name left
on the button. A screen reader user who has already started a transaction
gets no indication that anything is happening, or that the button is even
still a button with a purpose, until it either succeeds or fails.

1. Run this repo's own playground app (`packages/playground`), which already
   demonstrates `Transaction` with `TransactionButton` in
   `packages/playground/components/demo/Transaction.tsx`, with a screen
   reader running or the accessibility tree open in devtools.
2. Connect a wallet and click the demo's `TransactionButton` to submit a
   transaction.
3. While the transaction is being signed/submitted and while it is pending
   on-chain, inspect the button (`data-testid="ockTransactionButton_Button"`).

The button should keep an accessible name throughout, for example
"Transact, pending" or similar, so a screen reader announces both that the
control is still a button and what state it is in. Marking it `aria-busy`
would also be reasonable in addition to a name.

### Actual behavior

`TransactionButton`'s content is computed in `buttonContent`
(`src/transaction/components/TransactionButton.tsx`, lines 63-75): when
`isLoading` is true, it returns `<Spinner />` and nothing else. The `Spinner`
component (`src/internal/components/Spinner.tsx`) renders two `<div>`
elements with no text, `aria-label`, or `role`. The `<button>` element
itself (lines 108-127) sets no `aria-label` and has no visually-hidden text
fallback, so once `buttonContent` becomes the spinner, the button's
accessible name is empty. axe reports this as `button-name`.

```tsx
// TransactionButton.tsx
const buttonContent = useMemo(() => {
  if (receipt) return 'View transaction';
  if (errorMessage) return 'Try again';
  if (isLoading) return <Spinner />;   // <- no text alternative
  return idleText;
}, [isLoading, errorMessage, receipt, idleText]);
...
<button ... disabled={isDisabled} data-testid="ockTransactionButton_Button">
  {buttonContent}
</button>
```

@coinbase/onchainkit 1.1.2

### WCAG success criterion

4.1.2 Name, Role, Value.

### Rule

axe-core `button-name`.

### Suggested fix

Keep visually-hidden text on the button while `isLoading` is true (e.g. a
`sr-only`/`aria-label` reflecting the idle text plus "pending", or a fixed
string like "Transaction pending"), and consider `aria-busy="true"` on the
button for the duration. `Spinner` itself could also take an optional label
prop rather than always rendering as purely decorative markup.

Found while auditing wallet-connect kits for accessibility. A reproduction
script is available on request.

owenpkent:
**What changed? Why?**

Fixes #2675.

`TransactionButton` replaces its label with a bare `<Spinner />` while a transaction is being signed or is pending. `Spinner` renders no text, so the `<button>` has no accessible name for the whole time it is disabled. axe-core reports it as `button-name` (WCAG 4.1.2), and a screen reader user who has just submitted a transaction is told nothing about the control they were on.

This adds `aria-label="Transaction in progress"` and `aria-busy="true"` to the button for exactly the branch that shows the spinner (no receipt, no error, `isLoading`). Both come off again as soon as the button reads "View transaction", "Try again" or the idle text, so the visible label stays the accessible name in every other state.

The wording is the one the codebase already uses: `CheckoutButton` sets `aria-label={isLoading ? 'Transaction in progress' : buttonText}`, and `TransactionToastLabel` shows "Transaction in progress" for the same state, so the button and the status line now say the same thing.

The `render` prop path is unchanged. Custom renderers already receive `status` and own their own markup.

**Notes to reviewers**

- Scoped to `TransactionButton`, which is what the issue reports. The same spinner swap happens in `SwapButton`, `BuyButton`, `FundButton`, `NFTMintButton`, `ConnectWallet` and the earn and send render buttons, and none of them set a name while loading. Happy to follow up on those separately if wanted.
- `aria-busy` is only rendered while pending rather than as `aria-busy="false"` the rest of the time, so the idle DOM is exactly as before.
- Changeset included (patch).

**How has it been tested?**

- Three new tests in `TransactionButton.test.tsx`: the button is found by role with the name "Transaction in progress" and carries `aria-busy="true"` while loading; the visible text is the name and neither attribute is present when idle; and the "View transaction" and "Try again" states are unaffected even with `isLoading` still true. `pnpm f:ock test:coverage`, `pnpm f:ock lint` and prettier all pass locally.
- Re-ran the axe-core and IBM Equal Access scans that found the issue against the built package, in a small Vite app with an injected test EIP-1193 wallet that can hold a transaction in the approving and pending states. axe `button-name`: 2 occurrences before (approving, pending), 0 after. Equal Access `input_label_exists`: 2 before, 0 after. No other finding changed, apart from two unrelated `color-contrast` hits on the confirmed state's toast that did not recur on the second run.

### DIFF
--- a/.changeset/transaction-button-accessible-name.md
+++ b/.changeset/transaction-button-accessible-name.md
@@ -1,0 +1,5 @@
+---
+"@coinbase/onchainkit": patch
+---
+
+fix: Give TransactionButton an accessible name and aria-busy while a transaction is in progress
--- a/packages/onchainkit/src/transaction/components/TransactionButton.test.tsx
+++ b/packages/onchainkit/src/transaction/components/TransactionButton.test.tsx
@@ -325,4 +325,50 @@ describe('TransactionButton', () => {
     fireEvent.click(button);
     expect(onSubmit).toHaveBeenCalled();
   });
+
+  it('exposes an accessible name and busy state while the transaction is in progress', () => {
+    (useTransactionContext as Mock).mockReturnValue({
+      isLoading: true,
+      lifecycleStatus: { statusName: 'init', statusData: null },
+    });
+    render(<TransactionButton text="Transact" />);
+    const button = screen.getByRole('button', {
+      name: 'Transaction in progress',
+    });
+    expect(button).toHaveAttribute('aria-busy', 'true');
+    expect(screen.getByTestId('ockSpinner')).toBeInTheDocument();
+  });
+
+  it('uses the visible text as the accessible name when not in progress', () => {
+    (useTransactionContext as Mock).mockReturnValue({
+      isLoading: false,
+      lifecycleStatus: { statusName: 'init', statusData: null },
+    });
+    render(<TransactionButton text="Transact" />);
+    const button = screen.getByRole('button', { name: 'Transact' });
+    expect(button).not.toHaveAttribute('aria-label');
+    expect(button).not.toHaveAttribute('aria-busy');
+  });
+
+  it('keeps the result text as the accessible name once the transaction has settled', () => {
+    (useTransactionContext as Mock).mockReturnValue({
+      isLoading: true,
+      lifecycleStatus: { statusName: 'init', statusData: null },
+      receipt: '123',
+    });
+    const { rerender } = render(<TransactionButton text="Transact" />);
+    const successButton = screen.getByRole('button', {
+      name: 'View transaction',
+    });
+    expect(successButton).not.toHaveAttribute('aria-busy');
+
+    (useTransactionContext as Mock).mockReturnValue({
+      isLoading: true,
+      lifecycleStatus: { statusName: 'init', statusData: null },
+      errorMessage: 'blah blah',
+    });
+    rerender(<TransactionButton text="Transact" />);
+    const errorButton = screen.getByRole('button', { name: 'Try again' });
+    expect(errorButton).not.toHaveAttribute('aria-busy');
+  });
 });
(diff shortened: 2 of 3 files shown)


cb-heimdall:
<h1>🟡 Heimdall Review Status</h1>
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
🟡
<code>0/2</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>1</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>2</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>

PR #2678, opened 2026-09-09
fix: harden create-onchain --manifest websocket against CSWSH and code injection

dgca:
**What changed? Why?**

Fixes SECBUGS-22415 (HackerOne #3858933).

`create-onchain --manifest` served an HTTP + WebSocket server on `0.0.0.0:3333` with no Origin validation and no authentication, then wrote the first message it received verbatim into `minikit.config.ts` and `.env`. `minikit.config.ts` is imported and executed by the scaffolded Next.js app, so a quote character in any field escaped the string literal and ran arbitrary code on every `next dev`, `next build`, and server render. The injected code persisted in the file, and the `domain` field poisoned `NEXT_PUBLIC_URL`, which controls `homeUrl`, `webhookUrl`, and `iconUrl` in the shipped Farcaster manifest.

Any LAN host, or any cross-origin page already open in the developer's browser, could win the race against the wallet-signing flow. The same-origin policy does not restrict WebSocket connections, so this was reachable from a malicious ad or a compromised site.

The ticket proposed three layered fixes. All three are implemented, since each closes a different step of the chain:

| Control | Change | Stops |
| --- | --- | --- |
| Transport | Bind `127.0.0.1` only; accept a handshake only from `http://127.0.0.1:3333` | LAN hosts, cross-origin browser pages, DNS rebinding |
| Authentication | Single-use 256-bit token, delivered in the URL fragment of the page the CLI opens, checked during the handshake with `timingSafeEqual` | Any client that did not receive the page |
| Output encoding | Validate `header`/`payload`/`signature` as base64url and `domain` as a plain `http(s)` URL, then write every value with `JSON.stringify` | The CWE-94 injection primitive itself |

Everything uses the literal `127.0.0.1` rather than `localhost`. `localhost` can resolve to either `::1` or `127.0.0.1`, so a local process squatting `[::1]:3333` could have received the opened page, read the token from the fragment, and then connected to `127.0.0.1:3333` under an allowlisted origin.

**Notes to reviewers**

The fragment is the delivery channel for the token because fragments are stripped before the request is sent and are not included in `Referer`, so the wallet-connect popups the page opens cannot leak it.

This also fixes lifecycle bugs surfaced while adding the above:

- Setting `maxPayload` made an oversized frame emit an unhandled `'error'` on the client socket, which terminates Node. The denial-of-service guard was itself a denial of service. Added `'error'` handlers on the socket, the `WebSocketServer`, and the HTTP server.
- `server.close()` leaves upgraded WebSocket connections open, so an exported call to `createMiniKitManifest()` could keep the process alive. Replaced with an idempotent `settle()` that terminates clients first. The CLI masked this by calling `process.exit(0)` immediately.
- `open()` rejection was an unhandled rejection and an indefinite hang. It now prints the URL so the developer can continue manually.
- `cli.ts` always exited `0`. It now exits `1` on failure, which matters because malformed payloads now intentionally take the failure path.
- `domain` now rejects credentials, a query, and a fragment. These land in `NEXT_PUBLIC_URL`, which the app concatenates onto to build the manifest URLs.

The `JSON.stringify` layer is unreachable given the character allowlists that run before it. That is intentional defence in depth, and it is not separately testable, so no test asserts it in isolation.

Not included: no session timeout. That is pre-existing behaviour, and the token now prevents an attacker from holding the flow open.

**How has it been tested?**

`packages/create-onchain/src/minikit.test.ts` (unit, `http` and `ws` mocked) covers the loopback bind argument, the `verifyClient` predicate for each allow and deny case, `maxPayload`, per-run token uniqueness, and rejection of each malicious field shape.

`packages/create-onchain/src/minikit.manifest-server.test.ts` is new. `minikit.test.ts` mocks the transport, so it can only assert which options were passed. This file runs the real server and drives real WebSocket clients against it, which is the only way to prove a hostile client never reaches the code that writes `minikit.config.ts`. It asserts the config file is byte-identical to the template after each attack.

Verified the new tests are load-bearing by mutating the source:

| Mutation | Result |
| --- | --- |
| Remove the Origin check | 2 tests fail |
| Remove the token check | 3 tests fail |
| Remove the loopback bind argument | 8 tests fail |
| Remove token forwarding in `Page.tsx` | 2 tests fail |

Also ran the built CLI end to end against live WebSocket clients. Every attack variant from the report is refused at the handshake with a 401, the LAN interface is unreachable, an injection payload sent over an authorised connection is rejected by validation and leaves both files untouched, and the legitimate flow writes exactly what it wrote before.

Commands run:

- `pnpm f:create test` — 84 passed, 7 files
- `pnpm f:create test:coverage` — thresholds met
- `pnpm f:create check:types` — clean
- `pnpm f:create build` — clean
- `pnpm f:manifest test` — 61 passed
- `pnpm f:manifest test:coverage` — 100% maintained
- `npx tsc -b --noEmit` in `miniapp-manifest-generator` — clean
- `npx prettier --check` on all touched files — clean

### DIFF
--- a/.changeset/eighty-donkeys-shave.md
+++ b/.changeset/eighty-donkeys-shave.md
@@ -1,0 +1,5 @@
+---
+"create-onchain": patch
+---
+
+fix: Harden the `create-onchain --manifest` WebSocket server against cross-origin connections and config file code injection
--- a/packages/create-onchain/src/cli.ts
+++ b/packages/create-onchain/src/cli.ts
@@ -72,8 +72,8 @@ Available Templates:
   }
 
   if (isManifest) {
-    await createMiniKitManifest();
-    process.exit(0);
+    const succeeded = await createMiniKitManifest();
+    process.exit(succeeded ? 0 : 1);
   }
 
   if (isMiniKit) {
--- a/packages/create-onchain/src/minikit.manifest-server.test.ts
+++ b/packages/create-onchain/src/minikit.manifest-server.test.ts
@@ -1,0 +1,288 @@
+/**
+ * End-to-end coverage for SECBUGS-22415.
+ *
+ * `minikit.test.ts` mocks `http` and `ws`, so it can only assert that the
+ * hardening options are passed. These tests run the real server and drive real
+ * WebSocket clients against it, which is the only way to prove that a hostile
+ * client never reaches the code that writes `minikit.config.ts`.
+ */
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import fs from 'node:fs';
+import net from 'node:net';
+import os from 'node:os';
+import path from 'node:path';
+import { WebSocket } from 'ws';
+import { createMiniKitManifest } from './minikit';
+
+vi.mock('open', () => ({ default: vi.fn() }));
+// `analyticsPrompt` is unused by the manifest flow but pulls in prompts on import.
+vi.mock('./analytics.js', () => ({ analyticsPrompt: vi.fn() }));
+
+const open = (await import('open')).default as unknown as ReturnType<
+  typeof vi.fn
+>;
+
+const CONFIG_TEMPLATE = `export const minikitConfig = {
+  accountAssociation: {
+    header: "",
+    payload: "",
+    signature: "",
+  },
+};
+`;
+
+const validAssociation = {
+  header: 'eyJmaWQiOjEyMywidHlwZSI6ImN1c3RvZHkifQ',
+  payload: 'eyJkb21haW4iOiJleGFtcGxlLmNvbSJ9',
+  signature: 'MHgxMjM0NTY3ODkwYWJjZGVm',
+  domain: 'https://example.com',
+};
+
+const EXPECTED_ORIGIN = 'http://127.0.0.1:3333';
+
+let projectDir: string;
+let previousCwd: string;
+
+function readOpenedUrl() {
+  const url = open.mock.calls[0]?.[0] as string | undefined;
+  if (!url) {
+    throw new Error('the manifest flow never opened a browser');
+  }
+  return url;
+}
+
+function tokenFromOpenedUrl() {
+  return new URL(readOpenedUrl()).hash.replace('#token=', '');
+}
+
+/**
+ * Resolves `'connected'` once the handshake succeeds, or `'rejected'` if the
+ * server refuses the upgrade.
+ */
+function handshake({
+  origin,
+  token,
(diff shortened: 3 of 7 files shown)


cb-heimdall:
<h1>✅ Heimdall Review Status</h1>
<table>
<tr>
<th>Requirement</th>
<th>Status</th>
<th>More Info</th>
</tr>
<tr>
<td>Reviews</td>
<td>
✅
<code>1/1</code>
</td>
<td>
<details>
<table>
<tr>
<td>
<details>
<table>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td>
</td>
<td>1</td>
</tr>
<tr>
<td>
<details>
<table>
<tr>
<td><strong>Max</strong></td>
<td><strong>0</strong></td>
</tr>
</table>
</details>
</td>
<td>0</td>
</tr>
<tr>
<td>
From <code>CODEOWNERS</code>
</td>
<td>0</td>
</tr>
<tr>
<td>Global minimum</td>
<td>0</td>
</tr>
<tr>
<td><strong>Max</strong></td>
<td>
<strong>
1
</strong>
</td>
</tr>
</table>
</details>
</td>
<td>1</td>
</tr>
<tr>
<td>0</td>
</tr>
<tr>
<td><strong>Sum</strong></td>
<td><strong>1</strong></td>
</tr>
</table>
</details>
</td>
</tr>
</table>
<details>
<summary><h2>⚠️ Ignored Reviews (1)</h2></summary>
<table>
<tr>
<th>Reviewer</th>
<th>Reason</th>
</tr>
<tr>
<td>0xClouds</td>
<td>Soft review, as user does not have write permissions</td>
</tr>
</table>
</details>

dgca:
The two failing `test (22.x)` checks are the `cli: build & install onchainkit` and `cli: build & install minikit` workflows. Both are pre-existing failures on `main`, unrelated to this change.

They scaffold a project with the CLI and run `next build`, which fails resolving `@x402/core/client`, `@x402/svm/exact/client` and `@x402/evm`. Those come in transitively via `@coinbase/cdp-sdk` → `@base-org/account` → `@wagmi/connectors` → `onchainkit/wallet`, and have nothing to do with `create-onchain --manifest`.

Last green run of these workflows was 2025-09-22; every run since 2025-12-15 has failed on `main`.

Everything else is green, including all 8 `build (22.x)` jobs, which are what run the unit tests.

0xClouds (approved):

aayushijain21 (approved):

